PR #390, opened 2022-12-05
Read input from stdin

shonfeder:
I thought this would help make tntc a bit more unixy and also make it
easier to write some simple stuff in our integration tests.

This follows the relatively common convention documented in
https://www.baeldung.com/linux/dash-in-command-line-parameters

However, it is currently blocked by https://github.com/yargs/yargs/issues/1312#issuecomment-1338157693

Two other alternatives would be:

- read from `stdin` if a `--stdin` flag were given
- read from stdin if an input file isn't given. This is what `cat` does. This
  can sometimes be a git confusing however, as the program can appear to be
  hanging if accidentally leave off the file name and don't realize what you're
  doing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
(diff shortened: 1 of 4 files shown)


shonfeder:
Looks like we can do this after all: https://github.com/yargs/yargs/issues/1312#issuecomment-1454284840

PR #466, opened 2022-12-22
Build vscode plugin in all CI operating systems

shonfeder:
We have not been checking CI whether the vscode plugin can be compiled on Mac
and Windows. This change adds that CI check.


shonfeder:
VSCode builds in Windows and MacOS! (On CI at least :sweat_smile: ).

Issue #432, opened 2022-12-15
Enforce type variables/aliases naming pattern


bugarela:
Suppose we have a type alias named `proc`:
```scala
type proc = {id : int}

def Proc(id) : (int) => proc = { id : id }
```

According to our [type system specification](https://github.com/informalsystems/tnt/blob/main/doc/lang.md#type-system-12), lowercase names are always type parameters, so the type for `Proc` is interpreted as `forall proc . (int) => proc` (equivalent to `forall a . (int) => a`). You'd need to call it `PROC` to make it a type alias.

But, of course, the parser should have flagged an error at `type proc = {id: int}` since we can't have lowercase type aliases, or the name resolver should've complained.

Issue #406, opened 2022-12-09
Update examples


shonfeder:
Ensure all examples are updated to latest syntax and best practices and using correct notation.

PR #472, opened 2022-12-22
Update examples

bugarela:
Hello :octocat: 

This fixes our examples so they are all properly typechecked :tada: (except for `counters.tnt` which uses runs, as we don't have effects/modes implemented for runs yet).

I also made 4 small fixes (described in the first 4 commits of this PR), including enforcing that type names start with uppercase letters, which closes https://github.com/informalsystems/quint/issues/432.

Also closes https://github.com/informalsystems/quint/issues/406


examples/Paxos/Paxos.qnt
```diff
@@ -1,298 +1,270 @@
 module Paxos {
```
bugarela, line 1: IDK why git got so lost here, but I just:
1. Added indentation
2. Change `MSGS` to use discriminated unions and got rid of the `TypeOK` check for it.
3. Fixed an import issue

konnov (approved):
LGTM

PR #475, opened 2022-12-23
Prepare version 0.0.9 for publishing

bugarela:
Hello :octocat:

Here are some changes I made to allow publishing the VSCode extension to the marketplace. Here is the published extension: https://marketplace.visualstudio.com/items?itemName=informal.quint-vscode

I published before opening the PR because I had to iterate a bit with Microsoft's verification on publishing to get things right.

The README and the CHANGELOG still need some improvements. I'll do that next (in a new PR) and cut the `0.1.0` release (after the new PR is merged).


konnov (approved):
LGTM. For the future, it would be great to keep the versions of all quint packages (including the extension) in sync.

bugarela:
> For the future, it would be great to keep the versions of all quint packages (including the extension) in sync.

I don't think that is a good idea. There are dependencies between those packages and they are published independently (for now the server is embedded in the extension, but we'll probably publish it independently as well). Imagine that we have to cut a release of the LSP and the client every time we make a change to the tool? Or bump the tool because we change some UI thing in the extension client?

konnov:
How about keeping major.minor in sync and let the revisions deviate?

PR #476, opened 2022-12-23
Add C-like signatures

### LINKED ISSUE #102: Allow C-like type signatures
```
def opertor(p1: int, p2: string): int { ... }
```
on addition of or instead of
```
def operator(p1, p2): (int, string) => int
```

konnov:
Closes #102. This PR lets the user write C-like type signatures in definitions such as:

```scala
pure def min(x: int, y: int): int = if (x < y) x else y

pure def abs(x: int): int = if (x > 0) x else -x
```

I must say that these signatures look quite nice :)

The change itself is relatively small: A few lines in the grammar and one handler (`exitOperDef`) in `ToIRListener`. However, it required the following changes that occupy the largest part of this PR:

 - Using `Maybe<QuintType>` in `ToIrListener`. At the end, it was not necessary, but I decided to keep it, as the code looks better.
 - Fixing all test fixtures, as the output format has changed in the meantime.


quint/src/generated/Quint.g4
```diff
@@ -30,8 +30,17 @@ unit :    'const' IDENTIFIER ':' type                     # const
         //}                                                 # nondetError
         ;
 
-// an operator definition
-operDef : qualifier normalCallName params? (':' type)? ('=' expr)? ';'?
+// An operator definition.
+// We embed tow kinds of parameters right in this rule.
```
bugarela, line 34:
```suggestion
// We embed two kinds of parameters right in this rule.
```

bugarela (approved):
Congrats in making this so simple and thank you for updating the fixtures :heart:

PR #477, opened 2022-12-23
update codetour tutorial instructions

konnov:
Since we have the VSCode extension published, I've updated the instructions in the codetour tutorial.


bugarela (approved):
:heart:

PR #491, opened 2023-01-11
Cleanup and tests for the VSCode extension

bugarela:
Hello :octocat: 

This gives the VSCode extension code some deserved love. Since we plan to add several features to it, this was really necessary. What I did:
1. Removed a bunch of unused boiler-plating for features we don't support yet.
2. Moved some internal functionality away from `server.ts`, creating `inferredData.ts` for managing inferred types, effects and modes; and `reporting.ts` for functions related to reporting errors and useful information. The API for the functions in those files was also improved to reduce the usage of global server variables.
3. Added unit tests for the internal functions (not yet for server level requests).

Sorry for the huge diff, but this (hopefully) does not change any behavior, so there's no need for a very careful review.


quint/src/cliCommands.ts
```diff
@@ -175,24 +175,18 @@ export function typecheck(parsed: ParsedStage): CLIProcedure<TypecheckedStage> {
...
-        return cliErr("typechecking failed", { ...typechecking, errors })
-      } else {
-        return right({ ...typechecking, types, effects, modes })
-      }
-    })
+  const [modeErrMap, modes] = checkModes(module, effects)
+  const modeErrors: ErrorMessage[] = Array.from(modeErrMap, errorLocator)
+
+  // TODO add once logging functionality is added
+  // console.log("mode checking succeeded")
```
konnov, line 182: would it make sense to open an issue on adding logging to the vscode plugin?

quint/src/effects/modeChecker.ts
```diff
@@ -35,14 +34,10 @@ import { effectToString } from './printing'
  */
 export function checkModes(
   quintModule: QuintModule, effects: Map<bigint, Effect>
-): Either<Map<bigint, ErrorTree>, Map<bigint, OpQualifier>> {
+): [Map<bigint, ErrorTree>, Map<bigint, OpQualifier>] {
```
konnov, line 37: I thought `Either` was our preferred way. Why would we switch to a pair here?

konnov (approved):
The code after the refactoring looks much cleaner to me. I left only a couple of comments.

quint/src/cliCommands.ts
bugarela: This is actually related to quint CLI, but I'll check if there's an issue for that

quint/src/effects/modeChecker.ts
bugarela: In order to return partial information. That is, even if there are errors inferring modes somewhere, the inferred information for definitions where inference was successful can be used. This was an improvement I made for types and effects, but didn't do for modes at the time, so I'm doing it now.

Issue #483, opened 2023-01-09
Example in "Namespaces and Imports" does not type check


lasarojc:
The following example from the documentation

```
module Outer {
  var x: int

  def xPlus(z) = x + z

  module Inner {
    val x2 = x + x
  }

  val inv =
    (Inner.x2 - x == x)
}
```

Does not type check.

konnov:
Just to better understand what's going on, does not it typecheck in VSCode?

`quint typecheck Foo.qnt` actually works for me.

konnov:
As a side problem, it does not work in REPL, as REPL does not support constants yet. The error message is using the outdated syntax:

```
compile error: <input>:8:5 - error: Assignment <- needs two arguments
```

shonfeder:
I am also not able to reproduce on

```
$ quint --version
0.5.0
```

lasarojc:
My bad. I copied the first example by mistake and stuck to it, but it is actually the second example. The issue is updated.

lasarojc:
```
➭ quint typecheck scrap.qnt
... scrap.qnt:7:14 - error: Failed to resolve name x in definition for x2, in module Inner
7:     val x2 = x + x
                ^

/Users/lasaro/code/informal/commetbft/spec/consensus/reactor/scrap.qnt:7:18 - error: Failed to resolve name x in definition for x2, in module Inner
7:     val x2 = x + x
                    ^

error: parsing failed
```

shonfeder:
thanks for explaining! here's the error:

```
$  quint typecheck scratch.qnt
/home/sf/Sync/informal-systems/apalache/quint/scratch.qnt:7:14 - error: Failed to resolve name x in definition for x2, in module Inner
7:     val x2 = x + x
                ^

/home/sf/Sync/informal-systems/apalache/quint/scratch.qnt:7:18 - error: Failed to resolve name x in definition for x2, in module Inner
7:     val x2 = x + x
                    ^

error: parsing failed
```

konnov:
We should probably ask @bugarela. This may be outdated syntax actually.

bugarela:
I didn't notice we intended to support that, so this variable visibility is indeed not implemented. The question is: do we still want to support this? The outer module can already access names from the inner modules with namespaces, perhaps it would be too much to allow the other way around as well.

konnov:
This syntax is probably redundant, if we can `import Inner.x2`

shonfeder:
[...]

PR #500, opened 2023-01-12
Scoreboard and rules

konnov:
Adding a simple scoreboard on the main page and a few rules


README.md
```diff
@@ -34,6 +34,26 @@ element. A lot of alchemy is about transmutation and energy, and Quint makes it
 possible to transmute specifications into executable assets and empower ideas to
 become referenced artifacts.
 
+## Scoreboard
+
+We appreciate the efforts of our peers for going through our minimal viable
+solution -- a codeword for buggy tooling and documentation and nobody is sure
+how exactly it should be working -- and generating excellent suggestions for
```
shonfeder, line 41:
```suggestion
solution and generating excellent suggestions for
```

scoreboard-rules.md
```diff
@@ -0,0 +1,44 @@
+# Quint Scoreboard Rules
+
+We appreciate the efforts of our peers for going through our minimal viable
+solution -- a codeword for buggy tooling and documentation and nobody is sure
+how exactly it should be working -- and generating excellent suggestions for
```
shonfeder, line 5:
```suggestion
solution and generating excellent suggestions for
```

scoreboard-rules.md
```diff
@@ -0,0 +1,44 @@
...
+The tutorials contain secret codes that you can submit, in order to get
+:earth_africa: scores. Either submit a secret code in a DM to @konnov (if you
+know how to reach him), or create a secret [gist](https://gist.github.com/),
+write your secret code in the gist and mention @konnov there.
+
+## Github contributions [air :cyclone:]
+
+ - A few-line fix in the documentation: **3** :cyclone:
+
+ - A confirmed bug report for something that has a simple workaround: **5**
```
shonfeder, line 20: let's  fix bugs  rather than implement workaround ;)

```suggestion
 - A confirmed bug report for something that has a simple fix: **5**
```

scoreboard-rules.md
```diff
@@ -0,0 +1,44 @@
...
+   :cyclone:
+
+ - A confirmed bug report that has no simple workaround: **10**
```
shonfeder, line 23:
```suggestion
 - A confirmed bug report that has no simple fix: **10**
```

scoreboard-rules.md
```diff
@@ -0,0 +1,44 @@
...
+   :cyclone:
+
+ - A confirmed bug report that leads to language refactoring: **20** :cyclone:
```
shonfeder, line 26: I'm not sure what is meant by "language refectory"

shonfeder (changes requested):
I think this is a cute idea, however As I noted in our meetings, this really doesn't seem worth the effort to me. Maybe we can talk it over a  bit more. Aside from the amount of time would take to actually keep this up to date and track people's contributions, I'm worried this will make us seem amateurish and not a serious as a project. Personally,  this kind of thing in the read me of a project would not endear me.

If we do choose to do this, asked that we don't keep the scoreboard on the read me. Perhaps instead providing a link to wherever it lives. On the other hand providing a list of contributors or acknowledgements on the read me seems quite nice.

scoreboard-rules.md
konnov: What I have in mind is an issue that led to a change in the language. For instance, #483

konnov:
> I think this is a cute idea, however As I noted in our meetings, this really doesn't seem worth the effort to me. Maybe we can talk it over a bit more.

I would prefer to try it out. We can always remove the scoreboard, if takes too much effort. It may happen that it takes zero effort, if nobody wants to use it.

> Aside from the amount of time would take to actually keep this up to date and track people's contributions, I'm worried this will make us seem amateurish and not a serious as a project. Personally, this kind of thing in the read me of a project would not endear me.

At this point, I am OK to admit that Quint is amateurish. IMO, this is exactly the right time and state to attract external contributions. If people would believe that this project is just a complete release of a language developed in the shadows, they would be less willing to contribute.

> If we do choose to do this, asked that we don't keep the scoreboard on the read me. Perhaps instead providing a link to wherever it lives. On the other hand providing a list of contributors or acknowledgements on the read me seems quite nice.

Sure, we can place a link on the readme to the board. I just want it visible. A single link would be easy to miss.

bugarela (commented):
I like the approach of rewarding (with fun acknowledgments) the users that contribute to Quint. I like air, water and fire. For the earth part, I'm not so comfortable with the approach for getting the points. I can try to elaborate on this a bit if necessary, but I believe people would feel dis-encouraged to engage if they think secret codes are too silly. And I actually believe some few people might feel disrespected if they are asked or at some level expected to do it.

Let me propose an alternative specifically to the earth points: give those for people that provide structured feedback on the individual tutorials. We can create a google form for each tutorial with some nice questions and give points to anyone who answers them. This way, we'll get a better sense of how much the person actually invested in the tutorial (instead of just searching for secret codes), we'll get better feedback and people won't feel like they are spending their time in something silly (as sending Igor secret codes) only to get in the scoreboard - they are actually doing something important and will be acknowledged, just like with the other elements (air, water and fire).

konnov:
> And I actually believe some few people might feel disrespected if they are asked or at some level expected to do it.

How should we make it clear that nobody is expected to do that? It is just a fun game. I know that some competitions, especially in security, are doing that. Never heard of somebody feeling disrespected.

konnov:
> Let me propose an alternative specifically to the earth points: give those for people that provide structured feedback on the individual tutorials. We can create a google form for each tutorial with some nice questions and give points to anyone who answers them.

We already have discussions for that, and people can get water points other kinds of points for that. I was considering a google form too. But then read about how much spam people get with Google forms. This is definitely not something I want to deal with :)

konnov:
@shonfeder and @bugarela, I am really curious about why you think people may react negatively to the scoreboard idea. It's free software, everything is optional.

bugarela:
Before trying to write out my detailed thoughts, I thought I could find someone else with a similar opinion, and I think this points from [this blogpost] are similar to my concern points:

> (Ab)use of points, badges and leaderboards. This is the most visible and annoying aspect of gamification. Product designers started to attach virtual currencies to anything, under the silly premise that if you offer people something to collect, they will try to collect it no matter what. But virtual economies add cognitive noise, introducing unwanted distortions both when they are worth too much and when they are worth nothing.

> Displacement of rewards. [It’s been demonstrated](http://files.meetup.com/1789648/Review%20on%20Effects%20of%20Extrinsic%20Rewards%20on%20Intrinsic%20Motivation.pdf) that offering any kind of reward on behaviors that should happen spontaneously puts people into “transaction mode”, altering the original motivations system and leaving them less motivated than before. This includes, of course, virtual currencies and rewards.

> Condescending tone. Many gamified systems, for the sake of keeping users motivated, adopt a patronizing treatment, congratulating people in an overly cheerful voice for everything they do. Here, “user-friendly” was somehow interpreted as “toddler-friendly”, something that I suspect most adults won’t appreciate. A system that assumes that you need to be constantly led by the hand makes you feel sort of disabled (Clippy, anyone?). No one likes being treated as a puppy.

Basically, I think we should treat our contributors like interested professionals, and giving them fake awards for invaluable tasks (as submitting secret codes) in a system completely prone to abuse can be disrespectful. 

My suggestion with the google forms approach helps in the aspect that bothers me the most: the task (submitting secret codes) is invaluable.

konnov:
OK, I still find it fascinating that people react to virtual things like if they were real, but anyhow. How about keeping the other three elements (air, water, fire), as @bugarela suggested, and introducing a discussion poll on the tutorial. By doing so we can avoid obvious spam and let people write comments, if they want to contribute:

https://github.com/informalsystems/quint/discussions/508

shonfeder:
>  I am really curious about why you think people may react negatively to the scoreboard idea

I also have a negative reaction to it and i assume some people are like me in the relevant ways, so i infer some others will also. @bugarela's summary really rings true to me on all fronts.

additionally, imo there are many ways we can use our time to encourage learning and contribution more valuable than setting up and maintaining an adhoc points system. e.g.,

- 99 quint problems
- a learning roadmap
- add a table to thank contributors
- follow @bugarela's idea of fielding questions on SO (we can use their point system)
- dedicate regular time to *talking with our users* :smile_cat:

konnov:
> dedicate regular time to talking with our users 😸

Sure, this is what we have to do too. As far as I understand, it requires a lot of skills to get a real opinion on something in personal communication.

bugarela:
Are you concern about bot spam or random people spamming? For bot spam, I'm sure we can find some forms with captchas. For people spamming, they could spam with gists as well :laughing:. But yes, let's think this trough together. Thanks for listening to us!

konnov:
> Are you concerned about bot spam or random people spamming? For bot spam, I'm sure we can find some forms with captchas. For people spamming, they could spam with gists as well 😆. But yes, let's think this trough together. Thanks for listening to us!

It's both, as far as I understand. Google forms have some kind of captchas, but they require a Google account. Also staring at a Google spreadsheet produced by Google forms is not fun. I did that when processing student applications.

konnov:
> For people spamming, they could spam with gists as well 😆

This is true, but it definitely requires more skills. So I would expect people who have a GH account and are able to make a private gist have other things to do :)

konnov:
I have thought about the scoreboard. It may create wrong incentives among our first small group of users inside the company, as people may perceive this scoreboard as a form of control over their time invested in learning Quint. We can come back to this idea later. Q1 does not seem to be a good time for that.

PR #501, opened 2023-01-12
add codetour/markdown differentiation and a few codes

konnov:
This PR updates the tutorial generator and adds a few codes. Again, check `*.template.qnt`, not the generated files (hidden from the diff by default).

### DIFF
--- a/tutorials/script/mk-tutorial.py
+++ b/tutorials/script/mk-tutorial.py
@@ -131,7 +131,7 @@ def xmlToMarkdown(root, code, out):
                 if c.tag == "block":
                     out.write(f'{c.text}\n')
                 elif c.tag == "run":
-                    # use the special syntax for shell commands in CodeTour
+                    # use a code block for shell commands
                     out.write(f'\n```sh\n')
                     out.write(f'{c.text}\n')
                     out.write(f'```\n\n')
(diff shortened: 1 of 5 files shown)


tutorials/script/mk-tutorial.py
```diff
@@ -126,13 +127,14 @@ def stepsStr(i):
...
-            elif c.tag == "run":
-                # use the special syntax for shell commands in CodeTour
-                out.write(f'\n```sh\n')
-                out.write(f'{c.text}\n')
-                out.write(f'```\n\n')
+            if c.get("when", "markdown") == "markdown":
+                if c.tag == "block":
+                    out.write(f'{c.text}\n')
+                elif c.tag == "run":
+                    # use the special syntax for shell commands in CodeTour
```
bugarela, line 134:
```suggestion
                    # use a code block for shell commands
```

bugarela (commented):
It's not clear for me what are the secret codes here. Are they just the integers returned by evaluating the expressions? The first one has a "Secret1" name iiuc, but the second one just mentions extra scores? I think we'd need to clarify what exactly should be submitted as secret codes here, even if we are evaluating them manually (i.e. for the scoreboard).

konnov:
> It's not clear for me what are the secret codes here. Are they just the integers returned by evaluating the expressions? The first one has a "Secret1" name iiuc, but the second one just mentions extra scores? I think we'd need to clarify what exactly should be submitted as secret codes here, even if we are evaluating them manually (i.e. for the scoreboard).

I made it a bit obfuscated on purpose. Otherwise, people would just grep for "secret" and harvest all the codes. Yes, they are just the integers. But since the response is free form anyways, they can send whatever.

PR #504, opened 2023-01-13
Add missing flag to `copyfiles`

bugarela:
Hello :octocat: 

I had a false positive while testing this library and only now realize it is actually missing a flag. It is not trivial to test this in CI, as our test suite runs on top of the `ts` source files and not `dist` iiuc.

PS: I'm using the `copyfiles` lib instead of `cp` to provide support to all OS.

### DIFF
--- a/quint/package.json
+++ b/quint/package.json
@@ -88,7 +88,7 @@
     "yargs": "^17.2.1"
   },
   "scripts": {
-    "compile": "tsc && copyfiles ./src/builtin.qnt ./dist/src/",
+    "compile": "tsc && copyfiles -u 1 ./src/builtin.qnt ./dist/src/",
     "test": "mocha -r ts-node/register test/*.test.ts test/**/*.test.ts",
     "coverage": "nyc npm run test",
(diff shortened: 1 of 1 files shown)


quint/package.json
```diff
@@ -88,7 +88,7 @@
     "yargs": "^17.2.1"
   },
   "scripts": {
-    "compile": "tsc && copyfiles ./src/builtin.qnt ./dist/src/",
+    "compile": "tsc && copyfiles -u 1 ./src/builtin.qnt ./dist/src/",
```
konnov, line 91: what is `copyfiles`? I don't have such a command on MacOS

konnov (changes requested):
Trying to figure out `copyfiles`

quint/package.json
bugarela: It is a typescript library that handles copying files in multiple OS. `npm install` for `quint` will install it for you.

konnov: ah, ok, thanks for the clarification!

konnov (approved):
LGTM

PR #510, opened 2023-01-13
Hello world tutorial

konnov:
This PR adds a relatively simple "Hello, world!" tutorial. It is a lot of text, but not much is going on there.

As with the integers tutorial, please read and comment in `hello.template.qnt`, instead of `hello.{qnt,md}`, which are generated.


konnov:
Perhaps, @Kukovec could help here with the intro material.

tutorials/lesson0-helloworld/hello.template.qnt
```diff
@@ -0,0 +1,319 @@
...
+        <file>$file</file>
+        <line>$line</line>
+        <title>Introduce an action by the computer</title>
+        <description>
+Now it's time to print the "Hello, world!" message. Similar to `init`,
+we introduce the action `write`. In contrast to `init`, this action
+cannot be unconditionally executed in any state of our state machine.
+If you carefully read the description of the action `init` at the previous
+step, you should immediately see that `consoleOutput` changes
+to `"Hello, world!"` in the next state, if `write` returns `true`.
```
bugarela, line 147: Is this sentence correct? I can't understand it nor figure out what it was suppose to mean.

tutorials/script/mk-tutorial.py
```diff
@@ -133,29 +147,14 @@ def stepsStr(i):
...
-            for l in code[int(lastLine) + 1: int(line)]:
-                out.write(l)
-
-            out.write('```\n\n')
           
         # update the last line, so it can be used in the next iteration
         lastLine = line
 
-    out.write("# Done!\n")
+    out.write("## The end\n")
```
bugarela, line 154: Should we write the link to the feedback discussion to all tutorials?

>We are experimenting with different kinds of tutorials.
It would be great to learn, whether you liked this tutorial format, or not.
Please vote in the
[discussion](https://github.com/informalsystems/quint/discussions/509).

bugarela (approved):
I like this a lot, it's a great example! I think this "imperative" description (such as `all` "returning" `true` and executing things) is important for beginners. But I'd also like to see the declarative description here. Something like: "... More precisely, `write` defines all transitions from a state where `consoleOutput` is `""` to a state where `consoleOutput` is `"Hello, world!" and `readByUser` keeps its value". At some point, the users will have to think in declarative terms, and it might be misleading to completely hide them in our own tutorials.

tutorials/script/mk-tutorial.py
konnov: I would rather introduce a discussion for every tutorial separately. It may happen that something specific to a tutorial is affecting the answers. If we have different discussions, we would be able to see that there is another hidden variable in the tutorials.

konnov:
> I like this a lot, it's a great example! I think this "imperative" description (such as `all` "returning" `true` and executing things) is important for beginners. But I'd also like to see the declarative description here. Something like: "... More precisely, `write` defines all transitions from a state where `consoleOutput` is `""` to a state where `consoleOutput` is `"Hello, world!" and `readByUser` keeps its value". At some point, the users will have to think in declarative terms, and it might be misleading to completely hide them in our own tutorials.

I would like to avoid declarative semantics in Quint as much as possible, as it causes a lot of confusion. Moreover, many of our structural changes in the language allow for operational semantics. Do you think it would be impossible to give the users entirely operational semantics? The only source of problem I see is `Nat` and `Int`, as they introduce infinite sets.

tutorials/lesson0-helloworld/hello.template.qnt
konnov: I will rephrase it

PR #512, opened 2023-01-16
Show documentation on hover

### LINKED ISSUE #486: Show  documentation info on hover
The current implementation for overlay documentation shows it only on signature help request, that is, when `(` is typed. We should also show it on more accessible ways such as hovering over their calls.

bugarela:
Hello :octocat: 

The documentation overlay was being showed only as signature help, that is, when `(` gets typed. This PR introduces showing the documentation overlays when hovering over a name or operator application (i.e. hovering over `bar` in `foo(bar)` will show the documentation for the operator `foo`)

I couldn't find a perfect way of showing both the inferred information for the entire expression **and** the documentation for the name involved. This should be enough for now and we can polish it later with, perhaps, some user input.

hoverdocs

Closes #486


bugarela:
While working on a spec, I noticed the hover request is failing when I hover over types. I'll fix this soon, just noting it here so I don't forget.

konnov (approved):
Looks good. It seems hovering required quite a bit of careful coding. Thanks for taking care of it!

PR #515, opened 2023-01-16
remove the scores from the integers tutorial

konnov:
This PR removes any mentioning of scores from the integers tutorial. Also, the tutorial is generated again, after the updates done in the "Hello, world" tutorial.


bugarela (approved):
LGTM

PR #517, opened 2023-01-16
Refurbish the tutorial on Booleans

konnov:
In this PR, I have upgraded the tutorial on Booleans to the same format as hello world and the integers tutorials. So this tutorial is now in two formats: CodeTour and Markdown.


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Updated installation documentation (#471)
+- The tutorial on Booleans now comes in markdown and CodeTour (#517)
```
shonfeder, line 28: I guess this is an addition, not a fix of something broken?

tutorials/lesson1-booleans/booleans.qnt
```diff
@@ -1,42 +1,34 @@
...
-
- If you do not know how to start the lesson, do the following:
-
-  - Open Visual Studio Code.
-  - Install the Quint and Codetour extensions.
-  - Choose the tour "Lesson 1 - Booleans".
+/*
+ If you do not know how to start this lesson, see:
+ https://github.com/informalsystems/quint/tree/main/tutorials
  */
```
shonfeder, line 4: Is this alignment right on line 1?

tutorials/lesson1-booleans/booleans.qnt
```diff
@@ -1,42 +1,34 @@
...
 module booleansLesson {    
-    /// false is a built-in constant
-    pure val myFalse = false
 
-    // exercise: evaluate false in REPL
+    // false is a built-in constant
+    pure val myFalse = false
 
-    /// true is a built-in constant too
+    // true is a built-in constant too
```
shonfeder, line 10: afaiu, Quint doc comments are setoff by 3 `/`. So i'm unclear on the reason for this change.

tutorials/lesson1-booleans/booleans.qnt
```diff
@@ -1,42 +1,34 @@
...
     pure val myTrue = true
 
+    // Boolean negation, which is written as `!x` in some languages
+    pure def myNot(x) = not(x)
```
shonfeder, line 14: why define a wrapper around `not`?

shonfeder (commented):
A few initial notes.

shonfeder (approved):
I have some concerns with this format and the efforts here and duplicated work, but I’ve discussed that with Igor and elsewhere and none of it pertains to this PR in particular. I’m not sure how to review the template because I don’t understand the format, but I don’t want to block this since it is just an incremental improvement on a previously approved tutorial.

CHANGELOG.md
konnov: Good question. The tutorial was available before, but only in CodeTour. The new version is available in two formats. Yeah, that sounds more like an addition.

tutorials/lesson1-booleans/booleans.qnt
konnov: Yep, looks like wrong indentation. Fixed

tutorials/lesson1-booleans/booleans.qnt
konnov: These are not doc documents, but simple one liners

tutorials/lesson1-booleans/booleans.qnt
konnov: Since I wanted to write a real Quint module, the only way to introduce is expressions is via definitions. In REPL, we could just write the expression. If you check the generated markdown and CodeTour, there are also executable fragments that evaluate expressions directly with Quint.

konnov:
I agree that that the format is ad-hoc. The main reason was the integration of CodeTour and markdown. At some point, we would be able to converge at a common format for: tutorials, reference manuals, protocol specifications, and (maybe) integration tests.

PR #527, opened 2023-01-17
VSCode release script

### LINKED ISSUE #513: Document/script packaging and release process for the VSCode extension

bugarela:
Hello :octocat: 

Following @shonfeder's idea of writing a script with a mix of `echo` commands and actual automation, here's a first version of a script to release VSCode, used in #526.

This script prepares a branch with the release, and then we just have to run `vsce publish` on main after merging the release PR.

I'm also deleting some files that were not being used and/or shouldn't have been committed in the first place.

Closes #513


vscode/quint-vscode/scripts/prepare-release.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+QUINT_VERSION=0.5.2
```
shonfeder, line 3: can we take this from an env variable? or is there a reason we want it hard coded in the script?

vscode/quint-vscode/scripts/prepare-release.sh
```diff
@@ -0,0 +1,38 @@
...
+function pause(){
+   read -p "$*"
+}
+
+git checkout main;
+git pull origin main;
+echo "Ensure that the VSCode version was bumped."
+echo "Extension: $(cat package.json | jq .version | xargs)."
+echo "Server: $(cat server/package.json | jq .version | xargs)."
+echo "You can run npm version x.y.z to bump it. Remember to bump the server as well."
```
shonfeder, line 14: can we use `npm version <major|minor|patch>` rather than supplying the whole version? seems like it might reduce chance of errors. WDYT?

shonfeder (approved):
great start to automating and documenting :)

vscode/quint-vscode/scripts/prepare-release.sh
bugarela: TBH I still don't understand what that means. I've read this in your checklist and on `npm version` man page but didn't understand how to use it.

vscode/quint-vscode/scripts/prepare-release.sh
bugarela: I don't really know bash, so I didn't want to go to the trouble of figuring out how to handle undefined parameters at this point, so I just used the simplest approach I could think of. But this should be easy enough to copy past from stack overflow, I'll look it up.

bugarela:
I'm looking at the CI issue. I removed to many files, it seems. But we did have a sub-project for the client, which is a single file with a single dependency. Doesn't seem right to have 3 projects (with 3 versions in 3 different `package.json` files) for 1 extension. I'm trying to make it work without a project for the client.

vscode/quint-vscode/scripts/prepare-release.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+QUINT_VERSION=0.5.2
+
```
shonfeder, line 4: this script depends on being run from a particular directory. instead lets use 

```
# The directory of this file
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
```

to get the dir where this script lives and then use that to ensure we are in the appropriate dir.

shonfeder (changes requested):

vscode/quint-vscode/scripts/prepare-release.sh
shonfeder: `npm version patch` will increment just the patch version, `npm version minor` will bump the minor version.

vscode/quint-vscode/scripts/prepare-release.sh
shonfeder: at the top of the script, you can set these flags to handle undefined params and other common errors automatically:

```
set -euo pipefail

set -o xtrace
```

but a sanity check to report an error if this is unset probably also makes sense

vscode/quint-vscode/scripts/prepare-release.sh
```diff
@@ -0,0 +1,38 @@
...
+function pause(){
+   read -p "$*"
+}
+
+git checkout main;
+git pull origin main;
+echo "Ensure that the VSCode version was bumped."
+echo "Extension: $(cat package.json | jq .version | xargs)."
+echo "Server: $(cat server/package.json | jq .version | xargs)."
```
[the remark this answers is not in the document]
bugarela: `jq .version` returns a string like `"0.1.1"`. I was going to use `xargs` to transform that into `0.1.1`, but then realize `xargs` alone was enough to extract the content of the string.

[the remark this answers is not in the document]
shonfeder: Oh cool! That’s a nice trick :)

vscode/quint-vscode/scripts/prepare-release.sh
bugarela: Cool! I used this to `cd` into the parent directory of it, which should always be `quint-vscode` unless the script itself is moved. Is that what you had in mind?

shonfeder (approved):
Looks great!

vscode/quint-vscode/scripts/prepare-release.sh
shonfeder: Yep! Perfect. Thanks!

PR #538, opened 2023-01-18
Fix the order of arguments if `foldr` in REPL

### LINKED ISSUE #536: `foldr` inconsistent type definition and REPL implementation
`foldr` has the following type definition, inspired in programming languages:
```
(List[a], b, (a, b) => b) => b
```

But the REPL implements it as:
```
(List[a], b, (b, a) => b) => b
```

That is, the REPL expects the accumulator to be the first parameter in the lambda, while the type system and the documentation say it should be the second parameter.

cc @konnov

konnov:
Closes #536.


shonfeder (approved):
LGTM

PR #540, opened 2023-01-19
add warning for the features under construction

konnov:
This PR introduces warnings in the language manual, so people should not use unimplemented features, which we thought were too advanced for the beginners.


bugarela (approved):
Great!

Issue #399, opened 2022-12-07
Propagating effects is too manual


bugarela:
The effect system needs a way of propagating effects without us having to explicitly state the propagation of every single effect in every single signature. We have to introduce some new features and that will be too much work/noise if we don't fix this first. After fixing it, we should:
1. Ensure temporal effects can be propagated from most operators, see https://github.com/informalsystems/tnt/pull/392/files/248f3db1dfe1a4ee901afd1dec857d8c852e9f87#r1042053651
2. Introduce effect for runs
3. Introduce effect for nondet

bugarela:
#641 introduces a generalization that is a big step towards this. We should also find a way to generalize signatures and then we should be able to:
1. Improve propagation of temporal effects
2. Introduce new effects with no big overhead.

PR #551, opened 2023-01-19
Promote `assert` to be an action operator 

### LINKED ISSUE #542: Change the mode of `assert` from `run` to `action`
Initially, I thought that runs would be the best place for `assert`, since they are needed in tests the most.

However, after writing a few examples, I see this pattern emerging:

```scala
  run sendOkTest = {
    init
      .then(send("alice", "bob", Map("atom" -> 3)))
      .then(all {
        assert(balances.get("alice").get("atom") == 27),
        assert(balances.get("bob").get("atom") == 4),
        keepVars,
      })
  }
```

It works since the effects checker is ignoring asserts at the moment.

It seems that actions are the natural place for asserts. Since we want to reason about actions with operational semantics, it should be possible to define the behavior of asserts for actions. It's probably doable even in the symbolic case. I am less certain about allowing for `assert` in expressions.

konnov:
Closes #542. `assert` is now an action operator, not a run operator. I did not have to do much except moving the description in the manual.

As far as I understand, the mode checker does not enforce modes for some operators, e.g., I could do the following in REPL, which I expected the mode checker to complain about:

```
pure def a = assert(false)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Changed
+
+- `assert` is now an action operator, not a run operator (#542)
+
 ## v0.5.2 -- 2023-01-17
 
 ### Added
--- a/doc/lang.md
+++ b/doc/lang.md
@@ -57,10 +57,10 @@ TLA+](https://lamport.azurewebsites.net/tla/summary.pdf).
     - [Operators on actions](#operators-on-actions)
       - [Delayed assignment](#delayed-assignment)
       - [Non-deterministic choice](#non-deterministic-choice-1)
+      - [Assert](#assert)
     - [Runs](#runs)
       - [Then](#then)
       - [Repeated](#repeated)
-      - [Assert](#assert)
     - [Temporal operators](#temporal-operators)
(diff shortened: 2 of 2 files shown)


bugarela:
It doesn't complain about `assert(false)` but it should complain about `assert(x' = x)` for the same reason it will analyze `always(true)` as `Pure`. When I refactor the way we propagate effects in #399, I'll think about a way to produce effects without variables so we can say `assert` is always `action`, regardless of its parameters.

The current effect signature for it is`(Read[r]) => Read[r]`. Should we also allow the parameters to have updates?

konnov:
> It doesn't complain about `assert(false)` but it should complain about `assert(x' = x)` for the same reason it will analyze `always(true)` as `Pure`. When I refactor the way we propagate effects in #399, I'll think about a way to produce effects without variables so we can say `assert` is always `action`, regardless of its parameters.

Right. This is why. Is there a way to make the effects checker or the mode checker to complain if `assert(false)` is used in a `def` or `pure def`? Not an urgent question though.

> 
> The current effect signature for it is`(Read[r]) => Read[r]`. Should we also allow the parameters to have updates?

No, we should not allow for updates inside `assert`.

bugarela:
> Is there a way to make the effects checker or the mode checker to complain if assert(false) is used in a def or pure def? Not an urgent question though.

Currently, I believe you could do that by changing its signature to `Read[r] => Read[r] & Update['_asserted<unique_id>']` but that is messy. Let's reconsider this approach in #399 and figure out how to properly handle this case.

bugarela (approved):

PR #552, opened 2023-01-19
add the operator 'fail'

### LINKED ISSUE #543: Introduce an operator to fail a run
[Runs](https://github.com/informalsystems/quint/blob/main/doc/lang.md#runs) are designed for positive tests. Sometimes, we would like to write a test where an action is expected to fail:

```scala
  // test that `send` fails when sending a token with send disabled
  run sendNotEnabledErrTest = {
    init
      .then(send("alice", "bob", Map("photon" -> 3)))
    //.fail()
    // TODO: if `send` above passes, the run should fail.
    // If `send` fails, the test should pass.
    // We need an operator that flips the expected behavior.
  }
```

It should be very easy to implement a run operator that fails a run if and only if the prefix was successful.

konnov:
Closes #543. Introduce the operator `fail` that flips the result of a run from success to failure and vice versa. This operator is useful for writing tests that are expected to fail.

### DIFF
--- a/quint/src/builtin.qnt
+++ b/quint/src/builtin.qnt
@@ -852,6 +852,11 @@ module builtin {
   /// ```
   action repeated(a, n): (bool, int) => bool
 
+  /// `a.fail()` evaluates to `true` if and only if action `a` evaluates to `false`.
+  ///
+  /// This operator is good for writing tests that expect an action to fail.
+  action then(a, b): (bool, bool) => bool
+
   /// `assert(p)` is an action that is true when `p` is true.
   ///
   /// It does not change the state.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+### Added
+
+- Operator `fail` (#552)
+
 ## v0.5.2 -- 2023-01-17
 
 ### Added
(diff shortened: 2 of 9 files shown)


quint/src/builtin.qnt
```diff
@@ -852,6 +852,11 @@ module builtin {
   /// ```
   action repeated(a, n): (bool, int) => bool
 
+  /// `a.fail()` evaluates to `true` if and only if action `a` evaluates to `false`.
```
shonfeder, line 855: I’m a bit surprised fail has to be a built in. Why couldn’t it be defined as a negation?

quint/src/builtin.qnt
```diff
@@ -852,6 +852,11 @@ module builtin {
...
+  ///
+  /// This operator is good for writing tests that expect an action to fail.
+  action then(a, b): (bool, bool) => bool
```
shonfeder, line 858: Shouldn’t it be like

```suggestion
  action fail(a): (bool) => bool
```

shonfeder (approved):

quint/src/builtin.qnt
konnov: Great question. I think in TLA+ it could be defined as `~ENABLED(A)`, that is the condition that `A` is disabled. The issue is that it is not simply a negation, unless you drop all assignments, which is quite hard to do in TLA+ and relatively easy in Quint. Moreover, the more I am looking and actions and runs, the more I believe we need its own type system for action results. It should be either binary like `Executed` and `NotExecuted`. Or better ternary, like `Executed`, `NotExecuted`, `Fail`. There is a tricky distinction between a disabled action and the one that failed, because it has run into some inconsistency inside the action (the typical example is an `assert`).

I have not thought about this much, but it seems that a ternary system would help us to distinguish between preconditions and failures.

PR #554, opened 2023-01-20
Regenerate docs and add that to PR checklist

bugarela:
Hello :octocat: 

I noticed we forgot to update `builtin.md` in https://github.com/informalsystems/quint/pull/552, which was actually a hard thing to remember anyway. So I'm doing that and also adding an entry to our PR template to help us remember to do that in the future. Eventually we can automate this, of course.


konnov (approved):
Oh, that's good! Thanks for adding this to the checklist.

PR #557, opened 2023-01-20
add a few productivity hacks to repl

konnov:
I have added a few productivity tricks to running REPL for loading files:

 - CLI:
   - `quint -r filename.qnt` loads `filename.qnt` right after the start
   - `quint -r filename.qnt::myModule` loads `filename.qnt` right after the start and imports `myModule`
 - REPL:
   - `.load filename.qnt myModule` loads `filename.qnt` and imports `myModule` (I believe this should be replaced with an import from a file in the future)
   - `.reload` clears the history, loads the previously loaded file and imports `myModule` (if it was previously imported).

These hacks should help us in automation of follow-up tutorials.


bugarela (approved):
This is such a productivity improvement! Amazing

PR #546, opened 2023-01-19
Go to Definition functionality to VSCode

bugarela:
Hello :octocat: 

This adds one of my personal favorite features to VSCode: Go To Definition. 

gotodef

- [X] Tests added for any new code
- [X] Ran `npm run format` (or had formatting run automatically on all files edited)
- [X] Documentation added for any new functionality
- [x] Entries added to the respective `CHANGELOG.md` for any new functionality
- [ ] Feature table on [`README.md`](../README.md#roadmap) updated for any listed functionality

PR #562, opened 2023-01-24
Release VSCode extension v0.1.2

bugarela:
### Added

- Support for "Go to definition" (#546).


shonfeder (approved):
looks good to me, but if we are following [semver](https://semver.org/) new functionality (updates to “Added”) should induce a minor version increment (since we are on major 1 here).

Are we aiming to follow semver?

PR #573, opened 2023-01-27
add the work-in-progress example after Solidity's SimpleAuction

konnov:
Adding an interesting example of how Quint could be used to model a Solidity contract called SimpleAuction (from the Solidity intro). This example requires another pass, when we implement a few missing features (e.g., constant substitution). However, it is already a valuable example, and it can help new users in faster onboarding.

Every smart contracts turns into an interesting case study. So I'd like to merge this one before running out of time budget, spent 6 hrs on this one.


bugarela (approved):
I've only skim through the specification, let me know if you'd like a more careful review. This seems great, the test runs are awesome!

We should think more about this strategy of using records to define the state. It makes the lack of unchanged helpers less annoying, but also takes power away from the effects checker (i.e. there's no checking for multiple or unbalanced record updates). I feel like if we ever find a nice way to compose states/specs and embed it into the language, then we won't need to resort to this. Just something I've been keeping in mind.

konnov:
> We should think more about this strategy of using records to define the state. It makes the lack of unchanged helpers less annoying, but also takes power away from the effects checker (i.e. there's no checking for multiple or unbalanced record updates). I feel like if we ever find a nice way to compose states/specs and embed it into the language, then we won't need to resort to this. Just something I've been keeping in mind.

Yep, we definitely have to understand the emerging idioms :)

PR #592, opened 2023-01-31
Fix: REPL Output Color In Dark Terminal

### LINKED ISSUE #574: Improve REPL color contrast
REPL output is hardly readable with a dark terminal theme:

Screen Shot 2023-01-27 at 14 16 00

I think the solution would be not to explicitly color delimiters like `,`, `(`, and `)` using `chalk.black` – instead, just use the default foreground terminal color.

HosamIntel:
repl color output (1)

removed the chalk.black so that the terminal would use the default foreground

resolves #574


konnov (approved):
That's a very neat and small fix. Thanks a lot for taking care of this issue!

PR #603, opened 2023-02-02
require EOF after modules

### LINKED ISSUE #585: Parser doesn't flag errors at end of file
If you write

```scala
module foo {
}

any
invalid
quint

module bar {
}
```

the parser will only parse the module `foo` and ignore the rest, without flagging any errors. I just tried doing:

```scala
module foo {
  ...
}

/// Module bar is really cool
module bar {
  ...
}
```

and spent a bunch of time debugging while things inside module bar weren't working correctly in the plugin: they weren't being parsed at all.

Note: we should also support docstrings for modules. Right now, triple-slashed comments before modules like this one are invalid syntax.

cc @konnov

konnov:
Closes #585. This PR simply extends the Quint grammar with EOF after `modules`.


shonfeder (approved):
Looks great!

PR #604, opened 2023-02-02
produce runtime errors on unsupported operators

### LINKED ISSUE #591: REPL shows an error when a spec contains temporal operators
I have never tried it before. REPL does not compile a spec that contains temporal operators, even if they are never used in REPL, e.g.:

```sh
$ quint
>>> always(true)
compile error: <input>:1:1 - error: Called unknown operator always
1: always(true)
   ^^^^^^^^^^^^


>>> eventually(true)
compile error: <input>:1:1 - error: Called unknown operator eventually
1: eventually(true)
   ^^^^^^^^^^^^^^^^
```

We should issue a warning instead and show an error only if the user is trying to evaluate an expression that involves a temporal operator.

konnov:
Closes #591. REPL compiles unsupported operators, but produces runtime errors when they are actually evaluated.

- [x] Regenerated documentation with `quint docs` if `builtin.qnt` was changed


shonfeder (approved):
LGTM :smile_cat:

PR #606, opened 2023-02-02
add better error handling

### LINKED ISSUE #444: Confusing error message: `undefined not found in undefined`
To reproduce:

```
>>> val x = 3
>>> x' = 1
compile error: <input>:2:1 - error: undefined not found in undefined <- ...
2: x' = 1
   ^^^^^^
```
Note that I'm not sure why this is illegal/unexpected; seems like an opportunity to teach the user why this is wrong (and how to fix it?)

konnov:
Closes #444. Add a less confusing error messaging for the example:

```
val x = 1
x' = 2
```

It's actually surprising that this code made it to REPL without an error by the mode checker. I guess, this is one of the known issues, @bugarela? 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
  - The parser complains about junk in the end of file (#603)
+ - Fix a confusing error message in assignments (#606)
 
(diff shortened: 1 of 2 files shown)


shonfeder (approved):
:rocket:

bugarela:
@konnov It is a direct consequence of the propagation issue, but I was not aware that it could impact updates like this. Fortunately, fixing that should be enough to fix this case.

Issue #502, opened 2023-01-12
Add short explanations to error messages for effect errors


shonfeder:
As we discussed in our meeting today, it would help with usability if we can provide some information explaining why and effect check fails in the error messages for effect errors. For example, if a `pure` operator fails because it is writing to a state variable, perhaps we could say something like

> pure operators may not write to state variables, but operator foo rights to variable x in ...

PR #619, opened 2023-02-09
Improve mode error reporting

### LINKED ISSUE #620: Remove `'pred'` from `OpQualifier`
We have not been using the `'pred'` qualifier for long time

bugarela:
Hello :octocat: 

This makes mode errors better explained and, in the VSCode plugin, they come with a "Quick Fix" action to correct the mode automatically.

modereplace

I also introduced a new error structure that is more general and has better compatibility with VSCode diagnostics (which seem like a great error representation). This is a first iteration towards a refactor in the error representation, and additional steps are required to make it our standard representation (replacing `ErrorMessage` and `ErrorTree`).

The errors have a `data` field to store additional data, including a possible fix for the problem - which can be used by VSCode to edit the document and fix the error, or printed in the CLI as a suggestion (TODO). 

Towards #502
Closes #620 


quint/src/effects/modeChecker.ts
```diff
@@ -107,12 +129,19 @@ class ModeFinderVisitor implements EffectVisitor {
...
       this.currentMode = 'temporal'
-    } else if (anyVariablesAdded(paramUpdates, r.update)) {
+      this.explanation = `performs temporal operations over variables ${addedTemporal.map(variablesToString)}`
+    } else if (addedUpdates.length > 0) {
       this.currentMode = 'action'
-    } else if (anyVariablesAdded(paramReads, r.read)) {
+      this.explanation = `updates variables ${addedUpdates.map(variablesToString)}`
+    } else if (addedReads.length > 0) {
       this.currentMode = 'def'
+      this.explanation = `reads variables ${addedReads.map(variablesToString)}`
```
shonfeder, line 144: Why do with these explanations need to be set by a mutable variable? In both of these two blocks where the explanations are formed, the text of the explanation is repeated verbatim and, iiuc, each message appears to be a function of two things:

- the current mode (e.g., `temporal`)
- the effect (from which the relevant set of variables are derived)

It looks to me as if both of these values are available at the point at which the final error message is constructed on line 68, so I'm wondering why we cannot computer along with the `modeConstraint`. On the other hand if we do need an additional mutable very able to carry some information here, it looks like we only really need the  set of variables.

WDYT? Am I missing something, or could this be cleaned up a bit?

quint/src/effects/modeChecker.ts
```diff
@@ -65,27 +65,49 @@ export class ModeChecker implements IRVisitor {
 
     if (generalMode === mode) {
       this.errors.set(def.id, {
-        location: `Checking modes for ${definitionToString(def)}`,
-        message: `Expected ${mode} mode, found: ${qualifierToString(def.qualifier)}`,
-        children: [],
+        code: 'QNT200',
+        message: `${qualifierToString(def.qualifier)} operators ${modeConstraint(def.qualifier)}, but operator \`${def.name}\` ${this.modeFinderVisitor.explanation}. Use ${qualifierToString(mode)} instead.`,
+        data: { fix: { kind: 'replace', original: qualifierToString(def.qualifier), replacement: qualifierToString(mode) } },
```
shonfeder, line 70: Since we now have these unique IDs that need to be given to every distinct kind of error, I think it would be a good idea to gather them all into one file. Otherwise I don't see how we will avoid conflict between error IDs, nor will we have a way of surveying all the error we report.

At minimum, I think this would mean defining a union type to enumerate all of the error IDs. But even better IMO would be functions that construct the relevant error given the needed params, tho i can see how this might get a bit unwieldy.

quint/src/quintError.ts
```diff
@@ -0,0 +1,48 @@
...
+ * --------------------------------------------------------------------------------- */
+
+/**
+ * Representation for quint errors
+ *
+ * @author Gabriela Moreira
+ *
+ * @module
+ */
+
```
shonfeder, line 14: this cold be a good place to enumerate the error IDs/error constructors?

quint/test/repl.test.ts
```diff
@@ -143,7 +143,7 @@ describe('repl ok', () => {
     )
     const output = dedent(
       `>>> 1 + false
-      |static analysis error: <input>:1:1 - error: Couldn't unify int and bool
+      |static analysis error: <input>:1:1 - error: QNT000: Couldn't unify int and bool
```
shonfeder, line 146: We have a lot of colons piling up here, might something like be more standard/reabable:

```
      |static analysis error: <input>:1:1 - error [QNT000]: Couldn't unify int and bool
```

I'm not sure here, and this is low stakes. Just a thought.

shonfeder (commented):
looks great! super cool to see the error becoming standardized and the basis for automated fixes!

a few questions/suggestions on code organization.

quint/src/effects/modeChecker.ts
bugarela: I was writing a big response to this, but I actually think that we don't even need the visitor with the mutable `currentMode` at all. I've changed the behavior of this since I first wrote it and now it seems like we don't need to look at nested effects anymore, only the top-level definition effect :smile: that will simplify things a lot, thank you for the thoughtful observation.

quint/src/effects/modeChecker.ts
bugarela: We have some error enumeration at ADR 2, but I don't think that will scale well. I have an idea from your suggestions: we add a `ErrorCode` type with enumerated values `'QNT000' | 'QNT001' | ...` and add docstrings to document each of the possibilities. This way, we have some enforcement without adding a big overhead of constructors. WDYT?

shonfeder: sounds good!

quint/src/effects/modeChecker.ts
shonfeder: oh very nice! Look forward to seeing the new code :)

quint/test/repl.test.ts
bugarela: I did this instead:
```
      |static analysis error: <input>:1:1 - error: [QNT000] Couldn't unify int and bool
```

quint/src/effects/modeChecker.ts
bugarela: It's much better now! I didn't extract the explanation to a different function because I feel like when I read the explanation along with the logic, one thing helps understanding the other.

quint/src/quintError.ts
```diff
@@ -0,0 +1,74 @@
...
+  /* The error code in the form QNTnnn */
+  code: ErrorCode,
+  /* The error explanation */
+  message: string,
+  /* Additional data for the error */
+  data: QuintErrorData
+}
+
+export type ErrorCode =
+  /* Placeholder for errors that have not being assigned a proper code yet */
```
shonfeder, line 28:
```suggestion
  /* Placeholder for errors that have not been assigned a proper code yet */
```

shonfeder (approved):
Looks great to me!

quint/src/effects/modeChecker.ts
shonfeder: makes sense1

Issue #618, opened 2023-02-09
Refactor REPL to work incrementally


konnov:
Currently, REPL is parsing and type checking the whole input on every new expression. This introduces plenty of computation overhead on large modules. We should cache the parsed and type checked results instead and process only the new parts.

PR #621, opened 2023-02-13
Refactoring REPL and the compiler to avoid nested modules

### LINKED ISSUE #305: Simulator should clean up lookup tables
In the example below, the simulator should compute the value of `x`.

```sh
$ tntc repl
TNT REPL v0.0.2
>>> .load t.tnt
module A {
  var x: int
}

module B {
  val x = 3
}

true
>>> import B.*

>>> x
<result undefined>

>>> 
```

### LINKED ISSUE #426: REPL session is not a proper Quint file
Currently, REPL dumps all definitions a the top-level, whereas a proper Quint spec should be wrapped into a module. We should save REPL sessions in a predefined module, e.g., `__repl`. Then, on `.load` we would automatically import all definitions from `__repl`.

### LINKED ISSUE #550: Refactor REPL to avoid using nested modules
Since we are going to remove support for nested modules in #548, we have to refactor REPL to avoid using them.

konnov:
- [x] Closes #550.
 - [x] Closes #305.
 - [x] Closes #426.

This PR contains refactoring of the REPL and compiler code, to achieve the following:

 - The JS compiler is using lookup tables and types now, instead of using its own lookup stack.
 - REPL supports several top-level modules. By default, `__repl__` is the main module.
 - The compiler is (mostly) using expression identifiers instead of names. This should help us in avoiding name clashes in the future.

Making REPL incremental is postponed for #618. This probably requires some refactoring to the parser, name resolver, and type checker.


quint/src/runtime/compile.ts
```diff
@@ -59,46 +66,92 @@ function errorContext(errors: ErrorMessage[]): CompilationContext {
...
+  if (!moduleTable) {
+    return left(`Module ${moduleName} not found`)
+  }
+  const defs = moduleTable.valueDefinitions.get(defName)
+  if (!defs) {
+    return left(`Definition ${moduleName}::${defName} not found`)
+  }
+  if (defs.length !== 1) {
+    return left(`Multiple definitions (${defs.length}) of ${moduleName}::${defName} found`)
+  }
```
bugarela, line 93: Should this be using `lookupValue`? Although there's no real encapsulation, we should only manipulate `lookupTable` with the accompanying functions.

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -116,13 +142,16 @@ export class CompilerVisitor implements IRVisitor {
   }
 
   exitOpDef(opdef: ir.QuintOpDef) {
-    // either a runtime value (if val), or a callable (if def, action, etc.)
+    // Either a runtime value, or a def, action, etc.).
```
bugarela, line 145:
```suggestion
    // Either a runtime value, or a def, action, etc.
```

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -38,12 +43,21 @@ import { lastTraceName } from '../compile'
...
  */
 export class CompilerVisitor implements IRVisitor {
+  // the id of the current module
+  private moduleId: bigint = 0n
+  // the lookup table to use for the module
+  private lookupTable: LookupTable = newTable({})
+  // the scope tree to be used with the lookup table
+  private scopeTree: ScopeTree = { value: 0n, children: [] }
+  // types assigned to expressions and definitions
+  private types: Map<bigint, TypeScheme>
```
bugarela, line 53: Are we using this value for something?

bugarela (approved):
LGTM, just a few comments

quint/src/runtime/compile.ts
konnov: I have been thinking about that. The issue is that we do not have a scope id in this case. So I will keep it as it is atm. But it would be good to come back to the lookup API at some point.

quint/src/runtime/impl/compilerImpl.ts
konnov: Interestingly, not :)

We would need it for higher-order definitions though. So I would keep it

PR #626, opened 2023-02-14
add two code snippets on the landing page

konnov:
So people can see that:

 - Quint is different from TLA+,
 - It looks more like a programming language

~It would be great to have actual highlighted code instead of images. But images were the easiest to add quickly.~


README.md
```diff
@@ -10,6 +10,20 @@ distributed systems and blockchain protocols. It combines the robust theoretical
 basis of the [Temporal Logic of Actions][TLA] (TLA) with state-of-the-art static
 analysis and development tooling.
 
+This is how a typical Quint code looks like in the [Quint VSCode Plugin][]:
+
+<p align="center">
+  <img src="./doc/img/validate-quint.png"
+       width="70%" title="Code snippet in Quint">
+</p>
```
shonfeder, line 18: This ends up being very tiny on mobile:

image

I wonder if it wouldn't be better to just use a codeblock here? I know we don't
have syntax highlighting on GitHub, but we could rig up a landing page using JS
highlighting once our priorities shift to publicity. This would give us the
benefit of accessibility and responsive scaling, at the cost of some superficial
color for the time being. WDYT?

README.md
```diff
@@ -10,6 +10,20 @@ distributed systems and blockchain protocols. It combines the robust theoretical
...
+
+If you would to see the same code in TLA<sup>+</sup>, here is how it looks like
+in the [TLA+ VSCode Plugin][]:
```
shonfeder, line 21:
```suggestion
If you would to see the same code in TLA<sup>+</sup>, here is how it looks
in the [TLA+ VSCode Plugin][]:
```

I also think the link here might be missing?

README.md
konnov: That's what I also thought. But I am not sure how to embed HTML code with JS into markdown. Github restricts HTML in Markdown.

shonfeder: yeah, afaik, we can't to highlight.js or whatever on github's markdown rendering, but we could of course on a static site hosted from our gh page.

README.md
```diff
@@ -10,6 +10,28 @@ distributed systems and blockchain protocols. It combines the robust theoretical
 basis of the [Temporal Logic of Actions][TLA] (TLA) with state-of-the-art static
 analysis and development tooling.
 
+This is how a typical Quint code looks:
```
shonfeder, line 13:
```suggestion
This is how typical Quint code looks:
```

README.md
```diff
@@ -10,6 +10,28 @@ distributed systems and blockchain protocols. It combines the robust theoretical
...
+```scala
+  // `validateBalance` should only be called upon genesis state.
+  pure def validateBalance(ctx: BankCtx, addr: Addr): bool = and {
+    ctx.accounts.contains(addr),
+    val coins = getAllBalances(ctx, addr)
+    coins.keys().forall(denom => coins.get(denom) > 0),
+  }
+```
+
+If you would to see the same code in TLA<sup>+</sup>, here is how it looks:
```
shonfeder, line 24:
```suggestion
If you would like to see the same code in TLA<sup>+</sup>, here is how it looks:
```

shonfeder:
yup! looks good :)

shonfeder (approved):
- [ ]

PR #634, opened 2023-02-16
Introduce the testing framework

### LINKED ISSUE #628: Implement the command `test`
Similar to #312, we should implement the command [test](https://github.com/informalsystems/quint/blob/main/doc/quint.md#command-test).

It is higher priority than #312, as the users are already able to run specs in REPL. We need the test command to run the integration tests.

konnov:
Closes #628.

This PR introduces the command `test` that mimics the behavior of a unit-testing framework, e.g., of mocha. This command should let us integrate basic tests for protocols in the CI.

### DIFF
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -237,9 +237,14 @@ true
 
 ### Tests works as expected
 
+The command `test` finds failing tests and prints error messages.
+
 <!-- !test in test runs -->
 ```
-quint test --main counters --seed 1 ../examples/language-features/counters.qnt 2>&1 | sed 's/([0-9]*ms)/(duration)/g'
+quint test --main counters --seed 1 \
+  ../examples/language-features/counters.qnt 2>&1 | \
+  sed 's/([0-9]*ms)/(duration)/g' | \
+  sed 's#^.*counters.qnt#      HOME/counters.qnt#g'
 ```
 
 <!-- !test out test runs -->
@@ -252,6 +257,12 @@ quint test --main counters --seed 1 ../examples/language-features/counters.qnt 2
   1 passing (duration)
   1 failed
   1 ignored
+
+  1) failingTest:
+      HOME/counters.qnt:84:9 - error: Assertion failed
+      84:         assert(n == 0),
+                  ^^^^^^^^^^^^^^
+      
 
 ```
 
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -251,7 +251,7 @@ quint test --main counters --seed 1 \
 ```
 
   counters
-    ok  passingTest
+    ok passingTest
     1) failingTest
 
   1 passing (duration)
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -235,3 +235,23 @@ true
 >>> 
 ```
 
+### Tests works as expected
+
+<!-- !test in test runs -->
+```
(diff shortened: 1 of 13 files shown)


quint/src/runtime/compile.ts
```diff
@@ -102,17 +102,67 @@ export function
...
+          mainName: string): CompilationContext {
+  // Push back the main module to the end
+  const main = modules.find(m => m.name === mainName)
+  // when the main module is not found, we will report an error
+  const mainNotFound =
+    main ? [] : [{
+      explanation: `Main module ${mainName} not found`,
+      refs: [],
+    }]
+  const reorderedModules =
```
bugarela, line 131: I think this comment should be further down

```suggestion
  const main = modules.find(m => m.name === mainName)
  // when the main module is not found, we will report an error
  const mainNotFound =
    main ? [] : [{
      explanation: `Main module ${mainName} not found`,
      refs: [],
    }]
  // Push back the main module to the end
  const reorderedModules =
```

bugarela: Also, it would be nice to have an explanation why it has to be pushed to the end

quint/src/runtime/testing.ts
```diff
@@ -0,0 +1,92 @@
...
+          }
+        } else {
+          // produce errors
+          const errors = ctx.runtimeErrors
+          // we do not want to see the same errors again
+          ctx.runtimeErrors.length = 0
+          return { name, status: 'failed', errors }
+        }
+      })
+  }))
```
bugarela, line 90: We can avoid this deep nesting by using some guard clauses instead of `else`:

```suggestion
  return merge(testDefs.map(def => {
    const name = def.name
    return contextLookup(ctx, main.name, name, 'callable')
      .map(comp => {
        const result = comp.eval()
        if (result.isNone()) {
          // produce errors
          const errors = ctx.runtimeErrors
          // we do not want to see the same errors again
          ctx.runtimeErrors.length = 0
          return { name, status: 'failed', errors }
        }
  
        const ex = result.value.toQuintEx()
        if (ex.kind !== 'bool') {
          return { name, status: 'ignored', errors: [] }
        }
        if (ex.value) {
          return { name, status: 'passed', errors: [] }
        }
  
        const e = {
          explanation: `${name} returns false`,
          refs: [def.id],
        }
        return { name, status: 'failed', errors: [e] }
      })
  }))
```

quint/io-cli-tests.md
```diff
@@ -235,3 +235,23 @@ true
...
+```
+
+  counters
+    ok  passingTest
+    1) failingTest
+
+  1 passing (duration)
+  1 failed
+  1 ignored
+
```
bugarela, line 255: Why don't we have any explanation for this failure? Is the explanation included only when there are parsing/typechecking errors?

bugarela (approved):
The output looks really nice! Just a couple of suggestions and a question

quint/src/runtime/compile.ts
konnov: That's a good point. I have restructure the whole function body. There is no need for compiling anything if the the module is not found.

quint/src/runtime/testing.ts
konnov: That's much better! Thanks a lot for clearing my spaghetti code there :)

quint/io-cli-tests.md
konnov: That's a good point. I have to check that

konnov: Oh, it's my bad typescript fu. I copied an array by reference and then emptied it immediately %)

PR #639, opened 2023-02-18
add --with-lookup to parse

konnov:
A small add-on that add the option `--with-lookup` to the command `parse`. When this option is passed, the JSON output contains the lookup table, referred by the attribute `table`.


quint/src/cliCommands.ts
```diff
@@ -37,11 +37,14 @@ export type stage =
 /** The data from a ProcedureStage that may be output to --out */
 interface OutputStage {
   stage: stage,
+  // the modules and the lookup table produced 'parse'
```
shonfeder, line 40:
```suggestion
  // the modules and the lookup table produced by 'parse'
```

shonfeder (approved):

PR #641, opened 2023-02-20
New effect representation (and error messages)

bugarela:
Hello :octocat: 

This refactors the representation of effects and variables, and changes error reporting for multiple updates of the same variable to reduce the noise of unification details and proper highlight the meaningful expression (as oppose to the entire `all` expression in the image example)

image

This PR:
1. Changes concrete effects from a ` { ..., read: [], update: [], temporal: [] }` representation to a list of components that can have kinds `'read' | 'update' | 'temporal'`. This will help us introduce runs an nondet with much less impact.
2. Adds an id to the representation of a state variable in an effect, in order to track the origin expression where it appears and properly report errors.
3. Removes the multiple update checking at unification time, and instead, check for multiple updates after all effect inference is done in a new class called `MultipleUpdatesChecker`, which is a new step in the static analysis phase. You may notice several removed tests that covered the old behavior - all of these is now covered in `MultipleUpdatesChecker` unit tests.

Towards #502 and #399


quint/src/effects/MultipleUpdatesChecker.ts
```diff
@@ -0,0 +1,70 @@
...
+    const updateVariables = e.components.reduce((updates: Variables[], c) => {
+      if (c.kind === "update") {
+        updates.push(c.variables);
+      }
+      return updates;
+    }, [])
+
+    const vars = findVars({ kind: 'union', variables: updateVariables })
+
+    const repeated = vars.filter(v => vars.filter(v2 => v.name === v2.name).length > 1)
```
konnov, line 48: I wonder, whether it would be more optimal to sort the array first and then check for two adjacent duplicates? For small values of $n$ e.g., from 1 to 50, quicksort would be essentially linear, as it is $O(n \cdot log\ n)$ in general.

quint/src/effects/modeChecker.ts
```diff
@@ -109,18 +109,34 @@ function modeConstraint(mode: OpQualifier): string {
...
+])
+const modesForArrow = new Map<ComponentKind, OpQualifier>([
+  ['temporal', 'temporal'],
+  ['update', 'action'],
+  ['read', 'def'],
+])
+const modesForConcrete = new Map<ComponentKind, OpQualifier>([
+  ['temporal', 'temporal'],
+  ['update', 'action'],
+  ['read', 'val'],
```
konnov, line 126: why does 'read' correspond to `def` in one case, and to `val` in another? What happens if pass another operator as an argument?

konnov (approved):
Looks good to me! The code looks easier to understand now!

quint/src/effects/MultipleUpdatesChecker.ts
bugarela: Sure, but also this is an iteration over state variable updates, and there shouldn't be reasonable specs with more than 20 of these, so I wouldn't sacrifice readability.

Perhaps this solution from stack overflow using lodash is linear (I haven't looked at the source code for lodash to be sure):
```ts
_([1, 1, 2, 2, 3]).groupBy().pickBy(x => x.length > 1).keys().value()
```

quint/src/effects/modeChecker.ts
bugarela: Arrow effects will have `def` mode, since the operator takes arguments, while a concrete effect will have `val`. I don't see how passing another operator as an argument would affect that, can you elaborate?

quint/src/effects/MultipleUpdatesChecker.ts
konnov: Yep, `groupBy` should be similar, just using a map instead of sorting.

quint/src/effects/modeChecker.ts
konnov: Maybe I am mixing things up with modes. What happens if we have a `def` that accepts an operator as an argument and that operator happens to be `temporal`?

bugarela: If the `def` actually uses that operator in its body, it would produce a temporal effect in its body and an error would be reported saying `def` should be `temporal`.

This is what I believe should happen, I'll test it later to make sure this is working correctly.

bugarela: Ok, I'm testing this and actually there's no way to strict the mode of an operator given as argument since we don't allow effect signatures. This is what I wrote:

```scala
  def applyTemporal(f: (int) => bool): bool = {
    f(1)
  }

  temporal a = applyTemporal(x => always(x > 1))
```

There's no way to tell whether `f` is temporal or not, and there is no way to strict it as well.

This is bad, as we can also use this to apply action operations:

```ocaml
  action b = applyTemporal(x => z' = 1)
```

Perhaps we should enforce that no high order operators can have those modes in their arguments? It's not clear for me what consequences could arise from that, it might prevent users from doing something that they should be able to. We can try!

Either way, is it ok if I open a separate issue for this and we move on here? This is not a change introduced by this refactoring.

Issue #567, opened 2023-01-25
Mode checker loses track of parameters in nested definitions


konnov:
I am trying the following piece in the VSCode plugin:

```
    pure def foo(x: int): int = {
        pure val y = x + 1
        y
    }
```

The mode checker is complaining about the definition of `y`. On the other hand, the following surprisingly works:

```
    pure def foo(x: int): int = {
        val y = x + 1
        y
    }
```

Is it a bug, @bugarela?

bugarela:
The first case not working is a bug. The second one should work fine, with a linter suggestion pointing out that you could add `pure` (but that won't happen right now since the modeChecker thinks the correct mode is `val`).

The problem is that the mode checker does not know that `x` is a parameter when it is looking inside the definition of `y`. I don't know how to solve this atm, I might have to rethink how we define effect variables for parameters at all.

Thanks for reporting :smile:

bugarela:
This issue is blocking #646 since it breaks a lot of integration tests once we properly propagate temporal effects everywhere we should. Therefore, I'm prioritizing this.

PR #646, opened 2023-02-21
Generalize effect propagation + propagate temporal

bugarela:
Merging blocked by #567 


quint/src/effects/builtinSignatures.ts
```diff
@@ -12,175 +12,241 @@
...
+ *
+ * Used in operators like `map`, `filter` and `fold`
+ *
+ * Example: propagationWithLambda(['read', 'temporal'])(2) results in
+ * (Read[r1] & Temporal[t1], (Read[r1] & Temporal[t1]) => Read[r2] & Temporal[t2]) => Read[r1, r2] & Temporal[t1, t2]
+ *
+ * @param kinds the kinds of components to propagate
+ * @returns an arrow function taking arity and returning the effect for that arity
+ */
+function propagationWithLambda(kinds: ComponentKind[]): ((arity: number) => EffectScheme) {
```
shonfeder, line 81: It looks to me like these functions are helpers for generating specific effect signatures, rather than any sort of effect propagation inference of generator etc. With that understanding, a name like

```suggestion
function effectSignatureWithLambda(kinds: ComponentKind[]): ((arity: number) => EffectScheme) {
```

would be a bit clearer to me. That said, I may be misunderstanding, and this isn't a substantive concern for me.

quint/src/effects/builtinSignatures.ts
```diff
@@ -12,175 +12,241 @@
...
+    })
+
+    const result: Effect = { kind: 'concrete', components: resultComponents }
+    const effect: Effect = { kind: 'arrow', params: params, result: result }
+
+    return { effect, ...effectNames(effect) }
+  }
+}
+
+const p = parseAndQuantify
```
shonfeder, line 66: This abbrev. seems ex. shrt.; poss. to mk. l.? :stuck_out_tongue: 

The abbreviation seems too cryptic to me given the specificity of the value and how big its scope is. Do you think something a bit more descriptive could be found?

shonfeder (commented):
Looks like a good simplification! I've a few small suggestions you might consider, but nothing blocking.

shonfeder (approved):

quint/src/effects/builtinSignatures.ts
bugarela: Hmm, I'm not sure. My intention with "propagation" here is that it defines an effect that propagates the given components. `effectSignatureWithLambda` is more generic, as a lot of different signatures like that can be defined. A signature that does propagation (that is, combines the effect of the parameters in the result) is much more specific. I'm also only using these helpers for operators that do full propagation, and there are some effects that are still defined manually.

quint/src/effects/builtinSignatures.ts
bugarela: I intentionally used the shortest possibility to clean out the rest of the file since this is used a lot, and frequently with long strings that take a lot of line space. I can probably achieve the same with indentation, I'll try.

PR #647, opened 2023-02-21
introduce idGenerator and globalIdGen

### LINKED ISSUE #138: Introduce an id generator
When we construct Quint expressions, we have to assign a unique id. Currently, fresh identifiers are only generated in `ToIrListener`, that is, in the parser.

We are producing new Quint expression in the simulator, when converting JS objects back to Quint. Hence, we need a reliable way of generating unique identifiers. It can be as simple as having a function that keeps incrementing an internal counter. However, we should understand how atomicity and thread-safety works in JS. That's why I have opened a separate issue.

konnov:
Closes #138 and unblocks #618. This PR introduces:
 
 - an interface of an `IdGenerator`,
 - a sequential implementation of `IdGenerator` that is instantiated with `newIdGenerator()`,
 - the default global id generator that is accessible via `globalIdGen`.
 - In addition, it finally assigns unique ids to the expressions that are produces from the runtime JS values.

By default, the parser is using the global id generator now. This would be useful when we incrementally call the parser. All tests are creating a new id generator, as we assume specific id values.


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Mode errors are better explained and include a fix suggestion (#619)
 - Typechecking now reports errors when instance overrides are not compatible with
   the original definition (#622)
+- the parser and tests are using the id generator now (#647)
```
shonfeder, line 20: Is this a user-facing breaking change? Seems like just an internal optimization?

quint/src/quintParserFrontend.ts
```diff
@@ -97,7 +101,7 @@ export function parsePhase1(text: string, sourceLocation: string): ParseResult<P
     return left(errorMessages)
   } else {
     // walk through the AST and construct the IR
-    const listener = new ToIrListener(sourceLocation)
+    const listener = new ToIrListener(sourceLocation, idGen)
     ParseTreeWalker.DEFAULT.walk(listener as QuintListener, tree)
```
shonfeder, line 105: Why make the id generator global but still have to thread it thru all the classes as a parameter? Seems a bit like the worst of both worlds :smile:

shonfeder (commented):
Looks straight forward! But can we do without the global mutable state and just stick to the dependency injection that you're using here in any case?

CHANGELOG.md
konnov: I hope it's not breaking anything, but if something breaks, we could track it by looking at the changelog, or the commit log. I am not sure.

quint/src/quintParserFrontend.ts
konnov: I would prefer it to be a parameter in the most cases. But there a few places, where it is hard to carry the id generator around (e.g., runtime values). That's why the global variable is there.

CHANGELOG.md
shonfeder: If we are following “keep a change log”, changes we shouldn’t be breaking and don’t change the api at all aren’t really meant for the changelog I think?

> The purpose of a commit is to document a step in the evolution of the source code. Some projects clean up commits, some don't.
> 
> The purpose of a changelog entry is to document the noteworthy difference, often across multiple commits, to communicate them clearly to end users.

https://keepachangelog.com/en/1.0.0/

so this seems pretty clearly like a commit thing to me. Let’s be sure we don’t break things, by adding new tests if needed? But i worry adding stuff as breaking changes because it *might* break something in some unforeseen  way just makes that part of the changelog unreliable and noisy.

End users don’t need to know about refactoring :)

konnov: sure, let's remove it then

konnov, line 20:
```suggestion
```

konnov: removed

quint/src/quintParserFrontend.ts
konnov: Also, we have to override the generator when doing tests. So we have the global generator to server as a common source of identifiers for various components. But the components themselves should be able to consume various generators.

bugarela (changes requested):
If that's ok, I'd like to discuss this in tomorrow's meeting before merging. Ideally I should start a discussion on this PR, but I've been all-in in the improvements of the effect system and haven't got the brain space to process this.

konnov:
We have not discussed this today. IMO this is a simple PR, so it would be good to merge it.

bugarela (commented):
I really don't like the introduction of a global mutable counter, this is like the first bad practice I ever learned and there are tons of reasons to avoid it.

Seems like there are three motivations for it:
1. Generating different IDs in different runtime values (I don't get why exactly)
2. Incremental parsing
3. Flattening

For (1), can't we store the id of the original expressions in the runtime value and get it back when we are converting again, or something like that? Are we using this new values for something?

For (2) and (3), I believe we could go with an easier form of composing different parsing runs with themselves and flattening, such as including the `lasId`/`biggestId` in the `QuintModule` representation and then providing that as a parameter to the next parsing run or flattening procedure.

shonfeder:
iiuc, for the time being we could just drop the global ID generator and require an ID generator be supplied thru the constructor https://github.com/informalsystems/quint/pull/647/files#diff-6f59f39b1cee888b5d07c01bbbba59ee7dbc42eaf407ac71054d16f3d49ab6f9R371-R373

Would this let us do everything we need at the moment, and postpone the question of whether we need the global mutability?

konnov:
I have removed the global id generator and added the code to propagate the generator, whenever needed. This seems to work, though the diff is obviously larger. At some point, we would have to think how to deal with the growing number of parameters that are passed around.

bugarela (approved):
Thanks for removing the global id generator! I'm unblocking.

I'm still wondering whether we could re-use the ids from the original expressions in `RuntimeValue` instead of generating new ones. Have you considered this?

konnov:
> Thanks for removing the global id generator! I'm unblocking.
> 
> I'm still wondering whether we could re-use the ids from the original expressions in `RuntimeValue` instead of generating new ones. Have you considered this?

That's an interesting idea. However, I don't know which ids would be considered the original ones. There are plenty of computations going on with runtime values. Moreover, if we start using ids of the original expressions, they stop being unique. Not sure how much that would break.

PR #649, opened 2023-02-21
Fix the tutorial after the comments by @pdini

konnov:
This PR fixes a few issues noticed by @pdini:

> Step 3
> Just a small thing: you write [type], but in the code meta statement you use [tp]. Maybe the same form should be used?

> Step 5
> What is the difference between a pure value and a const? The reason I ask is that in ASMs they are both static functions, where the const is a 0-ary static function. Interestingly, your MAX_UINT is a 0-ary static function, but you call it a pure value rather than a const. Therefore, there must be some difference between these two constructs that I do not see yet.

> Step 6
> You mention operators, but they have not been defined yet. I suggest adding "(defined in the next step)" after "In contrast to operator definitions".

> Step 6/7
> I am surprised that the variable balances is not a state variable. The reason is that the signature (Addr -> UInt) implies a dynamic function, i.e. the balance UIint at a given address may change in different states.

> Step 8
> Not clear why minter affects the totalSupply 0-ary dynamic function. Maybe you could add a few words of explanation to the first bullet?


tutorials/lesson3-anatomy/coin.md
```diff
@@ -144,27 +144,28 @@ The main property of pure values is that they always return the same value.
 Pure definitions always return the same value, if they are supplied with the
 same arguments.
 
-To see that no state is needed, evaluate these definitions in REPL:
+To see that no state is needed, evaluate these definitions in REPL
+(read-evaluate-print-loop):
```
shonfeder, line 148:
```suggestion
To see that no state is needed, evaluate these definitions in the REPL
(read-evaluate-print-loop):
```

tutorials/lesson3-anatomy/coin.md
```diff
@@ -213,9 +214,20 @@ still make sense. The typical parameters are:
  - timeout values.
 
 For this purpose, Quint offers `const` declarations. You can see one of them
-in the commented out section of the code above. Constant declarations are not
-fully supported yet. They will be available as soon as the issue
-[#528](https://github.com/informalsystems/quint/issues/528) is closed.
+in the commented out section of the code above. You may be wondering, what is
+the difference between `const` and `pure val`. They mean to express the same
+concept: A value that stays the same for all computations. However, they differ
+in the time when they are bound to a value:
```
shonfeder, line 220: I think a tweak along these lines might remove some possible ambiguity:

```suggestion
the difference between `const` and `pure val`. Both declare a value that stays the same throughout the execution of a computation, but the time at which they are bound to values differs:
```

tutorials/lesson3-anatomy/coin.md
```diff
@@ -213,9 +214,20 @@ still make sense. The typical parameters are:
...
+
+ - The `pure val` values are immediately defined via an expression in the
+   right-hand side.
+
+ - The `const` values are first declared and later they are substituted
+   with actual values via an `instance` declaration.
+
+Constant declarations are not fully supported yet. They will be available
+as soon as the issue [#528](https://github.com/informalsystems/quint/issues/528)
+is closed.
```
shonfeder, line 230:
```suggestion
**Note:** Constant declarations are not fully supported yet. They will be available
as soon as the issue [#528](https://github.com/informalsystems/quint/issues/528)
is closed.
```

tutorials/lesson3-anatomy/coin.md
```diff
@@ -246,9 +258,8 @@ a state of the protocol. In the above code, we introduce two such variables:
    The type of this variable is `Addr -> UInt`, which means a map from
    values of type `Addr` to values of type `UInt`.
 
-In contrast to operator definitions, variable definitions always require
-a type. Otherwise, it may be too hard for the type checker to infer the
-types of the state variables.
+Variable definitions always require type. Otherwise, it may be too hard
+for the type checker to infer the types of the state variables.
```
shonfeder, line 262: missing article and i think the second clause isn't helpful.

```suggestion
Variable definitions always require a type.
```

tutorials/lesson3-anatomy/coin.md
```diff
@@ -279,15 +290,15 @@ It is often convenient to define a few helper operators.
 
 We start with the definition of `state` that represents the entire
 state as a record. This is what we do in the above code with the definition of
-`state`. Notice that the definition of `state` is not `pure` anymore. It is
-referring to the state variables and thus cannot be pure.
+`state`. Notice that the definition of `state` is prefixed with `val`, not `pure val`.
+Since `state` accesses state variables is it impure.
```
shonfeder, line 294:
```suggestion
`state`. Notice that the definition of `state` is prefixed with `val`, not `pure val`:
since `state` accesses state variables, is it impure.
```

shonfeder (approved):
Looks good!  Just a few suggestions.

PR #650, opened 2023-02-21
Fix a bug introduced during REPL refactoring

### LINKED ISSUE #648: Evaluation of imported state variables is broken in REPL
As noted by @pdini, [Step 10](https://github.com/informalsystems/quint/blob/main/tutorials/lesson3-anatomy/coin.md#10-assign-initial-values-to-the-state-variables) in the tutorial 3 does not work. This is caused by the recent refactoring of how REPL is treating modules.

konnov:
Closes #648. Since REPL is using multiple top-level modules now, we have to be careful about mapping variables from different modules to different modules. In this PR, all variable names are prefixed with the module name. This fixes the bug, but it slightly changes the behavior of `_lastTrace`, which now contains fully qualified names instead of local names. Since we have some rework planned for `_lastTrace` in any case, I will revisit this behavior of `_lastTrace` later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - REPL avoid name clashes in different modules (#621)
 - REPL session is now a proper Quint file (#621)
+- A regression in REPL caused by multiple modules (#650)
 
 ### Security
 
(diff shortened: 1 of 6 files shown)


bugarela (approved):
LGTM! Let's try to prioritize the `_lastTrace` improvements as they are very user facing :smile:

PR #617, opened 2023-02-08
Run typecheck on the SuperSpec and fix signature in builtin.qnt 

shonfeder:
This changeset was initially meant to fix https://github.com/informalsystems/quint/pull/617/commits/3370abe9d524ff7051c68267dae493daa3ccbc9a but along the way I encountered a few bits of tech debt and stale/untested stuff. Please see individual commits for details.

- [x] Tests added for any new code
- [x] Ran `npm run format` (or had formatting run automatically on all files edited)
- [x] Documentation added for any new functionality
- [x] Entries added to the respective `CHANGELOG.md` for any new functionality
- [x] Feature table on [`README.md`](../README.md#roadmap) updated for any listed functionality
- [x] Regenerated documentation with `quint docs` if `builtin.qnt` was changed

PR #651, opened 2023-02-22
Release v0.6.0

shonfeder:
### Added

- command `test` to run unit tests (#634)
- option `--with-lookup` of the command `parse` (#639)

### Changed

- `docs` subcommand now outputs generated docs to stdout (#617)
- Mode errors are better explained and include a fix suggestion (#619)
- Typechecking now reports errors when instance overrides are not compatible with
  the original definition (#622)

### Removed

- REPL does not support nested modules any longer (#621)

### Fixed

- REPL avoid name clashes in different modules (#621)
- REPL session is now a proper Quint file (#621)
- A regression in REPL caused by multiple modules (#650)

### Documentation

- the [Prisoners puzzle](./examples/puzzles/prisoners) (#634)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 
 ### Added
+### Changed
+### Deprecated
+### Removed
+### Fixed
+### Security
+
+## v0.6.0 -- 2023-02-22
+
+### Added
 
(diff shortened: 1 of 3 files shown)


bugarela (approved):
LGTM! Are we bumping the minor because of the removal of nested modules? If so, I'd like to note that they were only removed from the REPL so far, and we still have to remove them from parsing and static analysis. But I guess it makes sense to bump the version on the first deprecation change.

PR #660, opened 2023-02-23
Add followup tweaks and docs related to releasing the VSCode extension

shonfeder:
Thanks to @bugarela for the pointers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -189,10 +189,28 @@ executable and the VSCode plugin.
   
 ### VSCode Plugin
 
+#### Prerequisites
+
+- [vsce](https://github.com/microsoft/vscode-vsce)
+- [yalc](https://github.com/wclr/yalc)
+- Access to manage https://marketplace.visualstudio.com/manage/publishers/informal
+
+#### Steps
+
 - Prepare a release of the VSCode plugin by running the script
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -189,10 +189,28 @@ executable and the VSCode plugin.
   
 ### VSCode Plugin
 
+#### Prerequisites
+
+- [vsce](https://github.com/microsoft/vscode-vsce)
+- [yalc](https://github.com/wclr/yalc)
```
bugarela, line 195: Just a note: `yalc` is a bit of a weird dependency here because we only need to run `yalc remove` if `yalc add` was run before. If someone doesn't even have it installed, then they shouldn't worry about it. But, since we are using `yalc` on the script, they'd have to install it in order to run the script.

I don't think that's necessary for now, but at some point we could try to remove this requirement by changing the script to something like "if there are traces of yalc usage (i.e. files), run `yalc remove`"

bugarela (approved):
Awesome :rocket:

CONTRIBUTING.md
shonfeder: Thanks for the info!

This is indeed currently recording what is needed to run the script (I hope :) ).

PR #661, opened 2023-02-24
Fix modes in nested definitions

### LINKED ISSUE #567: Mode checker loses track of parameters in nested definitions

bugarela:
Hello :octocat: 

This closes #567 and unblocks #646. 

The previous behavior of the mode checker would consider a definition as "having an effect" iff the effect for that definition refer to a variable in the result that didn't appear on its parameters (when present). That means `(Read[v0]) => Read[v0]` would be `pure`, while `(Read[v0]) => Read[v1]` wouldn't. 

Then, nested definitions had issues because some paramater of the outer definition wouldn't be considered a parameter when looking at the nested definition, and therefore the effects that referred to outer parameters would be considered to be introducing that effect, while they were not.

See the example from the issue:
```scala
pure def foo(x: int): int = {
  val y = x + 1
  y
}
```

lets say `x` is assigned effect `Read[v42]`. The definition `val y = x + 1` would have effect `Read[v42]`, and that would be considered a state-level definition (`val`) while it should be a `pure val` since it's just adding `1` to something already in the context.

### The solution

We filter out free variables when looking for which effects have affected variables. This way, in the example, `v42` would be filtered out of `Read[]` and that would result in `pure`.


quint/src/effects/base.ts
```diff
@@ -174,14 +174,39 @@ function bindVariables(name: string, variables: Variables): Either<string, Subst
...
 
-function variablesNames(variables: Variables): string[] {
+/**
+ * Finds all variable names refered to by some variables
+ *
+ * @param variables the variables to be searched
+ *
+ * @returns a list of names
+ */
+export function variablesNames(variables: Variables): string[] {
```
konnov, line 184: The bodies of `variablesNames` and `stateVariables` look very similar. Would it make sense to generalize these two functions into one, so we can choose between `variables.name` and `variables.vars` in a lambda?

quint/test/effects/modeChecker.test.ts
```diff
@@ -179,6 +195,16 @@ describe('checkModes', () => {
 
     assert.isEmpty(errors, `Should find no errors, found: ${[...errors.values()].map(quintErrorToString)}`)
     assert.deepEqual(suggestions.size, 0)
+  })
+
+  it('keeps track of parametes in nested definitions', () => {
```
konnov, line 200:
```suggestion
  it('keeps track of parameters in nested definitions', () => {
```

konnov (approved):
This makes sense to me!

PR #665, opened 2023-02-25
Add integration tests with Apalache

### LINKED ISSUE #635: Integration tests for Quint -> Apalache
To ensure we don't break the transpilation, and the correctness of the translation, we will need some integration test in which:

- `quint` parses spec into JSON encoded `QuintIR` 
- `apalache` runs `check` or `typecheck` on the `QuintIR` input

I think these integration tests should live in this repo, since any breakage is likely to come from changes to the quint builtins or the quint IR.

As Igor noted, at some point soon we may want to make it so that Apalache's modelchecking can be invoked from `quint` (or the other way around).

shonfeder:
Closes #635

Adds basic integration tests with Apalache, to ensure we can catch any breakage
in our integration with the model checker.

The approach here is dead simple:

We add a phony make target that will always fetch the latest Apalache release,
then a suite for txm tests that will parse specs with quint and then feed the
JSON IR to Apalache.

As per our discussions, ultimately this will be supplanted with a quint
subcommand that routes invocations thru apalache's server-mode. For now, we just
want to catch breakage.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,6 +78,14 @@ jobs:
         # See https://github.com/anko/txm/issues/10
         run: cd ./quint && txm io-cli-tests.md
         if: matrix.operating-system != 'windows-latest'
+      - uses: actions/setup-java@v3
+        with:
+          distribution: "temurin" # See 'Supported distributions' for available options
+          java-version: "19"
+      - name: Apalache integration tests
+        run: cd ./quint && npm run apalache-integration
+        env:
+          GH_TOKEN: ${{ github.token }}
 
   quint-antlr-grammar:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


konnov (approved):
That's great! Looking at the scrips, this is probably how we should automatically download apalache in `quint` in general. It should be user-friendly, so people would not have to go and figure how to install apalache. They would need java though :(

bugarela (approved):
LGTM! At this point, apalache is a (quite specific) dev dependency, so I think we should avoid dowloading it for people (as we do with `txm`).

One suggestion if you think it's useful: explain this test, make command and dependency in `CONTRIBUTING.md`

shonfeder:
Thanks for the reviews!

@bugarela I added a note about the tests to the place that `CONTRIBUTING.md` said they should be :)

PR #669, opened 2023-02-28
Fix effect checker performance killer

### LINKED ISSUE #662: Obvious effect checker performance killer with high-impact
https://github.com/informalsystems/quint/blob/1f578f397acc930e067420b798aae377398626e9/quint/src/effects/inferrer.ts#L194-L201

When inferring effects for operator application, we incrementally compose and apply substitutions for the entire inference scope (previous and current modules). This is very silly and a product of me not polishing the scope of this substitutions after the first prototype.

I've made some time measurements and the accumulated time in this block of code takes 60s of the total 66s taken to run `quint typecheck` on the ics23 spec.

bugarela:
Hello :octocat: 

Closes #662. In operator application, instead of applying new substitutions to all previously inferred effects, apply them only to effects from the application itself (parameters and result). In order for this to work, we need to compose these new substitutions with the previously known substitutions in a way where the "older" variables are binded to the new ones, as the new ones wouldn't occur elsewhere.

This reduces typechecking of `ics23` to 7 seconds (from over 70s in `main`).

I realize that accumulating all substitutions in `this.substitutions` is also reckless, but this is not considerably affecting performance at this point and it is not trivial to figure out the scope to which we should propagate those.

### DIFF
--- a/quint/src/effects/inferrer.ts
+++ b/quint/src/effects/inferrer.ts
@@ -197,14 +197,17 @@ export class EffectInferrer implements IRVisitor {
           .chain(s => {
             this.substitutions = s
 
-            paramsResult.map(effects => 
+            paramsResult.map(effects =>
               zip(effects, expr.args.map(a => a.id)).forEach(([effect, id]) => {
                 if (!effect || !id) {
-                  throw new Error('Invalid arrays')
+                  // Impossible: effects and expr.args are the same length
+                  throw new Error(`Expected ${expr.args.length} effects, but got ${effects.length}`)
                 }
+
                 const r = applySubstitution(s, effect).map(toScheme)
-              this.addToResults(id, r)
-            }))
+                this.addToResults(id, r)
+              })
+            )
 
             return applySubstitution(s, resultEffect)
           })
--- a/quint/src/effects/inferrer.ts
+++ b/quint/src/effects/inferrer.ts
@@ -19,12 +19,13 @@ import { expressionToString } from '../IRprinting'
 import { IRVisitor, walkModule } from '../IRVisitor'
 import { QuintApp, QuintBool, QuintEx, QuintInt, QuintLambda, QuintLet, QuintModule, QuintModuleDef, QuintName, QuintOpDef, QuintStr } from '../quintIr'
 import { Effect, EffectScheme, Signature, effectNames, toScheme, unify } from './base'
-import { Substitutions, applySubstitution, applySubstitutionToScheme, compose } from './substitutions'
+import { Substitutions, applySubstitution, compose } from './substitutions'
 import { Error, ErrorTree, buildErrorLeaf, buildErrorTree, errorTreeToString } from '../errorTree'
 import { ScopeTree, treeFromModule } from '../scoping'
 import { getSignatures } from './builtinSignatures'
 import { FreshVarGenerator } from '../FreshVarGenerator'
 import { effectToString } from './printing'
+import { zip } from 'lodash'
 
 export type EffectInferenceResult = [Map<bigint, ErrorTree>, Map<bigint, EffectScheme>]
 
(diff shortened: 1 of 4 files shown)


quint/src/effects/substitutions.ts
```diff
@@ -87,23 +87,6 @@ export function applySubstitution(subs: Substitutions, e: Effect): Either<ErrorT
...
- * Applies substitutions to an effect scheme, replacing all *non-quantified* names with their
- * substitution values when they are defined.
- *
- * @param subs the substitutions to be applied
- * @param e the effect to be transformed
- *
- * @returns the effect resulting from the substitutions' application on the given
- *          effect, when successful. Otherwise, an error tree with an error message and its trace.
- */
-export function applySubstitutionToScheme(subs: Substitutions, e: EffectScheme): Either<ErrorTree, EffectScheme> {
```
bugarela, line 100: No longer used

quint/test/effects/inferrer.test.ts
```diff
@@ -1,39 +1,37 @@
...
-      { kind: 'const', identifier: 'S', reference: 11n },
-      { kind: 'param', identifier: 'tup', reference: 12n },
+      { kind: 'const', identifier: 'N', reference: 1n },
+      { kind: 'const', identifier: 'S', reference: 1n },
+      { kind: 'var', identifier: 'x', reference: 1n },
     ]),
   })
 
-  const definitionsTable: LookupTableByModule = new Map<string, LookupTable>([['wrapper', table]])
+  function inferEffectsForModule(quintModule: QuintModule): EffectInferenceResult {
```
bugarela, line 20: I gave up on mocking definitions tables last week. This is an equivalent change to what I made on `modeChecker` tests recently: use the definitions collector instead of trying to keep these complicated mocks (as the id matters and it's super hard to figure out the correct ones).

quint/test/effects/inferrer.test.ts
```diff
@@ -158,22 +148,39 @@ describe('inferEffects', () => {
...
 
     const expectedEffect = "∀ v0, v1 . (Read[v0] & Temporal[v1]) => Read[v0, 'x'] & Temporal[v1]"
 
     assert.isEmpty(errors, `Should find no errors, found: ${[...errors.values()].map(errorTreeToString)}`)
     assert.deepEqual(effectSchemeToString(effects.get(11n)!), expectedEffect)
   })
 
+  it('keeps track of substitutions with nested defs', () => {
+    const quintModule = buildModuleWithDefs([
+      'pure def a(x) = and{' +
```
bugarela, line 161: This is the smallest example I could find to reproduce a scenario where subsititutions where composed the wrong way and the final signature would contain non-quantified variables for no reason. This is based on `ics23`'s `leftBranchesAreEmpty`.

quint/src/effects/inferrer.ts
```diff
@@ -192,14 +193,17 @@ export class EffectInferrer implements IRVisitor {
...
           .chain(s => {
             this.substitutions = s
 
-            this.effects.forEach((effect, id) => {
-              const r = applySubstitutionToScheme(s, effect)
+            paramsResult.map(effects => zip(effects, expr.args.map(a => a.id)).forEach(([effect, id]) => {
+              if (!effect || !id) {
+                throw new Error('Invalid arrays')
+              }
+              const r = applySubstitution(s, effect).map(toScheme)
```
shonfeder, line 204: I think a line break for the body of the mapped function can help indicate the scope of the mapped function here:

```suggestion
            paramsResult.map(effects => 
              zip(effects, expr.args.map(a => a.id)).forEach(([effect, id]) => {
                if (!effect || !id) {
                  throw new Error('Invalid arrays')
                }
                const r = applySubstitution(s, effect).map(toScheme)
```

otherwise, i need to count parens to figure out if we have of the `map(zip).forEach` or `map(zip.forEach)`.

quint/src/effects/inferrer.ts
```diff
@@ -192,14 +193,17 @@ export class EffectInferrer implements IRVisitor {
...
-          .chain(s => compose(s, this.substitutions))
+          .chain(s => compose(this.substitutions, s))
           .chain(s => {
             this.substitutions = s
 
-            this.effects.forEach((effect, id) => {
-              const r = applySubstitutionToScheme(s, effect)
+            paramsResult.map(effects => zip(effects, expr.args.map(a => a.id)).forEach(([effect, id]) => {
+              if (!effect || !id) {
+                throw new Error('Invalid arrays')
```
shonfeder, line 202: This error seems a bit cryptic. Is it meant to be an impossible case? Is the actual problem here that we don't have proper paring between effects and ids? Maybe the message could indicate what is wrong with the arrays.

shonfeder (approved):
Looks great!

shonfeder:
If this has resulted in such a massive improvement in typechecking perf., maybe worth a note in the changelog?

quint/src/effects/inferrer.ts
bugarela: It's an impossible scenario: it would happen if `effects` and `expr.args` had different sizes, but they can't. I'll add a comment and improve the message :smile:

PR #670, opened 2023-02-28
ADR006: Design of modules and lookup tables

### LINKED ISSUE #667: ADR on instances and lookup tables
as discussed, we need one

konnov:
Closes #667. This is the first version of ADR006 that addresses four design issues that we are facing right now:

 - No identifiers for operator parameters
 - Using `Map[string, ValueDefinition]` for lookup
 - No distinction between variables produced by multiple instances
 - Confusing syntax of module instances

### DIFF
--- a/doc/adr006-modules.lit.md
+++ b/doc/adr006-modules.lit.md
@@ -228,10 +228,22 @@ concept that is not supported by the language anymore.
 
 ## 3. Options
 
+### Make IR classes rich and use references
+
 TBD
 
-
 ## 4. Solution
+
+**The general design principle:** All static analyses, including
+name resolution and type inference, should produce a map `Map<bigint, T>`
+as an output. In this map:
+
+ - `bigint` is the unique identifier of an IR object. For example, the
+   identifier of a definition or of a parameter.
+
+ - `T` is a data structure representing the result of the analysis.  For
+   example, this could be the inferred type (for the type checker), the effect
+   (for the effect checker), or the definition itself (for name resolution).
 
 ### 4.1. Addressing D1 and D2
 
@@ -326,8 +338,47 @@ module general {
   action init = x' = N
   action step = x' = dec(x)
 }
-
-TBD
 ```
+
+As you see, we do not change anything in the syntax. Similar to the table
+of `fun`, we organize the lookup table of `general` as follows:
+
+```mermaid
+graph TB
+    subgraph "table for 'general'"
+      fdec["4"]
+      finc["8"]
+      i1["1"]
+      i2["5"]
+      i3["10"]
+      i4["11"]
+      gN["13"]
+      gx["15"]
+      ...
+    end
+
+    subgraph "Intermediate representation"
+      dec["def 'dec': { id: 4, ... }"]
+      inc["def 'inc': { id: 8, ... }"]
+      dec_i["param 'i': { id: 10, ...}"]
+      inc_i["param 'i': { id: 11, ...}"]
+      N["const 'N': { id: 13 }"]
+      x["const 'x': { id: 15 }"]
+      ...
+    end
+
+    finc --> inc
+    fdec --> dec
+    i1 --> dec_i
+    i2 --> inc_i
+    i3 --> dec_i
+    i4 --> inc_i
+    gN --> N
+    gx --> x
+```
+
+The interesting outcome of this approach is that we do not need a separate
+lookup table per module. Since we are using unique identifiers as keys, we
+could simply produce a flat table for all modules.
 
 
--- a/doc/adr006-modules.lit.md
+++ b/doc/adr006-modules.lit.md
@@ -2,7 +2,7 @@
 
 | Revision | Date       | Author           |
 | :------- | :--------- | :--------------- |
-| 1        | 28.03.2023 | Igor Konnov      |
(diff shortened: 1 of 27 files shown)


doc/adr006-modules.lit.md
```diff
@@ -0,0 +1,553 @@
...
+
+### Make IR classes rich and use references
+
+A straightforward solution to the issue D2 would be to introduce the results of
+the analyses directly into the intermediate representation. For instance, we
+could extend `QuintName` with an optional type, an optional effect, an optional
+reference to the definition that the name refers to:
+
+```ts
+export interface QuintName extends WithId {
```
bugarela, line 277: In order for name resolution to "fill" the `_ref` fields, type checker "fill" the `_type` field, etc. We would have to make this representation mutable somehow, right? I feel like we might be going towards a direction of having a mutable IR anyways (i.e. to implement flattening), and that seems ok to me. If we can assign types and effects to all components of the IR, then we don't need the `Map<bigint, T>` structures anymore.

doc/adr006-modules.lit.md
```diff
@@ -0,0 +1,553 @@
...
+module and tune them with the supplied expressions for the constants.
+Effectively, the flattener would transform the module `fixed` of our example
+into the following code:
+
+```bluespec generated/adr006instances.qnt +=
+module fixed {
+  // the instance general(N = 3)
+
+  // this is how `import fun.*` is transformed
+  import fun.* as I3
```
bugarela, line 493: These lines are unclear to me. Are we replacing `include fun.* as I3` with `import fun.* as I3`? Seems like we should be omitting this last line in flattening.

doc/adr006-modules.lit.md
```diff
@@ -0,0 +1,553 @@
...
+```
+
+We believe that this syntax better reflects the nature of instances in Quint:
+
+ - All names of an instance are included into the namespace of the containing
+   module, possibly with a prefix, e.g., with `I3` or `I5`.
+
+ - Some names replaced with expressions, e.g., `N`.
+
+ - The behavior is different from `import`.
```
bugarela, line 547: Why? Isn't import a form of including but with no constant replacement? What is the motivation to differ them?

doc/adr006-modules.lit.md
konnov: I would really like to avoid such a rich data structure. It is typically much harder to serialize and debug. Also, it's not that obvious which stages modify which parts of the IR in this case.

bugarela: Actually, we might want different approaches for this "filling" and for the flattening. For filling, we could but all these new optional fields under a mutable object in a `analysisData` field, and then do things like `expr.analysisData.addType(inferredType)` to "update" the IR.

For flattening, we just have to add/remove definitions, so we could directly modify the `defs` array from the module.

bugarela (approved):
I'm already excited to implement this! Seems amazing. I left a few questions regarding implementation and the `include` syntax/behavior.

doc/adr006-modules.lit.md
bugarela: > I would really like to avoid such a rich data structure. It is typically much harder to serialize and debug. Also, it's not that obvious which stages modify which parts of the IR in this case.

I'm not sure I understand. Are you suggesting we implement this "rich" data structure you suggested on the ADR but don't make it mutable? So we'd have to update it with data from `Map<bigint, T>` at some point(s)?

konnov: I mean, this is under the options. So I did not intend to implement this one. I should probably make it clear though.

doc/adr006-modules.lit.md
konnov: This is just how the translation of `import fun.*` could look like, if we did flattening by code transformation. In principle, this could all be done via the lookup table.

doc/adr006-modules.lit.md
konnov: `include` is almost like `import` but not only. It should also copy the definitions, when required. The semantics is similar but not exactly the same.

konnov:
> Actually, we might want different approaches for this "filling" and for the flattening. For filling, we could but all these new optional fields under a mutable object in a analysisData field, and then do things like expr.analysisData.addType(inferredType) to "update" the IR.

We could. I just remember working with a few projects, where AST became a rich data structure, and this was a nightmare to navigate and understand. I agree that if we find a good API for managing these tables, it would not matter whether they are actually stored as maps or object attributes. I just want to avoid direct modifications of the IR objects. In principle, they should be immutable.

> For flattening, we just have to add/remove definitions, so we could directly modify the defs array from the module.

Yes, we would have to copy the definitions and the respective maps for them. This should not be hard.

konnov:
I have added a discussion about the options that we have. @bugarela, your comment made me thinking that we could combine the both approaches, that is, store the analyses results as object fields, but provide an API that imitates a map. This would give us the required efficiency and, at the same time, isolate the results under a clear interface. What we will lose though is the ability to assign different results (for the same analysis) to the same IR object. Not sure that we need it though.

I have found that we are repeating some code over `Map`, so it may be a good opportunity to design a better API for this.

konnov:
I have merged this ADR. If we decide to design a better API for the analysis maps, we should update this ADR later.

PR #671, opened 2023-03-01
Renaming across effect system

### LINKED ISSUE #668: Names for effect variables are confusing
I'm using `Variables` to refer to an abstraction on top of `StateVariables`, and now we have also quantified variables. I was also using the term `quantified` to refer to a reference to one or more state variables. So both words `Variables` and `quantified` are too overloaded and I need to replace some of their instances with new words.

https://github.com/informalsystems/quint/pull/658#discussion_r1119053458

bugarela:
Hello :octocat: 

This fixes #668 by renaming `Variables` to `Entity`, and the kind `'quantified'` to `'variable'` (plus some consequential renaming from that).

### DIFF
--- a/quint/src/effects/base.ts
+++ b/quint/src/effects/base.ts
@@ -57,8 +57,8 @@ export type EffectScheme = {
   entityVariables: Set<string>,
 }
 
-/* A state entity */
-export interface StateVariables {
+/* A state variable */
+export interface StateVariable {
   /* The variable name */
   name: string,
   /* The id of the expression on which the state variable ocurred */
@@ -71,7 +71,7 @@ export interface StateVariables {
  */
 export type Entity =
   /* A list of state variables */
-  | { kind: 'concrete', stateVariables: StateVariables[] }
+  | { kind: 'concrete', stateVariables: StateVariable[] }
   /* A variable representing some entity */
   | { kind: 'variable', name: string, reference?: bigint }
   /* A combination of entities */
@@ -194,7 +194,7 @@ export function entityNames(entity: Entity): string[] {
  *
  * @returns a list of state entities
  */
-export function stateVariables(entity: Entity): StateVariables[] {
+export function stateVariables(entity: Entity): StateVariable[] {
   switch (entity.kind) {
     case 'variable':
       return []
--- a/quint/src/effects/base.ts
+++ b/quint/src/effects/base.ts
@@ -12,9 +12,9 @@
  * @module
  */
 
-import { effectComponentToString, effectToString, variablesToString } from './printing'
+import { effectComponentToString, effectToString, entityToString } from './printing'
 import { Either, left, mergeInMany, right } from '@sweet-monads/either'
-import { Substitutions, applySubstitution, applySubstitutionToVariables, compose } from './substitutions'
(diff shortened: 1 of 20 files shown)


quint/src/effects/base.ts
```diff
@@ -12,80 +12,78 @@
...
+  /* The id of the expression on which the state variable ocurred */
   reference: bigint
 }
 
 /*
- * The variables an effect acts upon. Either a list of state variables, a
- * quantification over them or a combination of other variables.
- * Uses the plural form since all forms represent a plurality of variables.
+ * The Entity an effect acts upon. Either a list of state variables, a
+ * variable or a combination of other entities.
```
shonfeder, line 70: If one kind of "entity" is "a list of state variables", then it doesn't seem like a single  state variable can be a "state entity"? This seemes to confuse wholes and parts?

shonfeder (commented):
This is a hard problem (one of the two!). My initial review is not convincing me that adding the extremely abstract and vague term "entity" into the mix clarifies our naming problem.

Perhaps, rather than jumping to introducing new abstract terms, we should update our design documents with an anaysis and exposition of the variety of variables and names at play in the effect system, which level they operate on, and how they interact. This may help us clarify our understanding of the system enough that clear, disambiguating names fall out. But, if not, we'll at least have the documentation to help us keep track of what's going on!

This will take a bit of time, and may be lower prioity this quarter, but I think it could yield a kind of substantive improvement that we don't get by just introducing a new abstract term.

This is my feeling at the moment, but if another reviewer finds that this change helps with reabability, I'm ok deferring.

WDYT?

quint/src/effects/base.ts
bugarela: The comment was suppose to say "state variable", not entity. This slipped through my reviews (auto-replacing was not trivial :stuck_out_tongue:), sorry

bugarela:
@shonfeder I think I did a poor job at setting expectations here. This is not a change that is, alone, suppose to make the effect system easier to understand. The naming problem of calling two different things by the same name (as it was happening with both "variables" and "quantified") makes it impossible to properly define what is what, and makes the code very hard to maintain.

If it's not clear what X is, calling it `Variables` or `Entity` won't make a difference, I agree. But it's (pretty much) impossible to explain what `Variables` are if there are 3 different things being called `Variables`. So this PR only disambiguates the names. Our discussion about ADRs this morning reminded me that the Effects ADR is very outdated, specially after these changes, and I completely agree we need better documentation.

WDYT about me updating the ADR with proper definitions for all the new distinct names (which shouldn't take long!) and then, if we can think of better abstractions on top of those, I can make an additional iteration on this?

shonfeder:
> WDYT about me updating the ADR with proper definitions for all the new distinct names (which shouldn't take long!) and then, if we can think of better abstractions on top of those, I can make an additional iteration on this?

That makes good sense to me! Being able too survey all these kinds of names in a table should help I think!

With that expectation, I'll mark this as approved, and leave it up to you whether you'd rather merge this and then iterate, or hold off on these changes until the design doc updates.

shonfeder (approved):

bugarela:
I'll merge this because I really don't want it to get stale somehow, since conflicts with this won't be nice to solve. I'll start working on an ADR update right away.

PR #676, opened 2023-03-03
Prevent self imports and instances

### LINKED ISSUE #523: Module imports and circular imports
In the course of #489, I started to write a simple example of bank transfer. Somehow, my coding habits kicked in and I have produced the following code:

```scala
module bank {
  type Addr = str
  // we isolate the keeper interface in a separate namespace
  module keeper {
    // to my surprise, this import works
    // import bank.*
    pure def validateBalance(addr: Addr): bool = {
      true
    }
}
```

There are several points to discuss here:

 - do we want `Addr` to be accessible in `keeper`? I think it is normal behavior in many languages.
 - I can access `Addr` in `keeper` via `bank::Addr`, but is it what we want?
 - Surprisingly, `import bank.*` is working inside `keeper`. It probably leads to circular references in `keeper`.

/cc @bugarela, @shonfeder 

This is probably related to the issue #483 by @lasarojc.

bugarela:
Hello :octocat: 

Closes #523 by preventing a module to import or instance itself. I believe the remaining problems of the issue were already addressed by removing nested modules :smile: 

### DIFF
--- a/quint/src/importResolver.ts
+++ b/quint/src/importResolver.ts
@@ -79,7 +79,7 @@ class ImportResolverVisitor implements IRVisitor {
       // Importing current module
       this.errors.set(def.id, {
         code: 'QNT407',
-        message: `Cannot instance ${def.protoName} inside itself`,
+        message: `Cannot instantiate ${def.protoName} inside ${def.protoName}`,
         data: {},
       })
       return
--- a/quint/src/importResolver.ts
+++ b/quint/src/importResolver.ts
@@ -136,7 +136,7 @@ class ImportResolverVisitor implements IRVisitor {
       // Importing current module
       this.errors.set(def.id, {
         code: 'QNT407',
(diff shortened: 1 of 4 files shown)


quint/src/importResolver.ts
```diff
@@ -75,6 +75,16 @@ class ImportResolverVisitor implements IRVisitor {
   }
 
   enterInstance(def: QuintInstance): void {
+    if (def.protoName === this.currentModule?.name) {
+      // Importing current module
+      this.errors.set(def.id, {
+        code: 'QNT407',
+        message: `Cannot instance ${def.protoName} inside itself`,
```
konnov, line 82:
```suggestion
        message: `Cannot instantiate ${def.protoName} inside ${def.protoName}`,
```

konnov (approved):
Looks good!

PR #679, opened 2023-03-06
Update Effect System ADR

### LINKED ISSUE #675: Update effect system ADR

bugarela:
Hello :octocat:

Our Effect System ADR was severely outdated. This PR updates it and attempts to clarify how the concepts translate to the implementation.

I'm having trouble iterating on this without external feedback. I feel like there is room for improvement here, so please help me find the most confusing parts so I can focus on adding more detail to those.

Closes #675


doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
 Option 1 may be too restrictive and barely used, since, for most operators, it's
 desirable that they can be applied to state variables (i.e. `+` in `x + 1` if
 `x` is a state variable).
 
-PS: alternative names for `pure` are `static` and `stateless`.
+### Decision
+
+We have decided to implement option 2 for all modes. According to this option, a
+definition is considered pure if it does not add any read effect on top of the
+effects that the parameters may carry. A definition is considered `def` if it
```
ivan-gavran, line 74:
```suggestion
effects that the parameters may carry. A definition is considered `pure` if it
```

bugarela: This is actually meant to be `def`. Since the option 2 was implemented for all modes, not only pure, I gave a description of the `def` mode as well. I'll update this to clarify that this is another example, as I agree it is a bit confusing in the current structure.

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
+structured as a list of effect components. An effect component has a kind and an
+entity. The kind can be `read`, `update` or `temporal`, and the entity is the
+entity the effect acts upon.
+
+```typescript
+/* Kinds for concrete effect components */
+export type ComponentKind = 'read' | 'update' | 'temporal'
+
+/* A concrete effect component, specifying a kind and the entity it acts upon */
+export interface EffectComponent { kind: ComponentKind, entity: Entity }
```
shonfeder, line 170: I'm not sure why we need the `EffectComponent` now. Iiuc, it's just a product of an effect kind and entity, so maybe using a tuple would suffice and reduce the number of abstract nouns we have to track by one? Just thinking out loud... :)

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
-with the goal of reaching the normal form. Equivalence between `E1` and `E2` is
-expressed by `E1 ≡ E2`, and the equivalence symbol `≡` has the lowest precedence
-on this system.
+```typescript
+/* Arrow effects for expressions with effects depending on parameters */
+export interface ArrowEffect { kind: 'arrow', params: Effect[], result: Effect }
+```
 
+Just like with entities, an effect can also be a variable, which can be
+substituted with another effect during inferrence. Those are the most general
```
shonfeder, line 199: I think the phasing 'an effect can also be a variable' sets up a point of possoble conceptual and terminological confusion. Rather, there is a class of variables that range over effects (and a different class that range over "entities", and "state variables" aren't actually variables that range over anything, lol). So how about something like:

```suggestion
These is a separate class of variables that range over effects, which can be
substituted with another effect during inference. Those are the most general
```

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
+kinds of effects, and will be used when no effect information for a given
+expression is available, i.e. `def foo(x) = x` will have a `(e) => e` effect
+signature, where `e` is an effect variable, since there are no constraints on
+the effect of `x`. An effect variable is equivalent to `Read[v1] & Update[v2] &
+Temporal[v3]`, where `v1`, `v2` and `v3` are entity variables.
```
shonfeder, line 204: I think that possibly part of the conceptual confusion i am hitting in this paragraph results from not keeping a clear separation between the ts representation forming our meta language, and the object language? So, `EffectVariable` is one of the types comprising the `Effect` union type`, but an effect variable *stands for* but *is not itself* an effect. Does this make sense?

This would suggest a rephrasing like:

```suggestion
When no effect information for a given expression is available, the effect is left
undetermined, as indicated by an effect variable. I.e. `def foo(x) = x` will have be inferred to have the effect signature `(e) => e`, where `e` is an effect variable, since there are no constraints on the effect of `x`. An **effect variable** `e` is equivalent to the effect `Read[v1] & Update[v2] & Temporal[v3]`, where `v1`, `v2` and `v3` are _entity variables_.
```

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
+
+```typescript
+/* A variable representing some effect */
```
shonfeder, line 207:
```suggestion
/* A variable standing for an unknown effect */
```

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
+export interface EffectVariable { kind: 'variable', name: string }
 ```
-E1 ≡ Ex & Read[x0, ..., xn]   E2 ≡ Ey & Read[y0, ..., yn]
----------------------------------------------------------- (READ)
-E1 & E2 ≡ Ex & Ey & Read[x0, ..., xn, y0, ..., yn]
 
+### Effect Schemes
+
+Both effects and entities can be universally quantified, and we need to keep
+track of which of them are. Effect schemes are responsible to keep track of that
```
shonfeder, line 214:
```suggestion
track of which of them are. Effect schemes are responsible for keeping track of that
```

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
+information.
+
+```typescript
+/*
+ * An effect scheme, listing which effect variables (names referring to effects)
+ * and entity variables (names refering to entities) are universally quantified
```
shonfeder, line 220: I think this also confuses levels a bit. IIUC, the names here refer to variables, which stand for unknown effects, they do not refer to effects directly.

```suggestion
 * An effect scheme, listing which effect variables (represented by the names of effect variables)
 * and entity variables (represented by the names of entity variables) are universally quantified
```

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
+ * in the effect
+ */
+export type EffectScheme = {
+  /* The effect */
+  effect: Effect,
+  /* Universally quantified names refering to Effects */
+  effectVariables: Set<string>,
+  /* Universally quantified names refering to Entities */
```
shonfeder, line 228:
```suggestion
  /* The names of universally quantified Effect variables */
  effectVariables: Set<string>,
  /* The names of universally quantified Entity variables */
```

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
 
 ### Unification
 
-Should be pretty straightforward variable substitution for both effect variables
-and variables inside Read and Update statements.
+Unification is pretty standard. It recursively looks at effect variables and
+entity variables to bind. The most complicated part must be the unification of
+unions of variables, which requires unifying sets. Since that is too complicated
+to implement, we use some workarounds to avoid having to do that in the few
+cases where we would need to. Those workarounds revolve on reasoning about one
```
shonfeder, line 255:
```suggestion
cases where we would need to. Those workarounds rely on reasoning about one
```

Or 'revolve around'

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
+of the unions as a single new variable, which is done by introducing a new
+hashed variable, which is then binded to the other union. A hashed variable for
+the union `r1, r2, r3` is `r1#r2#r3`. These hashing procedure is also used to
```
shonfeder, line 258:
```suggestion
the union `r1, r2, r3` is `r1#r2#r3`. This hashing procedure is also used to
```

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
 ### Inference rules
 
-Inferring names: variables have effect `Read[v]` (unless they are used in as
-targets of assignment, where their resulting effect will be inferred correctly
-as `Update[v]`), constants have no effect (Pure), values used as parameters can
-read variables (the read variables are introduces as a new quantification over
-variables with the parameter name prefixed by `r_`) operators resolve to the
-effect of their respective bodies.
+Inferring names: variables have effect `Read[<variable_name>]` (unless they are
+used in as targets of assignment, where their resulting effect will be inferred
```
shonfeder, line 264:
```suggestion
used as targets of assignment, where their resulting effect will be inferred
```

shonfeder (approved):
Thanks for writing this update. Going thru it properly i really found it helpful. 

I've offered a bunch of suggestions and questions and musings. Feel free to ignore/deffer anything that isn't immediately helpful or that seems out of scope.

Sorry the review here took so long!

bugarela:
Thanks for the review! Since it's been a while and I'm not totally in context anymore, I'll finish other tasks that I'm currently working on before addressing your comments (which I'd usually prioritize).

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
+introduces a new `Read` effect, but does not introduce any `Update` or
+`Temporal` effects. The same principle applies to other types of definitions.
 
 ### Notation
 
 ```
 Identifiers v, c, op
-Effects E, Ei ::= Read[vars] | Update[vars] | Pure | E & E | (E0, ..., EN) => E
+Effects E, Ei ::= e | Read[entity] | Update[entity] | Temporal[entity] | Pure | E & E | (E0, ..., EN) => E
+Entities V ::= v | 'x' | v1, ..., vn
```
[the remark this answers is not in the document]
bugarela: `v1, ..., vn` was suppose to be `V1, ..., Vn`, so `V` is an entity that can be replaced by entity variables `v` or state variables `'x'`

doc/adr004-effect-system.md
```diff
@@ -38,150 +36,292 @@ simplicity.
...
+effects that the parameters may carry. A definition is considered `def` if it
+introduces a new `Read` effect, but does not introduce any `Update` or
+`Temporal` effects. The same principle applies to other types of definitions.
 
 ### Notation
 
 ```
 Identifiers v, c, op
-Effects E, Ei ::= Read[vars] | Update[vars] | Pure | E & E | (E0, ..., EN) => E
+Effects E, Ei ::= e | Read[entity] | Update[entity] | Temporal[entity] | Pure | E & E | (E0, ..., EN) => E
```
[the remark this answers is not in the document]
bugarela: `e` is suppose to be an effect variable. I will change the symbol for `Expression` to `expr`

doc/adr004-effect-system.md
bugarela: We have to define a name for the field in `ConcreteEffect` that carries this, which is now `components`. So I don't think we can reduce the number, and this might help clarify it.

doc/adr004-effect-system.md
```diff
@@ -71,19 +71,23 @@ desirable that they can be applied to state variables (i.e. `+` in `x + 1` if
...
+StateVariable 'x', 'y', 'z'
+ConstantExpression c
+Operator op
+EntityVariable v
+Entities V, Vi ::= v | 'x' | V1, ..., Vn
+EffectVariable e
+Effect E, Ei ::= e | Read[V] | Update[V] | Temporal[V] | Pure | E & E | (E0, ..., EN) => E
+Expression expr, expri ::= any Quint expression
+Context Γ ::= lookup tables and previously inferred effects
+Substitution S, Si ::= {e ↦ E} | {v ↦ V} | S ∪ S
```
shonfeder, line 95: This totally clears up my confusions. Thanks!

doc/adr004-effect-system.md
```diff
@@ -38,234 +36,426 @@ simplicity.
...
+parameters of an operator and the effect resulting from the operator's application.
+
+```typescript
+/* Arrow effects for expressions with effects depending on parameters */
+export interface ArrowEffect { kind: 'arrow', params: Effect[], result: Effect }
+```
+
+These is a separate class of variables that range over effects, which can be
+substituted with another effect during inference. Those are the most general
+When no effect information for a given expression is available, the effect is left
```
shonfeder, line 209: Looks like part of the sentence was lost in applying suggestions:

```suggestion
substituted with another effect during inference. Effect variables are the most general kind of effects.
When no effect information for a given expression is available, the effect is left
```

shonfeder (approved):
Looks great to me! A few suggestions to fix typos introduced by my suggestions, but all the clarifications and elaborations look excellent. Thank you!

PR #686, opened 2023-03-08
Updating the REPL tutorial

### LINKED ISSUE #339: Update the REPL documentation
https://github.com/informalsystems/tnt/blob/main/doc/repl.md is seriously outdated

konnov:
Closes #339. This is an update of the REPL tutorial:

 - [x] It is now written in the literate programming style (kind of) with `lmt`
 - [x] All definitions are extracted in the module `kettle.qnt`
 - [x] Most of the REPL interactions are extracted in the file `replTest.txt`, which is postprocessed in `Makefile`
   - The postprocessing is a bit adhoc. It would be great to make this approach more systematic in the future. In principle, it is not so far from notebooks. 
 - [x] An integration test replays the session from `replTestIn.txt` with `quint` and compares the output to `replTestOut.txt`.
 - [x] Quint now supports the flag `-q`, which suppresses the noisy prompts `>>> ` and `... `. Great for testing.
 - [x] REPL does not output line numbers anymore. There is no good understanding of which line numbers should be used: relative to the last expression, relative to the invisible buffer, relative to the input file, etc. It's also hard to do reliable scripting, when line numbers are printed in REPL.


konnov:
For some reason, diff is failing on windows. Perhaps, I call it wrong way.

konnov:
@shonfeder, perhaps you know how to fix this problem with Windows, as you had it with `txm` once.

shonfeder:
I recommend just not running the diff check on windows. That’s how I’ve dealt with the other tests that requiring doffing.

quint/cli-tests.md
```diff
@@ -211,3 +211,16 @@ Temporarily disabled.
...
+The REPL tutorial is reproducible in REPL.
+
+<!-- !test check REPL tutorial -->
+```
+if [ "$UNAME" == "Linux" -o "$UNAME" == "Darwin" ]; then
+  quint -q -r ../tutorials/repl/kettle.qnt::kettle <../tutorials/repl/replTestIn.txt \
+    | diff - ../tutorials/repl/replTestOut.txt
+fi
+# else diff does not work as expected on windows
+```
```
bugarela, line 226: I'd rather move this to `io-cli-tests.md` which are not run on windows anyway. Then we can get rid of the `if` statement.

bugarela (approved):
This is a nice improvement on structure! I have only a suggestion on testing that I think we should implement.

quint/cli-tests.md
konnov: oh, I knew that one of them should be win-free :)

PR #689, opened 2023-03-09
Ids for lambda parameters

### LINKED ISSUE #624: Introduce a structure for lambda parameters
Currently, lambda definitions have their parameters as strings:

https://github.com/informalsystems/quint/blob/7c29089ab746c4d925eaf423a2e96b4619e14245/quint/src/quintIr.ts#L121-L130

Since the operator definitions are represented with lambdas, there is no good way to refer exactly to the place where the parameter has been defined (except for giving the definition's name and the parameter name). As we discussed with @bugarela and @shonfeder, it would be good to introduce a structure for parameters, e.g.:

```scala
export interface QuintParamDef extends WithId {
  /** Expressions kind ('param' -- an operator parameter) */
  kind: 'param',
  /** The parameter name */
  name: string,
}
```

This would enable easier goto definition and simplify code in REPL.

bugarela:
Hello :octocat: 

Big diff because of tests and generated (antlr) code.

This PR introduces ids for lambda parameters, which unblocks several issues and allows us to report errors and provide information more precisely.

Closes #624 


examples/solidity/ERC20/erc20.qnt
```diff
@@ -147,7 +147,7 @@ module erc20 {
     balanceOf' = balanceOf,
     val failed = or {
      tx.value < 0,
-     tx.sender > tx.spender,
+     tx.sender == tx.spender,
```
bugarela, line 150: This is a fix for a type error that was not flagged by the type checker before, but is now. The reason why this PR change impacted this is not quite clear to me yet. However, @konnov reported a similar bug this morning which is not solved by this, so I'll further investigate this behavior when looking at that.

quint/src/effects/inferrer.ts
```diff
@@ -91,24 +91,6 @@ export class EffectInferrer implements IRVisitor {
     }
 
     switch (def.kind) {
-      case 'param': {
```
bugarela, line 94: No need for this special treatment/workaround anymore now that lambdas have their own IDs.

quint/src/generated/Quint.g4
```diff
@@ -150,14 +150,17 @@ unitOrExpr :    unit | expr;
...
+lambda:         parameter '=>' expr
+        |       '(' parameter (',' parameter)* ')' '=>' expr
         ;
 
+
 // an identifier or a hole '_'
 identOrHole :   '_' | IDENTIFIER
         ;
 
+parameter: identOrHole;
```
bugarela, line 162: This is not completely necessary, but makes the code in `ToIRListener` more consistent, and I think it makes some sense to differentiate parameters from simple identifiers at the grammar level: parameters will later become a `QuintLambdaParameter`, while `identOrHole` is just used as a string.

quint/test/effects/inferrer.test.ts
```diff
@@ -23,7 +23,16 @@ describe('inferEffects', () => {
 
     const inferrer = new EffectInferrer(definitionsTable)
     return inferrer.inferEffects(quintModule)
+  }
+
+  function effectForDef(quintModule: QuintModule, effects: Map<bigint, EffectScheme>, defName: string ) {
```
bugarela, line 28: Dealing with IDs was getting to hard, so I wrote this helper function instead.

quint/test/nameResolver.test.ts
```diff
@@ -77,19 +77,7 @@ describe('nameResolver', () => {
 
       result
         .mapLeft(errors => assert.deepEqual(errors, [
-          { kind: 'value', name: 'x', definitionName: 'd0', moduleName, reference: 2n },
-        ]))
-        .map(_ => assert.fail('Expected errors'))
-    })
-
-    it('find unresolved names inside lambdas', () => {
```
bugarela, line 85: This test was duplicated somehow, so I removed one copy of it.

bugarela:
@shonfeder This breaks compatibility with the Apalache translation, as we anticipated :disappointed:. How can we handle the broken tests? Since we don't have backwards compatibility in neither side, it might be tricky to release a new Apalache version with this change that will break every other PR tests except for this. Perhaps this should be a non-required CI check?

quint/src/quintIr.ts
```diff
@@ -118,11 +118,16 @@ export interface QuintApp extends WithId {
...
+  name: string,
+}
+
 export interface QuintLambda extends WithId {
   /** Expressions kind ('lambda' -- operator abstraction) */
   kind: 'lambda',
-  /** Identifiers for the formal parameters */
-  params: string[],
+  /** The formal parameters */
+  params: QuintLambdaParameter[],
```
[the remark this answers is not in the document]
shonfeder: To be clear, i expect there is a good reason, but i'm curious what it is :)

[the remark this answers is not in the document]
bugarela: Yep, so it would be tricky since, in many times we analyze a `NameEx`, we look for something already defined, which isn't the case here (i.e. the type inferrer, when visiting a `NameEx`, will use the lookup table to obtain the original definition and get the type for that)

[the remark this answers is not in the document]
shonfeder: I see, thanks.

quint/src/generated/Quint.g4
shonfeder: Good call I think this makes great sense!

shonfeder (approved):
Looks great! Thanks for the helpful comments!

PR #694, opened 2023-03-09
add a dashboard for the examples

konnov:
This PR adds a dashboard for our examples, which shows how far we could go for every example in the pipeline. This should be useful to new users, who would like to start with the examples that are known to work.


bugarela (approved):
Great! Shall we add a checklist to the PR template?

konnov:
> Great! Shall we add a checklist to the PR template?

Yes, how do we do that?

PR #697, opened 2023-03-09
update the feature table

konnov:
As we discussed, our feature table is a bit outdated. Here is an update according to my understanding.


konnov:
Shall we just merge the table in the current state and update it later, if needed?

bugarela (approved):
OK!

PR #674, opened 2023-03-02
Remove support for nested modules

bugarela:
Hello :octocat: 

Closes #548.

Removes support for nested modules in: parsing, IR, name resolution, import resolution, definition collection, type/effect checking and examples.

I did a few changes to interfaces, such as `importResolver` and `definitionsCollector` now return a single lookup table since they analyze a single module (with no nested ones from now on). There's more room for improvement on interfaces like those, but I feel like that's not a priority right now, specially since we want to change how lookup tables work (see https://github.com/informalsystems/quint/issues/625)

- [X] Tests added for any new code
- [X] Ran `npm run format` (or had formatting run automatically on all files edited)
- [ ] Documentation added for any new functionality
- [x] Entries added to the respective `CHANGELOG.md` for any new functionality
- [ ] Feature table on [`README.md`](../README.md#roadmap) updated for any listed functionality

PR #698, opened 2023-03-09
Release v0.7.0

shonfeder:
### Changed

- Inferred effects are now properly quantified (#658)

### Removed

- Nested modules are no longer supported (#674)

### Fixed

- Modes for nested definitions are now properly checked (#661)
- All basic operators can now be used with temporal formulas (#646)
- Effect checking performance for large specs is massively improved (#669)
- A module can no longer import or instance itself (#676)


konnov (approved):
LGTM

PR #703, opened 2023-03-10
add two previews

konnov:
I have added two short previews of the VSCode plugin and the CLI. Since these videos are over 50MB, they are tracked via [git-lfs](https://git-lfs.com/).


konnov:
Git does not show mp4 videos in markdown, even if they are uploaded via git-lfs. Do we have another option? Youtube or something? :)

doc/previews.md
```diff
@@ -0,0 +1,17 @@
+# Quick previews
+
+## VSCode plugin
+
+![VSCode Plugin Preview](./img/vscode-2023-03-10.mp4)
```
bugarela, line 5: I haven't tested this before, but perhaps it will work:

```suggestion
<video src='./img/vscode-2023-03-10.mp4.mp4' width=180/>
```

bugarela: If it doesn't work, we can try uploading to `githubusercontent` itself. Here's an example of how that would look: https://github.com/mfitzp/diffcast/blob/master/README.md

konnov: yeah, the video tag did not work

bugarela: Perhaps we should try to compress the videos. The one in the link has 38 seconds and only 2.14 MB, and the quality is ok

konnov:
thanks @bugarela! The trick with uploading a video to an issue and then using the link worked. Also, I've compressed the videos a bit.

Issue #625, opened 2023-02-14
Introduce reference maps to do lookup


konnov:
Currently, we lookup for definitions via `lookupValue`:

https://github.com/informalsystems/quint/blob/7c29089ab746c4d925eaf423a2e96b4619e14245/quint/src/lookupTable.ts#L120-L138

Since names do not have to be unique within a module or an operator definition, `lookupValue` does an additional search that relates the referencing expression to the definition. This may be a potential performance bottleneck (we don't know whether it is an actual bottleneck). Instead of doing the lookup via names and ids, we could simply store a map of identifiers to identifiers, which maps a reference id to the id of the definition that has introduced the name. This would make the lookup conceptually easier and, probably, more efficient.

This work is blocked by #624.

PR #707, opened 2023-03-14
Ids for instance override parameters

bugarela:
Hello :octocat: 

This is preparatory work towards #625:
1. Introduces ids for instance overrides parameters (i.e. `c` in `module A1 = A(c = 1)`) so we can lookup their references by id in the new tables.
2. Report partial results in the import resolver: related to the previous item, we would have some noisy errors if trying to resolve names (now for instance overrides as well) after import resolution fails. That is, if a non-pure val is given for an instance parameter, we don't want to see some name resolution failing for that parameter - so we proceed with import and name resolution and only report that error at the end.


quint/src/importResolver.ts
```diff
@@ -101,25 +97,25 @@ class ImportResolverVisitor implements IRVisitor {
     // For each override, check if the name exists in the instanced module and is a constant.
```
konnov, line 97:
```suggestion
    // For each override, check if the name exists in the instantiated module and is a constant.
```

konnov (approved):
LGTM

PR #708, opened 2023-03-16
Make Tendermint spec runnable in REPL

### LINKED ISSUE #692: Transfer the Tendermint spec to Quint
As discussed with @sergio-mena, it would be great to have a working Tendermint spec in Quint.

We actually have an older hand-written [spec in Quint](https://github.com/informalsystems/quint/blob/main/examples/cosmos/tendermint/TendermintAcc_004.qnt). It is syntactically correct and type correct. However, it is using ADTs which are not supported by REPL yet, and ADTs will be redesigned soon #539.

However, we had a newer [spec of Tendermint in TLA+](https://github.com/informalsystems/tendermint/blob/igor/tendermint-ind-inv/spec/light-client/accountability/TendermintAcc_004_draft.tla). It does not require variants or ADTs. I believe that we can translate it at any point, without any blockers.

konnov:
Closes #692. This is a newer version of the Tendermint spec in Quint that is compatible with the newer version in TLA+. I had to cut some features that are not supported in REPL yet (constants and tuple unpacking). Once these features are supported, we have to revisit the spec.


bugarela (approved):
LGTM! I'm not familiar enough with Tendermint to carefully review the spec, but it looks good in general and it's great to have some tests :tada:

PR #718, opened 2023-03-20
Update the syntax for instances in ADR006

konnov:
@bugarela mentioned several times that instances look very much like imports. At first, I could not see it. But if we accept that any module may be imported, including the modules with state variables and constants, then instances are indeed similar to imports. I have experimented with an alternative syntax that highlights this similarity, and I like it. This change in the ADR reflects the new import-like syntax.


doc/adr006-modules.lit.md
```diff
@@ -556,24 +559,36 @@ Using this syntax, our example looks like follows:
 
 ```bluespec
 module fixed {
-  include general(N = 3) as I3
-  include general(N = 5) as I5
-  
+  import general(N = 3).*
+  import general(N = 5) as I5
```
bugarela, line 563: I think we should also support normal imports with qualifiers like this.
```suggestion
  import general(N = 5) as I5
  import statelessModule as M
```

doc/adr006-modules.lit.md
```diff
@@ -556,24 +559,36 @@ Using this syntax, our example looks like follows:
...
-for action operators. Although the word `include` is not as nice as `instance`,
-the word `include` reflects the behavior of this construct better.
+for action operators.
+
+One potential problem with the approach of using import-like behavior is that
+the names that are introduced via this construct are not automatically
+exported. We can imagine that sometimes automatic export would be useful.
+Actually, the same occurs in programming languages, e.g., in TypeScript.  This
+is why TypeScript has the [export import][] combo. If we see a need for such
+behavior, we may add it in the future.
```
bugarela, line 592: I'm currently trying to write tendermint tests with instances and it's quite hard without the exporting. But I think the `import general(N = 5).*` syntax will fill the gap. Then we can do:

```bluespec
module TendermintAcc { ... }

module TendermintTest {
  import TendermintAcc.*

  ...
}

module Tendermint_n4_f1_Test {
  import TendermintAcc(
    Corr = Set("p1", "p2", "p3"),
    Faulty = Set("p4"),
    N = 4,
    T = 1,
    ValidValues = Set("v0", "v1"),
    InvalidValues = Set("v2"),
    MaxRound = 4
  ).*

  import TendermintTest.*
}
```

This would require a CLI `test` call for each instance since the CLI only tests a single module at a time. There should be other alternatives, this is one that doesn't require typing namespaces.

bugarela (approved):

doc/adr006-modules.lit.md
konnov: > I think we should also support normal imports with qualifiers like this.

Totally. There is no big conceptual difference.

doc/adr006-modules.lit.md
konnov: That's interesting. But I am not sure it works as expected. Would not it make sense to instantiate `TendermintTest` instead of `TendermintAcc`?

bugarela: Sorry, that what I meant.

PR #719, opened 2023-03-20
Output a summary after a trace

### LINKED ISSUE #704: Change output order in "quint run --invariant=..."
I use invariant checking for some basic checks. It gives me
```
[ok] No violation found (32000ms).
 You may increase --max-samples and --max-steps.

See the example:
```
... followed by a trace of 20 steps. 

That is, in my terminal, I always need to scroll up to figure out whether it is OK. It would be great to give the "ok" response at the very end of the output, so that one can easily get the most important information.

konnov:
Closes #704. Simply changes the output order.


quint/src/cliCommands.ts
```diff
@@ -373,16 +373,18 @@ export function runSimulator(prev: TypecheckedStage):
...
+        console.log(chalk.gray('An example execution:'))
         console.log('---------------------------------------------')
         printTrace(console.log, result.trace)
         console.log('---------------------------------------------')
+        if (result.status === 'ok') {
+          console.log(chalk.green('[ok]')
+            + ' No violation found ' + chalk.gray(`(${elapsedMs}ms).`))
+          console.log(chalk.gray('You may increase --max-samples and --max-steps.'))
+        } else {
+          console.log(chalk.red('[nok]')
```
bugarela, line 385: `nok` is kinda weird, I haven't seen it in any testing framework and it is very math-looking to me :grimacing:. Can we use `error` or go back to `violation`? I personally prefer `violation` since I have a delay on understanding what `error` means in apalache to this day :stuck_out_tongue:

bugarela (approved):

quint/src/cliCommands.ts
konnov: Yeah, we should not use `error`, as it may mean something different.

konnov: I also have not seen 'nok', but it looks like testing tools use it often: https://www.google.com/search?q=test+output+ok+nok&oq=test+output+ok+nok

PR #733, opened 2023-03-23
Remove extraneous punctuation

thpani:
Remove the "." at the end of one of the bullet points.

I come across this quite often, and it bothers me every time 😅


bugarela (approved):
I'm pretty sure that's on me, and my advisor would agree: I'm pretty inconsistent with bullet point punctuation :laughing:. Thanks for the fix and I'll try to pay more attention in the future (but I don't promise anything).

PR #736, opened 2023-03-23
Add a lottery example

konnov:
This is a yet another example of how we could use Quint to specify non-trivial behavior of smart contracts. The interesting thing about this example is that it is one of the highlights from the paper by [Zhang, Zhuo, Xu, Lin](https://github.com/ZhangZhuoSJTU/Web3Bugs) at ICSE'23. This example is in the category of contracts that cannot be reasoned automatically by static analysis tools. This is not really surprising, as the security property requires us to specify the behavior of the contract and the behavior of the Solidity mempool.

I was able to specify the property as a state invariant and reproduce it with a test in Quint. However, the random simulator did not find the violation after 10 min. Seems like a good call for Apalache (and more advanced fuzzing).

Also, this example screams: ADTs, ADTs :))

### DIFF
--- a/examples/solidity/icse23-fig7/README.md
+++ b/examples/solidity/icse23-fig7/README.md
@@ -9,7 +9,7 @@ the paper (Fig. 7):
 The paper and the accompanying material can be found at
 [Web3bugs](https://github.com/ZhangZhuoSJTU/Web3Bugs).
 
-This example belongs to one of the issues that are not detected
+This example belongs to a class of issues that are not detected
 automatically by static analysis tools. But even if this behavior
 cannot be automatically detected, can it be specified? The ability
 to specify it is the first step towards automatic verification.
--- a/examples/solidity/icse23-fig7/README.md
+++ b/examples/solidity/icse23-fig7/README.md
@@ -21,11 +21,7 @@ with the test:
     // This is the scenario reported in the paper.
     // By modeling the mempool, we can specify this scenario.
     run lotteryMultiBuyTest = {
-        init.then(submit(ResetTx("owner")))
-            .then(commit(ResetTx("owner")))
-            .then(submit(BuyTx("alice", 1, 1000)))
-            .then(commit(BuyTx("alice", 1, 1000)))
-            .then(submit(EnterDrawingPhaseTx("owner")))
+        init.then(submit(EnterDrawingPhaseTx("owner")))
             .then(commit(EnterDrawingPhaseTx("owner")))
             .then(submit(DrawTx("owner", 3)))
             // Bob can buy lottery tickets in the drawing phase,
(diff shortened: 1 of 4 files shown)


examples/solidity/icse23-fig7/README.md
```diff
@@ -0,0 +1,55 @@
...
+In [lottery.qnt](./lottery.qnt), we specify the lottery contract
+that is used as an illustration of atomicity violation in
+the paper (Fig. 7):
+
+    Zhuo Zhang, Brian Zhang, Wen Xu, Zhiqiang Lin, "Demystifying Exploitable Bugs in Smart Contracts." ICSE'23.
+
+The paper and the accompanying material can be found at
+[Web3bugs](https://github.com/ZhangZhuoSJTU/Web3Bugs).
+
+This example belongs to one of the issues that are not detected
```
thpani, line 12:
```suggestion
This example belongs to a class of issues that are not detected
```

examples/solidity/icse23-fig7/lottery.qnt
```diff
@@ -0,0 +1,374 @@
+// -*- mode: Bluespec; -*-
+// A Quint version of the contract presented in Figure 7 of the paper:
+//
+// https://github.com/ZhangZhuoSJTU/Web3Bugs/blob/main/papers/icse23.pdf
+
+// A specification of the lottery contract.
+// We write this specification by following functional idioms,
+// as this way allows us to easily compose this contract with mempool.
+// A non-functional way would be more concise but not that easier to compose.
```
thpani, line 9:
```suggestion
// A non-functional way would be more concise but not that easy to compose.
```

thpani (commented):
I guess this is useful for internal communication, but what is the benefit of having it live in the repo?

To the end user it only communicates "this does not currently work" and we'll need to revise it once we have proper integration with Apalache (which adds another work item we would need to track).


NB: I haven't reviewed the spec.

examples/solidity/icse23-fig7/README.md
```diff
@@ -0,0 +1,55 @@
...
+The paper and the accompanying material can be found at
+[Web3bugs](https://github.com/ZhangZhuoSJTU/Web3Bugs).
+
+This example belongs to a class of issues that are not detected
+automatically by static analysis tools. But even if this behavior
+cannot be automatically detected, can it be specified? The ability
+to specify it is the first step towards automatic verification.
+
+To be more exact, the problematic behavior is formally specified
+with the test:
```
bugarela, line 18: Maybe this will help clarifying that we were able to reproduce the bug, even though random simulation couldn't find it.

```suggestion
and can be reproduced with the test:
```

examples/solidity/icse23-fig7/README.md
```diff
@@ -0,0 +1,55 @@
...
+```bluespec
+    val noBuyInDrawingInv = {
+        and {
+            lastTx.status == "success",
+            lastTx.kind == "buy" or lastTx.kind == "multiBuy"
+        } implies not(lotteryState.drawingPhase)
+    }
+```
+
+We can try to disprove this invariant in the random simulator:
```
bugarela, line 46:
```suggestion
Instead of specifying a run with the steps leading to the bug, we can try to disprove this invariant in the random simulator:
```

bugarela (approved):
A lot of interesting idioms in this spec! I'm even more excited about ADTs now :smile:

bugarela:
> I guess this is useful for internal communication, but what is the benefit of having it live in the repo?
To the end user it only communicates "this does not currently work" and we'll need to revise it once we have proper integration with Apalache (which adds another work item we'll have to track).

I thought the same when firsti reading the description, but then I realized we actually have an example that reproduces the bug, even though random simulation cannot find an example for us. I made a few suggestions that can maybe help clarify that.

thpani:
Aaah, I see that I misunderstood both the PR description and the text.

Thanks for the edits @bugarela, that makes it much clearer!

PR #738, opened 2023-03-23
Fix the spec and the description. There was a bug in the search scope

konnov:
I was too quick at discarding the results of random simulation. One of the runs demonstrated that there was a bug in the search scope. Basically, the contract required "owner" to submit a transaction, but "owner" was not included in the search scope. In this PR, I am choosing the owner non-deterministically. Also, the invariant is stronger now, describing the vulnerability in more detail.


bugarela (approved):
LGTM

PR #739, opened 2023-03-23
New instances syntax

### LINKED ISSUE #11: Merge Quint modules
As outlined in [ADR001](https://github.com/informalsystems/tnt/blob/main/doc/adr001-transpiler-architecture.md#35-flatten), we should implement the flattening task

### LINKED ISSUE #724: New syntax for instances
From the latest update to ADR 6: https://github.com/informalsystems/quint/pull/718

### LINKED ISSUE #730: Update source maps after flattening
https://github.com/informalsystems/quint/pull/725#discussion_r1144619308

bugarela:
Hello :octocat: 

This introduces our new syntax for instances, introduced in ADR 6, which closes #724. In order to use that syntax in a nice way, I needed to combine it with imports (see the Tendermint example), so I also implemented flattening for modules, which closes #11.

With flattened modules, the main module is now self contained and we only need to evaluate that single module in the REPL. So I have removed some support for multiple modules in the compiler, which makes variable names easier to read. But that broke error reporting since source maps for flattened modules were still not consistent. So, I also fixed the source map problem, which closes #730.


quint/io-cli-tests.md
```diff
@@ -283,23 +283,23 @@ quint run --init=Init --step=Next --seed=abcde --max-steps=4 \
 An example execution:
 ---------------------------------------------
 action step0 = all {
-  counters::n' = 1,
+  n' = 1,
```
bugarela, line 286: This works because we only need to evaluate the main module now that it is self-contained by flattening. WDYT @konnov ? Is there something I'm missing here or is this great? :sweat_smile:

tutorials/repl/replTestOut.txt
```diff
@@ -8,7 +8,7 @@ false
 true
 -40
 104
-runtime error: error: Variable kettle::temperature is not set
+runtime error: error: Variable temperature is not set
```
bugarela, line 11: I changed this and the markdown by hand because setting up a go env to build `lmt` didn't fit my schedule for this week :stuck_out_tongue_closed_eyes:. I hope it doesn't break automatic generation.

quint/io-cli-tests.md
konnov: Yes, these traces look much better now! I was also annoyed by `counters::`.

quint/src/runtime/compile.ts
```diff
@@ -128,30 +128,36 @@ export function compile(
...
   rand: () => number): CompilationContext {
   const modulesByName = new Map(modules.map(m => [m.name, m]))
-  const flattenedModules = modules.map(m => flatten(m, lookupTable, modulesByName))
-  const flattenedAnalysis = parsePhase2({ modules: flattenedModules, sourceMap }).mapLeft(errors => {
-    // This should not happen, as the flattening should not introduce any errors
-    // Since parsePhase2 analysis of the original modules has already assured all names are correct.
-    throw new Error(`Error on resolving names for flattened modules: ${errors.map(e => e.explanation)}`)
-  }).unwrap()
+  const lastId = modules.map(m => m.id).sort((a, b) => Number(a - b))[modules.length - 1]
+  const idGenerator = newIdGenerator(lastId)
```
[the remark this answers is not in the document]
konnov: The simulator converts runtime values back to `QuintEx`. That's where it can introduces new identifiers, which would clash with the identifiers here.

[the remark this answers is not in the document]
bugarela: But the ids generated in `toQuintEx` are not used for anything, right? If we did use them, we should find a way to get the original ids back somehow (by storing them in the runtime value, for example). The current ids used there won't have a source map entry nor a lookup table entry for them, so anything would most likely break if we tried to use them.

If there's no real problem in the clashing right now, I'd like to merge this and take a look at this generator problem again. I don't think simulating a global state is a good solution here.

PR #726, opened 2023-03-21
Implement Go To Definition for type aliases

bugarela:
Hello :octocat: 

With the new lookup table, it's pretty easy to implement "go to definition" functionality for type aliases.

Closes #558 

- [X] Tests added for any new code
- [X] Ran `npm run format` (or had formatting run automatically on all files edited)
- [X] Documentation added for any new functionality
- [ ] Entries added to the respective `CHANGELOG.md` for any new functionality
- [ ] Feature table on [`README.md`](../README.md#roadmap) updated for any listed functionality

PR #745, opened 2023-03-24
VSCode Release v0.2.1

bugarela:
### Added

- Go to definition now works for parameters (#711)
- Go to definition now works for type aliases (#726)


shonfeder (approved):
:fire:

Issue #747, opened 2023-03-27
Need to lookup for operator qualifiers


konnov:
When working on #677, I have realized that we have to tell apart names that refer to actions from other names. This is needed to trace actions but not all kinds of other operators, as there may be a lot of calls to user-defined operators. I tried to use the inferred modes, but that's not exactly what is needed. Currently, `lookupTable` refers to a custom data structure:

https://github.com/informalsystems/quint/blob/218fc1b0ad9a6f9621090c3f66d6db925e2bbe93/quint/src/lookupTable.ts#L4-L11

It would be much more convenient, if the lookup table was returned `QuintOpDef`. I believe that `Definition` duplicates `QuintOpDef` to  large extent.

@bugarela, wdyt?

bugarela:
The inferred modes seem a more suitable answer here. The qualifier provided by the user (available at name resolving) might me too general (i.e. all non-temporal operators could be annotated with `action`).

Other than that, it should be totally possible to add the qualifier to `Definition`. I'm just trying to understand if that's really what you need.

konnov:
> The inferred modes seem a more suitable answer here. The qualifier provided by the user (available at name resolving) might me too general (i.e. all non-temporal operators could be annotated with `action`).

I tried that, and modes were a bit too hard to wire. On the other hand, we are already doing a lookup for names.

bugarela:
This was solved a while ago :)

PR #748, opened 2023-03-27
Track the executions and produce detailed logs

### LINKED ISSUE #677: Output traces similar to Solidity tools
Currently, we output traces as sequences of states. This is good for connecting traces to testing tools. However, these traces are harder to replay by a human engineer. Essentially, the users have to replay all calls to their actions and operators, given the current state. We could produce an output that is produced by testing tools for Solidity, e.g., [trace output by Foundry Forge](https://book.getfoundry.sh/forge/traces).

konnov:
Closes #677. As @lasarojc and @ivan-gavran mentioned, it would be good to sometimes see what actions were executed and what they returned. Sometimes, it is even desirable to see the log of executing all user-defined operators.

This PR partially implements this logic. Due to #747, it's hard to tell actions from other operators apart. So this PR treats actions and non-actions the same way. The user can enable operator tracking with `run --verbosity=3`, which is disabled by default. This PR happened to be much more work then I anticipated. It only implements execution tracking for `run`. After trying this feature with @ivan-gavran and @lasarojc, we will see how to proceed.

The main action is happening in `trace.ts`, which defines an `ExecutionListener` and its implementations. Having a listener, `compilerImpl` produces events to the listener, which takes care of traces.


quint/io-cli-tests.md
```diff
@@ -391,22 +385,83 @@ quint run --max-steps=5 --seed=123 --invariant=totalSupplyDoesNotOverflowInv \
...
+────────────────────────────────────────────────────────────────────────────────
+
+[Frame 1]
+ q::step() => true
+ ├─ mint("null", "null", 79724149679233970751606042021871459804676944172294020221997378231441129734144) => true
+ │  ├─ require(true) => true
+ │  └─ require(true) => true
+ │     └─ isUInt(79724149679233970751606042021871459804676944172294020221997378231441129734144) => true
+ └─ q::inv() => true
+    └─ isUInt(79724149679233970751606042021871459804676944172294020221997378231441129734144) => true
```
shonfeder, line 442: Looks cool and helpful!

shonfeder (approved):

PR #764, opened 2023-03-31
Showing execution details in REPL

### LINKED ISSUE #754: Produce detailed logs in REPL and `test`
As a follow up of #748, we should enable the same debugging logic in REPL and `test`

### LINKED ISSUE #762: REPL should show a warning when an action does not update all of the state variables
> Giuliano Losa [said](https://informal-systems.zulipchat.com/#narrow/stream/378959-quint/topic/questions.20while.20trying.20to.20write.20first.20quint.20spec/near/345806139):
>
> Also, still in the REPL, I can execute an action that does not set a variable, and then this causes an error only later on when I try to execute another action. Wouldn't it be better to throw an error immediately?

Sometimes, we have to call an action that does not update all variables. I think it would be reasonable to show a warning in this case. I will open an issue.

konnov:
Closes #754. Closes #762. This PR finishes the work on detailed output for operator evaluation. As in case of `run` and `test`, this output is enabled with `--verbosity=3`. Additionally, one can enable it with `.verbosity=3` directly in REPL.

I've also found the case when the action call was not tracked, namely, the case of nullary actions like `init` or `step`. Fixed that. As a result, the detailed output is a bit more verbose than before.

Also, I've renamed all operators like `__test` and `_lastTrace` to `q::test` and `q::lastTrace`. Let's use the namespace `q` for the operators internal to Quint.

- [x] Update the docs

### DIFF
--- a/quint/src/runtime/impl/compilerImpl.ts
+++ b/quint/src/runtime/impl/compilerImpl.ts
@@ -1432,8 +1432,9 @@ export class CompilerVisitor implements IRVisitor {
                   // the run. Hence, do not report an error here, but simply
                   // drop the run. Otherwise, we would have a lot of false
                   // positives, which look like deadlocks but they are not.
-                  break
                   this.execListener.onUserOperatorReturn(nextApp, [], nextResult)
+                  break
+
                 }
               }
             }
--- a/quint/src/runtime/impl/compilerImpl.ts
+++ b/quint/src/runtime/impl/compilerImpl.ts
@@ -30,7 +30,10 @@ import { RuntimeValue, rv } from './runtimeValue'
 
 import { lastTraceName } from '../compile'
 
-const specialNames = ['__input', '__runResult', '__nruns', '__nsteps', '__init', '__next', '__inv']
+const specialNames = [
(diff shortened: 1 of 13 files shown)


quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1418,8 +1432,8 @@ export class CompilerVisitor implements IRVisitor {
                   // the run. Hence, do not report an error here, but simply
                   // drop the run. Otherwise, we would have a lot of false
                   // positives, which look like deadlocks but they are not.
-                  this.execListener.onUserOperatorReturn(nextApp, [], nextResult)
                   break
+                  this.execListener.onUserOperatorReturn(nextApp, [], nextResult)
```
bugarela, line 1436: Seems like this line is not reachable anymore. Should we remove it?
```suggestion
```

bugarela (approved):
LGTM

quint/src/runtime/impl/compilerImpl.ts
konnov: Oh, it should actually be before `break`. Thanks for finding that!

Issue #701, opened 2023-03-09
Design and plan release and integration with Apalache


shonfeder:
We'll need a thoughtout policy and practice to ensure the best outcomes w/r/t integration and interoperation with Apalache.

Some options discussed today:

- keep a loose and unreliable coupling, and just let quint break combat with Apalache
- pin quint to specific versions of apalache, and run integration tests against the latest Apalache to catch breackage.
- move development of quint and apalache into a monorepo

Let's maybe use this issue to compile requirements before we settle on an approach and document it in an ADR.

cc/ @konnov and @bugarela 

- [ ] Update CLI docs https://github.com/informalsystems/quint/blob/51aaa638e278f250ab4da9d93756533eaa67c15e/doc/quint.md#L281-L287

shonfeder:
For distributing apalache along-with, or thru quint, we can consider these kinds of utilities:

- https://www.npmjs.com/package/node-jre would let us ensure a JRE is installed
- https://www.npmjs.com/package/jdeploy would us distribute apalache via npm (and we can then add it as a dep in our package manifest)

thpani:
> For distributing apalache along-with, or thru quint, we can consider these kinds of utilities:
> 
>     * https://www.npmjs.com/package/node-jre would let us ensure a JRE is installed

`node-jre` is deprecated, and its successor https://github.com/nvuillam/njre hasn't seen updates in over a year (in particular, it still lacks support for the Apple M1 `arm64` architecture). So these fall flat.

PR #770, opened 2023-04-03
VSCode Release v0.3.0

shonfeder:

### DIFF
--- a/vscode/quint-vscode/CHANGELOG.md
+++ b/vscode/quint-vscode/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 ### Changed
 
-- Update quint from `0.2.1` to `0.3.0`
+- Update quint from `0.2.1` to `0.3.0`. See #768.
 ### Deprecated
 ### Removed
 ### Fixed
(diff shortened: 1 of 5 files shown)


shonfeder:
The empty changelog is unfortunate. #701 should help address, but in the meantime, is Ok if we must accept the slight oddity? I'd rather not have to do busy work just cause or release process for the vscode plugin is a bit wonky :)  Tho i will if we think it's important. WDYT, @bugarela ?

bugarela:
It would be nice to automatically add a `update quint from x to y` entry to the changelog, shouldn't be too hard to do. I'd suggest making a manual edit to the changelog adding a note like that now, it shouldn't break the release since it is not very git-dependent afaict.

bugarela (approved):

vscode/quint-vscode/CHANGELOG.md
```diff
@@ -14,6 +14,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 
+## v0.3.0 -- 2023-04-03
+
+### Added
+### Changed
+
+- Update quint from `0.2.1` to `0.3.0`
```
shonfeder, line 22:
```suggestion
- Update quint from `0.2.1` to `0.3.0`. See #768.
```

shonfeder:
Sounds good. Made that change. Thanks for the :heavy_check_mark: !

Issue #530, opened 2023-01-18
Update language manual on `Namespaces and Instances` and `Imports`


bugarela:
The manual is outdated on
- Syntax for namespaces (should be `Inner::x2` instead of `Inner.x2`)
  - We sometimes refer to this as dot-notation, but it's not a dot anymore
- Out-of-order: we should note that this is not yet supported (#129) 
- Requiring an instance for modules that have variables is not yet implemented.
- Depending on what we decide on https://github.com/informalsystems/quint/discussions/496, we might have to remove examples with nested modules.

PR #771, opened 2023-04-03
Module exports

### LINKED ISSUE #753: Chains of imports
As we expected, the default behavior of exports being non-transitives makes it harder to reuse instances in the chains of imports. The example below by @lasarojc demonstrates the complications with chains of imports that include instances.

Load the example in REPL:

```sh
$ quint -r t.qnt
Quint REPL v0.0.3
true
>>> import scratchC.*

>>> B::A::Procs1
Set(1, 2)
>>> B::A::procs0
syntax error: error: Failed to resolve name B::A::procs0 in definition for __input, in module __repl__
B::A::procs0
^^^^^^^^^^^^
>>> import scratchB as B2
>>> B2::init
true
```

What is happening here? Since REPL introduces its own module internally, we have one more hidden import:

```bluespec
module __repl__ {
  import scratchC.*
  // ...
}
```

So the contents of `B` is not visible in `__repl__`, though `B` itself is visible in `__repl__`. I think, the contents of `B` should be actually visible in `__repl__`, since `__repl__` has imported all of `scratchC`.

Related to that, as we discussed earlier, we should probably add the combo `export import`, that transitively propagates imports.

@bugarela, @shonfeder, wdyt?

```bluespec
// t.qnt
module scratchA {
    type Proc = int
    const Procs1: Set[int]
    const Procs2: Set[Proc]
}

module scratchB {
    import scratchA(
        Procs1 = Set(1,2),
        Procs2 = Set(3,4)
    ) as A

    var procs1: Set[A::Proc]
    var procs2: A::Proc -> A::Proc

    action unchangedAll = all {
        procs1' = procs1,
        procs2' = procs2
    }

    action init = all {
        procs1' = A::Procs1,
        procs2' = A::Procs2.mapBy(p => p+1)
    }

    action step =
        unchangedAll
}

module scratchC {
//    import scratchA.*       //Conflicting definitions of Proc and B::Proc
//    import scratchA as A    //A::Procs1 and A::Procs2 are not resolved inside scratchC
    import scratchA(
        Procs1 = Set(5,6),
        Procs2 = Set(7,8)
    ) as A

    import scratchB as B

    var procs3: A::Proc -> Set[A::Proc]

    action unchangedAll = all {
        B::unchangedAll,
        procs3' = procs3
    }

    action init = all {
        B::init,
        procs3' = A::Procs1.mapBy(p => A::Procs2)
    }

    action step =
        unchangedAll
}
```

bugarela:
Hello :octocat: 

This introduces module exports to enable transitivity of imports and fix #753. The syntax for exports is the same for imports, with only the change of keyword (`import` -> `export`), similar to [typescript's approach](https://www.typescriptlang.org/docs/handbook/modules.html#re-exports).

The implementation is scope manipulation: imported definitions are always copied to the importing module with a scope that is that module's id, in order to prevent them from being visible outside of the module. This works since we only copy unscoped definitions when importing a module. To export definitions, we just need to make sure they are copied with no scope.

It is **not** required that exported definitions are previously imported. If exporting previously imported definitions, we first remove the imported definitions (which are scoped). Then, we perform the same steps as importing, but we don't include any scope in the copied definitions.

Documentation will be added with #530 


bugarela:
I just notice a weird behavior that arises from my strategy here: if you only `export` a module and don't `import` it, it works the same as if you have imported it. That is, the definitions will be available in the exporting module.

It's not trivial to fix this, I might need to introduce a separate map of tables to store exported things and available things separately. I'll wait for your review on this to get additional feedback on the behavior before making more iterations. I also think this weird behavior is not a big problem since it won't stop users from doing anything, so it is not urgent to fix it.

konnov:
> It is not required that exported definitions are previously imported.

I guess, it is is not required to import a module in the same scope where it is exported?

editor-plugins/vim/quint.vim
konnov, line 1: Nice! Thanks for taking care of the vim highlighting too :)

examples/language-features/imports.qnt
```diff
@@ -23,21 +19,23 @@ module B {
...
-  //                ^^^ error: sqr is not available
+  // val test_test_sqr = sqr(2) == 4
+  //                     ^^^^^^ error: Failed to resolve name sqr in definition for test_test_sqr, in module B
 
   // however, we can access test_sqr
   val test_test_sqr = test_sqr
+
+  // To enable transitivity, we need to export the names that we want to make
+  // available when importing the current module
+  export A.*
```
konnov, line 30: That's interesting. I would expect `export *` here, don't know whether it is even supported by our syntax. Is not it a bit strange that the namespace `A` is available within `A` itself? On the other hand, I think this is how PLs are doing that, so maybe it's ok.

quint/src/importResolver.ts
```diff
@@ -170,4 +172,89 @@ class ImportResolverVisitor implements IRVisitor {
...
+        data: {},
+      })
+      return
+    }
+
+    const moduleTable = this.tables.get(def.protoName)
+    if (!moduleTable) {
+      // Exporting unexisting module
+      this.errors.set(def.id, {
+        code: 'QNT404',
```
konnov, line 196: Nice error code :)

konnov (approved):
Looks great! We will have to experiment with this feature a bit, to see whether it fully implements the required logic.

examples/language-features/imports.qnt
bugarela: This is inside module `B`, not module `A`. `export *` is the default implicit behavior, and is not supported by our syntax. You can only export other modules or names from other modules.

quint/src/importResolver.ts
```diff
@@ -170,4 +172,89 @@ class ImportResolverVisitor implements IRVisitor {
...
+  // Imported names are copied with a scope since imports are not transitive by
+  // default. Exporting needs to turn those names into unscoped ones so, when
+  // the current module is imported, the names are accessible. Note that it is
+  // also possible to export names that were not imported.
+  enterExport(def: QuintExport) {
+    if (def.protoName === this.currentModule?.name) {
+      // Exporting current module
+      this.errors.set(def.id, {
+        code: 'QNT407',
+        message: `Cannot export ${def.protoName} inside ${def.protoName}`,
```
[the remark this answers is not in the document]
bugarela: As I mentioned in the previous comment, the example does `export A.*` inside `B`, not `A`.

quint/src/importResolver.ts
bugarela: It's from your initial ADR, so kudos to you :smile:

quint/src/importResolver.ts
```diff
@@ -170,4 +172,89 @@ class ImportResolverVisitor implements IRVisitor {
...
+        data: {},
+      })
+      return
+    }
+
+    const qualifier = def.defName ? undefined : (def.qualifiedName ?? def.protoName)
+    const exportableDefinitions = copyNames(moduleTable, qualifier)
+
+    if (!def.defName || def.defName === '*') {
+      // Avoid conflicts by removing the scoped versions of the definitions
```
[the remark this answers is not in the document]
bugarela: No, it's related to previously imported names. If we simply process an import and then an export, there will be two definitions for the same name: a scoped one, from the import, and  a unscoped one, from the export. That will raise an error. So we do this cleanup at every export to ensure only the unscoped one remains.

We could simply modify the previously imported definition to remove its scope, but we also support exports of non-imported definitions, so this is a more general solution.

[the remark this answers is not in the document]
konnov: Makes sense to me!

konnov:
This is a great feature to have. Let's merge it.

PR #773, opened 2023-04-04
Update run name in tutorial description

bugarela:
Hello :octocat: 

At some point, we renamed the run `test1` to `writeReadTest` inside the quint snippets, but forgot to update the markdown description. This fixes that.


konnov (approved):
Cool. Thanks a lot for fixing this!

PR #774, opened 2023-04-04
Fix test command error reporting

### LINKED ISSUE #772: `test` subcommand exits with `0` on failing tests
```sh
$  quint test spec/p2p/p2p.qnt ; echo $?

  p2p
    ok happyPathTest
    1) mixedDirectionsTest

  1 passing (549ms)
  1 failed

  1) mixedDirectionsTest:
      /home/sf/Sync/informal-systems/cometbft/spec/p2p/p2p.qnt:523:13 - error: Assertion failed
      523:             assert(invDirectionsOut),


0
```

shonfeder:
Closes #772

We have a uniform mechanism for reporting errors (along with exit codes) for
failed executions, but the `test` subcommand wasn't using it, so it was exiting
with a `0` and without reporting on stderr. 

This change makes it so that `test` produces a non-zero exit code on failures.

Note that this has also uncovered a failing test.

### DIFF
--- a/quint/src/cliCommands.ts
+++ b/quint/src/cliCommands.ts
@@ -283,9 +283,6 @@ export function runTests(prev: TypecheckedStage): CLIProcedure<TestedStage> {
       out(`\n  ${mainName}`)
     }
 
-    const _matchFun =
-      (n: string): boolean => isMatchingTest(prev.args.match, n)
-
     const options: TestOptions = {
       testMatch: (n: string) => { return isMatchingTest(prev.args.match, n) },
       rand: mkRng(prev.args.seed),
@@ -369,11 +366,11 @@ export function runTests(prev: TypecheckedStage): CLIProcedure<TestedStage> {
     } // else: we have handled the case of module not found already
 
     const errors = namedErrors.map(([_, e]) => e)
-    const result = { ...testing, passed, failed, ignored, errors }
+    const stage = { ...testing, passed, failed, ignored, errors }
     if (errors.length == 0 && failed.length == 0) {
-      return right(result)
+      return right(stage)
     } else {
-      return left({msg: "Tests failed", stage: { ...result, errors }})
+      return left({msg: "Tests failed", stage})
     }
   }
 }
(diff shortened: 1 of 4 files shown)


shonfeder:
I don't see anything in the lottery spec that could set the `lastTx.status` to `"drawing"`. I think this is a specification error?

@konnov Is it OK with you if i ignore this test for now, and open an issue to track need to fix it? I don't want to get side tracked into bebugging this demos spec given the other stuff on my plate, and this fix is blocking correct quint CI for the cosmosbft repo.

shonfeder:
I've tracked the need for followup on the spec test in #775

quint/src/cliCommands.ts
```diff
@@ -283,7 +283,7 @@ export function runTests(prev: TypecheckedStage): CLIProcedure<TestedStage> {
       out(`\n  ${mainName}`)
     }
 
-    const matchFun =
+    const _matchFun =
       (n: string): boolean => isMatchingTest(prev.args.match, n)
```
bugarela, line 288: Seems like we can get rid of this

```suggestion
```

quint/src/cliCommands.ts
```diff
@@ -369,7 +369,12 @@ export function runTests(prev: TypecheckedStage): CLIProcedure<TestedStage> {
     } // else: we have handled the case of module not found already
 
     const errors = namedErrors.map(([_, e]) => e)
-    return right({ ...testing, passed, failed, ignored, errors })
+    const result = { ...testing, passed, failed, ignored, errors }
+    if (errors.length == 0 && failed.length == 0) {
+      return right(result)
+    } else {
+      return left({msg: "Tests failed", stage: { ...result, errors }})
```
bugarela, line 376: `errors` is already inside `result`, so I think we can do this:
```suggestion
      return left({msg: "Tests failed", stage: { ...result }})
```

bugarela (approved):
LGTM! Thanks for fixing this right away

quint/src/cliCommands.ts
shonfeder: Hmm. I thought I had gotten rid of this in https://github.com/informalsystems/quint/pull/774/commits/1823c5ba0010b83ad30d2f504b09da40fc1f51ad  

I wonder if I undid something before commiting? Thanks!

konnov:
Nice! I will check the broken test!

PR #778, opened 2023-04-04
Start spec for ICS20 with the denom trace manipulation process

bugarela:
Hello :octocat: 

This is a first small iteration on an ICS20 spec, specifying only the denom trace functionality, incrementing and decrementing the trace based on packets being sent.

I'm not sure if this is something we want to merge at this stage, I'm open for suggestions on what to do with this right now. I think it can be a useful resource to play around and understand this part of the protocol.


examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,129 @@
...
+  type Uint = int
+
+  type State = {
+    trace: List[str],
+    denom: str,
+  }
+
+  type FungibleTokenPacketData = {
+    trace: List[str],
+    denom: str,
```
konnov, line 20: These two fields are packed into `denom` in [ICS20](https://github.com/cosmos/ibc/blob/main/spec/app/ics-020-fungible-token-transfer/README.md#data-structures). Shall we simply use `State` aka `IbcDenom` from the above type definition?

konnov: Also, `str` inside `List` are actually channel or port identifiers, so it would help to have a record there, e.g.. `{ channel: str, port: str }`. Our first goal is not to mirror the ICS/go/rust data structures one-to-one, but to keep the concepts simple.

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,129 @@
...
+    sourcePort: str,
+    sourceChannel: str,
+    destPort: str,
+    destChannel: str,
+  }
+
+  pure def onRecvPacket(packet: Packet): State =
+    pure val trace = packet.data.trace
+    pure val sourcePrefix = [packet.sourcePort, packet.sourceChannel]
+    if (trace.length() >= 2 and trace.slice(0, 2) == sourcePrefix) {
```
konnov, line 37: Would it make sense to introduce a `pure def` (a predicate) that tells us whether a pair `(port, channel)` are the source for an IBC denom?

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,129 @@
...
+
+  type FungibleTokenPacketData = {
+    trace: List[str],
+    denom: str,
+    amount: Uint,
+    sender: str,
+    receiver: str,
+  }
+
+  type Packet = {
```
konnov, line 26: I really wonder whether we should remove the concept of a packet from the spec of denominations. I find IBC denominations confusing enough, so we don't have to add packets in the picture.

konnov (commented):
Nice first spec! By reading it, I really wonder whether we should introduce a separate module/spec for denominations. I feel that there are plenty of unspoken assumptions about how the traces are supposed to work. We could contribute to the community by finally surfacing these assumptions.

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * Gabriela Moreira, Informal Systems, 2023
+ */
+module ics20 {
+  type Uint = int
+
+  type State = {
+    trace: List[str],
+    denom: str,
+  }
```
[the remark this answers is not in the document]
thpani: Is there a reason to carry around `denom` at all?
It's only the base denomination, and we currently set it once in `init` and then just propagate it.

examples/cosmos/ics20/ics20.qnt
thpani: I think it would be cool to have `isSink`/`isSource` predicates.
Personally, I find them confusing, but they are the basic building blocks in ICS-020.
People could use these to debug their understanding in REPL.

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,129 @@
...
+      "B" -> "channelToB",
+      "C" -> "channelToC"
+    )
+    ({
+      data: {
+        trace: currentState.trace,
+        denom: currentState.denom,
+        amount: 100,
+        sender: source,
+        receiver: dest,
```
thpani, line 82: We're not currently using these packet fields, but you're currently setting them to chain names. They're actually accounts.

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,129 @@
...
+  // │      Chain A      │           │               Chain B              │         │      Chain C      │
+  // │                   │           │                                    │         │                   │
+  // │ ┌───────────────┐ │           │ ┌──────────────┐  ┌──────────────┐ │         │ ┌───────────────┐ │
+  // │ │               │ │           │ │              │  │              │ │         │ │               │ │
+  // │ │  channelToB   │◄├───────────┤►│  channelToA  │  │  channelToC  │◄├─────────┤►│  channelToB   │ │
+  // │ │               │ │           │ │              │  │              │ │         │ │               │ │
+  // │ └───────────────┘ │           │ └──────────────┘  └──────────────┘ │         │ └───────────────┘ │
+  // │                   │           │                                    │         │                   │
+  // └───────────────────┘           └────────────────────────────────────┘         └───────────────────┘
+  pure def buildPacket(currentState: State, source: str, dest: str): Packet =
```
thpani, line 70: In connection to above, maybe it helps to disambiguate the naming here.

```suggestion
  pure def buildPacket(currentState: State, sourceChain: str, destChain: str): Packet =
```

thpani (commented):
Awesome spec! 🎉 
Really cool to have this to play around, the trace denoms are definitely the hardest part to understand about ICS-020.

examples/cosmos/ics20/ics20.qnt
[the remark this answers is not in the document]
bugarela: > Is there a reason to carry around denom at all?

The combination of `denom` and `trace` in this type represents the `denom` string we see in IBC packets (i.e. `"transfer/channel/mydenom"`). Since we don't have string manipulation in quint, this is the data structure I came up with.

I'll rename those to look similar to the hashed version of this field, which has the following structure after decoding:
```go
// DenomTrace contains the base denomination for ICS20 fungible tokens and the source tracing
// information
message DenomTrace {
  // chain of port/channel identifiers used for tracing the source of the fungible token
  string path = 1;
  // base denomination of the relayed fungible token
  string base_denom = 2;
}
```

`path` will have to be a list to allow slicing and concatenation, but other than that, I think the names will work better than the ones I picked (from studying the old non-hashed implementation)

examples/cosmos/ics20/ics20.qnt
bugarela: sounds much better, will do!

examples/cosmos/ics20/ics20.qnt
bugarela: Yes! When you suggested isolating this into its own spec, I though we should define something like that so the actual ICS20 spec can call those and decide whether to mint/burn tokens. I'll try to think about a nice interface.

examples/cosmos/ics20/ics20.qnt
bugarela: Ah, right!

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,227 @@
...
+
+  type DenomTrace = {
+    path: List[Hop],
+    baseDenom: str,
+  }
+
+  pure def senderIsSource(sourcePort: str, sourceChannel: str, denomTrace: DenomTrace): bool =
+    denomTrace.path.length() >= 1 and
+    denomTrace.path.head().port == sourcePort and
+    denomTrace.path.head().channel == sourceChannel
```
konnov, line 24:
```suggestion
    denomTrace.path.head() == { channel: sourceChannel, port: sourcePort }
```

Quint lets us write things more concisely than Golang ;-)

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,227 @@
...
+
+  pure def receiverIsSource(sourcePort: str, sourceChannel: str, denomTrace: DenomTrace): bool =
+    not(senderIsSource(sourcePort, sourceChannel, denomTrace))
+
+  pure def updateTrace(denomTrace: DenomTrace, sourcePort: str, sourceChannel: str, destPort: str, destChannel: str): DenomTrace = {
+    if (senderIsSource(sourcePort, sourceChannel, denomTrace)) {
```
konnov, line 30: Should not this be `receiverIsSource` actually?

I still find the naming in ICS20 quite confusing. I know that this is how it called in the IBC specs. However, could we find better names, and say in comments, how these things are called in ICSes?

The confusing thing here is when we check `senderIsSource`, we are not talking about the same "source" as in `sourceChannel` and `sourcePort`. In this context, the source is the chain that has sent the coins in the past, and now we are traversing back.

Would not it make sense to rename "source" in something like "origin"? The point is that a denomination was produced on a path from the origin at some point. On every step, we are either: (1) Getting closer to the origin, or (2) Getting further away from the origin. The source and destination are then reserved for the packet that is being processed.

konnov (commented):
I like this one better. This spec is already clarifying some confusing points. I believe that we should make the naming as clear as possible and map to the original vocabulary of ICSes.

examples/cosmos/ics20/ics20.qnt
bugarela: Makes sense to me! Good call :smile:

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,226 @@
...
+    path: List[Hop],
+    baseDenom: str,
+  }
+
+  pure def senderIsSource(sourcePort: str, sourceChannel: str, denomTrace: DenomTrace): bool =
+    denomTrace.path.length() >= 1 and
+    denomTrace.path.head() == { channel: sourceChannel, port: sourcePort }
+
+  pure def receiverIsSource(sourcePort: str, sourceChannel: str, denomTrace: DenomTrace): bool =
+    not(senderIsSource(sourcePort, sourceChannel, denomTrace))
```
thpani, line 26: Related to [Igor's comment](https://github.com/informalsystems/quint/pull/778/files#r1159521018), I think these got mixed up – if the denom trace is prefixed with the `{sourcePort}/{sourceChannel}`, it means the packet arrived through this channel and the sending chain is a sink.

bugarela: Right! I was just trying to understand what I got wrong here, and yep, I switched those up. Fix incoming!

bugarela: WDYT of the name `goingTowardsOrigin` [here](https://github.com/informalsystems/quint/pull/778/commits/9fa9b8fa3102af9f682878b9d106bfd868fbf726)? I find this much easier to reason about (thanks @konnov for the name suggestion).

thpani: Great idea! Maybe even `returningToOrigin`?

bugarela: I like `returningToOrigin`, but also feel like `goingTowardsOrigin` is more "to the point".  Returning can be seen as something about the whole trace, while this is about a single step: while it might be going towards the origin now, it might go in a different path in the next step and never return to the origin at all.

bugarela:
I'll go ahead and merge this, and then work on predicates and tests on channel setup in a follow-up

thpani (approved):
Looks great! 🤓

I think we should still add comments to the main definitions, to make it more understandable without reading the English spec.

examples/cosmos/ics20/ics20.qnt
bugarela: Hmm, that makes sense. This is actually not completely local, as it is looking at where it came from. `returning` makes a lot of sense now :smile:

thpani: What about `isBacktracking`?

bugarela: `isBacktracking` is good, I like it. Another idea: the [comment](https://github.com/cosmos/ibc-go/blob/457095517b7832c42ecf13571fee1e550fec02d0/modules/apps/transfer/keeper/relay.go#L18-L49) in the reference implementation describes this in terms of going forward or going back in the token timeline. I think this is a good way of explaining it, and if we can find a name that relates to that, it would be more coherent with the existing material. I guess `isBacktracking` is already close to that. Maybe something like `movingBackInTimeline`?

thpani: I didn't grok the timeline explanation the first time, because physical time is still going forward while a token is passing backward. So the mention of time confused me.
In general, I prefer this approach over the source/sink nomenclature though.

We're on the right track here 😄

A very verbose option would be `movingBackAlongDenomTrace`?

bugarela: Good point and good suggestions. Maybe we can remove `Denom` and use `movingBackAlongTrace` since, in the spec context, trace can only be a denom trace.

bugarela:
@thpani I renamed a bunch of stuff and added some comments. I'm not sure if we should keep the English spec nomenclature references in the quint spec. Maybe it's best to keep this clean and then add a correspondence file (i.e. `movingBackAlongTrace` is the same as `receiverIsSource`).

I tried a different approach at explaining things, but that's just a reflection of my personal understanding. Let me know whether it makes sense to you.

examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -0,0 +1,256 @@
...
+  /// as the source zone. When the token is sent back to the chain it previously
+  /// received from, the prefix is removed. This is a backwards movement in the
+  /// token's timeline and the sender chain is acting as the sink zone."
+  pure def movingBackAlongTrace(hopSource: HopElement, denomTrace: DenomTrace): bool =
+    denomTrace.path.length() >= 1 and
+    denomTrace.path.head() == hopSource
+
+  /// `updateTrace` updates a denomTrace's path either by removing or adding a
+  /// hop. It removes a hop if the token is moving backwards along the trace,
+  /// and adds a hop if the token is moving somewhere else.
```
thpani, line 51: Love this description, so much easier to comprehend! 🎉

thpani (approved):
Looks great! 🤩 

I just have some very small suggestions for the comments.
Feel free to resolve them as you see fit.

PR #779, opened 2023-04-05
RFC007: Design of foreign calls in Quint

konnov:
This ~ADR~ RFC discusses the problem space and the solution space for integrating non-Quint code into Quint specifications. ~The section on the actual solution is work in progress.~ I would like to receive feedback on the Context and Options ~first~.


thpani (commented):
One option I'm missing is messaging.

I think it would solve (at least some of) the issues here more elegantly, in particular those that have an event-driven or pub-sub flavor.

shonfeder:
I haven’t had time to review the ADR in detail, but If we need an RPC framework, then, given the existing uses of gRPC both in Apalache and (iiuc)throughout the cosmos ecosystem, I think we should give very good reasons for introducing a different framework, including discussion of the costs of the complication that would impose.

The single reason given for preferring JSON-rpc is that it doesn’t require proto files. But the protocol files serve the purpose of automating ser/de for all supported language based on a single data model. If JSON-rpc doesn’t use something like this, then is seems to me a trade off, not a clear advantage, since it means having to manually map and verify data.

konnov:
> One option I'm missing is messaging.
> 
> I think it would solve (at least some of) the issues here more elegantly, in particular those that have an event-driven or pub-sub flavor.

Do you mean something like a message queue? I guess, that would work for notifications but would not work for calls where we need the response.

konnov:
> I haven’t had time to review the ADR in detail, but If we need an RPC framework, then, given the existing uses of gRPC both in Apalache and (iiuc)throughout the cosmos ecosystem, I think we should give very good reasons for introducing a different framework, including discussion of the costs of the complication that would impose.

This is true. I think what we often face are the requests by systems engineers who are trying to apply the same standards to different kinds of systems. What works for high throughput low-level systems does not have to be the standard for all kinds of systems. For instance, Ethereum is using JSON-RPC. I guess, Cosmos SDK had it, but I cannot find it anymore.

> The single reason given for preferring JSON-rpc is that it doesn’t require proto files. But the protocol files serve the purpose of automating ser/de and handler/requester method plumbing for all supported language based on a single data model. If JSON-rpc doesn’t use something like this, then is seems to me a trade off, not a clear advantage, since it means having to manually map and verify data.

There is some serde support for JSON-RPC in Golang: https://pkg.go.dev/github.com/statechannels/go-nitro/rpc/serde

konnov:
CometBFT supports JSON-RPC: https://docs.cometbft.com/v0.34/rpc/

thpani:
> Do you mean something like a message queue?

Yes

> I guess, that would work for notifications but would not work for calls where we need the response.

You just need 2 queues 😉: https://www.enterpriseintegrationpatterns.com/patterns/messaging/RequestReply.html

thpani:
But more concretely, my suggestion for messaging is because at least `Example 4: Interactive trace visualization` sounds like you sometimes want the external tool to call into quint.

As in, quint tells the visTool "visualize this". And then quint would need to wait for a reverse RPC call where visTool says "I'm done visualizing, go on exploring the spec".

Whereas, in a message-based system, quint would post a message "visualize this" and go on to produce "visualize that", and visTool would consume them at its own pace.

konnov:
> Whereas, in a message-based system, quint would post a message "visualize this" and immediately go on to produce "visualize that", and visTool would consume them at its own pace.

Right. For visualization, an event queue sounds like a better solution.

bugarela:
I also believe that the ideal communication format is through message queues.

konnov:
As we discussed in the meeting, this is more an RFC than ADR. Renamed it to RFC :)

doc/rfc007-foreign-calls.md
```diff
@@ -0,0 +1,220 @@
...
+these nice features are somewhat in conflict with the expressivity and usability
+of the language, from the systems engineer's point of view. Even though Quint is
+obviously Turing-complete (for example, quantifiers over integers and integer
+arithmetic make many properties undecidable), this does not mean that all of the
+typical engineering tasks are easy to do in Quint.
+
+In the following, we introduce several examples that motivate the need for this
+RFC.
+
+### Example 1: Strings
```
thpani, line 34: For strings, there was another request for string concatenation on [Zulip](https://informal-systems.zulipchat.com/#narrow/stream/265309-apalache/topic/String.20concatenation), but for Apalache rather than Quint.

shonfeder (approved):
Approving and merging so we can incorporate these design  considerations into our design docs.

PR #784, opened 2023-04-06
Add ICS-020 bank module spec

thpani:
Add a bank module for ICS-020

As discussed synchronously yesterday, ICS-020 assumes a bank module, but its assumptions are not documented. So it makes sense to start from a coarse spec, rather than something modeling an implementation like `x/bank` in the Cosmos SDK.

This initial spec is intentionally ignorant. For example, it allows minting/burning/transferring negative amounts. We can refine these definitions as we learn about the assumptions made by ICS-020.

- [ ] ~Ran `npm run format` (or had formatting run automatically on all files edited)~


examples/cosmos/ics20/bank.qnt
```diff
@@ -0,0 +1,240 @@
...
+  /* Return `accounts`, where `address` is updated to hold
+   * `currentAmount + amount` tokens of denomination `denom`.
+   * 
+   * `currentAmount` refers to the number of tokens of denomination `denom` held
+   * by `address` in `accounts`.
+   *
+   * If there currently is no account under `address`, return accounts including
+   * `address` with exactly `amount`-many `denom` tokens.
+   */
+  pure def manipulateAccount(accounts: Accounts, address: Address,
```
konnov, line 56:
```suggestion
  pure def changeBalance(accounts: Accounts, address: Address,
```

Would not that be a better name? Although the operator also adds a new account, this behavior can be understood as adding `amount` to 0. (e.g., all maps are initialized with 0 by default in Solidity, so there are precedents for this behavior).

examples/cosmos/ics20/bank.qnt
```diff
@@ -0,0 +1,240 @@
...
+                             denom: Denomination, amount: Amount) : Accounts = {
+    val accountCoins = accounts.getBalances(address)
+    val newCoins = accountCoins.put(denom, accountCoins.getBalance(denom) + amount)
```
konnov, line 59: Just curious, the ICS specs do not assume any bound on the amounts? Or maybe they assume that there is a bound, but it is different for different implementations?

examples/cosmos/ics20/bank.qnt
```diff
@@ -0,0 +1,240 @@
...
+    accounts.put(address, newCoins)
+  }
+
+  /* Return `accounts`, where `amount`-many tokens of denomination `denom`
+   * vanished out of `sender`'s account into thin air.
+   *
+   * This intentionally does NOT check whether `amount > 0`.
+   */
+  pure def BurnCoins(accounts: Accounts, sender: Address,
```
konnov, line 68:
```suggestion
  pure def burnCoins(accounts: Accounts, sender: Address,
```

Let's just agree on whether we make the first letter capital or not. I would prefer to keep it small (non-capital).

examples/cosmos/ics20/bank.qnt
```diff
@@ -0,0 +1,240 @@
...
+                     denom: Denomination, amount: Amount): Accounts = {
+    manipulateAccount(accounts, sender, denom, -amount)
+  }
+
+  /* Return `accounts`, where `amount`-many tokens of denomination `denom` were
+   * added to `sender`'s account out of thin air.
+   *
+   * This intentionally does NOT check whether `amount > 0`.
+   */
+  pure def MintCoins(accounts: Accounts, receiver: Address,
```
konnov, line 78:
```suggestion
  pure def mintCoins(accounts: Accounts, receiver: Address,
```

examples/cosmos/ics20/bank.qnt
```diff
@@ -0,0 +1,240 @@
...
+  /* Return `state`, where `amount`-many tokens of denomination `denom` were
+   * moved from `sender`'s account into `receiver`'s acount.
+   *
+   * This intentionally does NOT check whether `amount > 0`.
+   */
+  pure def TransferCoins(accounts: Accounts, sender: Address, receiver: Address,
+                         denom: Denomination, amount: Amount): Accounts = {
+    accounts
+      .BurnCoins(sender, denom, amount)
+      .MintCoins(receiver, denom, amount)
```
konnov, line 92: That's a very interesting way of expressing a transfer. I have never seen anything like that before, but it makes a lot of sense!

examples/cosmos/ics20/bank.qnt
```diff
@@ -0,0 +1,240 @@
...
+                     denom: Denomination, amount: Amount): Accounts = {
+    manipulateAccount(accounts, receiver, denom, amount)
+  }
+
+  /* Return `state`, where `amount`-many tokens of denomination `denom` were
+   * moved from `sender`'s account into `receiver`'s acount.
+   *
+   * This intentionally does NOT check whether `amount > 0`.
+   */
+  pure def TransferCoins(accounts: Accounts, sender: Address, receiver: Address,
```
konnov, line 88:
```suggestion
  pure def transferCoins(accounts: Accounts, sender: Address, receiver: Address,
```

examples/cosmos/ics20/bank.qnt
```diff
@@ -0,0 +1,240 @@
...
+  action init = State' = Addresses.mapBy(_ => Map("atom" -> 10))
+
+  /* Nondeterministically burn, mint, or transfer tokens of arbitrary
+   * denomination and amount to/from arbitrary accounts.
+   */
+  action step = {
+    nondet addrA = Addresses.oneOf()
+    nondet addrB = Addresses.oneOf()
+    nondet denom = Denoms.oneOf()
+    nondet amount = Int.oneOf()
```
konnov, line 127:
```suggestion
    // since there are no assumptions about the range of amounts, this can be an arbitrary integer!
    nondet amount = Int.oneOf()
```

examples/cosmos/ics20/bank.qnt
```diff
@@ -0,0 +1,240 @@
...
+    accounts
+      .BurnCoins(sender, denom, amount)
+      .MintCoins(receiver, denom, amount)
+  }
+
+  /* ***************************************************************************
+   * STATE MACHINE
+   * **************************************************************************/
+
+  var State: Accounts
```
konnov, line 99:
```suggestion
  var state: Accounts
```

Alternatives: `accountsStore` or `accountsContext`

konnov (approved):
This looks very nice as a first approximation! I am sure there are many assumptions, e.g., about accounts being non-negative, capped by an upper bound, etc. We should only add these assumptions when they are spelled out in the ICSes.

examples/cosmos/ics20/bank.qnt
thpani: Since we have no access modifiers, I made the "public" ones uppercase and the "private" ones lowercase. I think it makes it easier to see what the actual interface is. Do you have other suggestions?

examples/cosmos/ics20/bank.qnt
thpani: So in ICS-020, these will be called with an `uint256`.
So I guess there is an implicit assumption that the bank handles _at least_ uint256.

examples/cosmos/ics20/bank.qnt
thpani: Renamed to `addToBalance`, I think that's most accurate

examples/cosmos/ics20/bank.qnt
konnov: Well, then we know the range of amounts. You could do something similar to: https://github.com/informalsystems/quint/blob/eeac951979a14b3fbba8266b3501857abc6c2a62/examples/solidity/ERC20/erc20.qnt#L23-L27

konnov: > So I guess there is an implicit assumption that the bank handles _at least_ amounts in 0..2^256-1.

Oh, that's interesting. So even if the amounts are in the range, it does not impose a bound on the balances. Maybe you should make a note about this and refer to the place where ICS20 refers to uint256?

thpani: Well, we know them for the ICS-020 calling code. So I think this sort of logic belongs there, not in the bank module.

konnov (commented):
More comments :)

examples/cosmos/ics20/bank.qnt
thpani: Aaah, well, but nothing says the bank cannot operate over arbitrary integers.
I guess we can only document this assumption (or, do two bank specs, one with bounded and one with unbounded ints).

josef-widder: Also I guess all the amounts are implicitly assumed to be non-negative? Might be good to also note that somewhere, as the minimal logic doesn't capture this.

thpani: There's a comment in pseudocode in ICS-020 saying `TransferCoins` fails with insufficient balance, so I guess yes. OTOH, there is no such comment for `BurnCoins`...

examples/cosmos/ics20/bank.qnt
bugarela: I'm not a big fan of this. Even though it works and makes sense at this specification's abstraction level, it might be misleading to people using this spec to understand how transfers work. Since this isn't specially clearer or significantly more concise than writting:
```suggestion
      .addToBalance(sender, denom, -amount)
      .addToBalance(receiver, denom, amount)
```

I'd avoid using `mint` and `burn` abstractions for expressing it.

This might me too much of a personal preference here, and even if this actually is confusing to people, this shouldn't be a reference for bank module implementations anyhow. So I'm fine either way :smile:

thpani: I was undecided about this, so let's go with @bugarela's suggestion.

examples/cosmos/ics20/bank.qnt
thpani: Checking for sufficient balances in [d5fee40](https://github.com/informalsystems/quint/pull/784/commits/d5fee40b52cd97d3853a63100b0b03cde2826360).

thpani: re: Igor's comments, I've added a README to document any assumptions we make: [9411f5e](https://github.com/informalsystems/quint/pull/784/commits/9411f5e705ae56dcb06c41cf4dcfc6de7bdadbb4)

examples/cosmos/ics20/bank.qnt
thpani: The pseudo-go in the English spec uses this casing, so I'd rather keep it in sync.

PR #786, opened 2023-04-06
Run non-deterministic tests multiple times

### LINKED ISSUE #678: Run non-deterministic tests multiple times
Currently, the tests that contain non-deterministic choice are run once. These tests were designed to be similar to property-based tests, e.g., like [fuzz tests in Foundry](https://book.getfoundry.sh/forge/fuzz-testing#fuzz-testing). We should check, whether a test contains `nondet`, which means that this test produces inputs non-deterministically. If a test contains `nondet`, it should be run multiple times.

konnov:
Closes #678. This PR runs non-deterministic tests up to `--max-samples` times, which is 10000 by default. Previously, non-deterministic tests were run just once. That was not very PBT-like :)


shonfeder (approved):
Nice!

PR #791, opened 2023-04-10
Extend flattening to support `export`

bugarela:
Hello :octocat: 

We all noticed some weird things breaking on the compiler last week, and that's because there were many scenarios (mostly involving `export`) where the flattening procedure was not doing the right thing. This fixes a bunch of those and introduces more tests, including integration tests for the `language-feature` examples for imports and instances (which were only being parsed and typechecked until now). 

### DIFF
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -645,13 +645,14 @@ true
 
 <!-- !test in compile instances -->
 ```
-echo "A2::g(false)" | quint -r ../examples/language-features/instances.qnt::Instances 2>&1 | tail -n +3
+echo -e "A1::f(1)\nA2::f(1)" | quint -r ../examples/language-features/instances.qnt::Instances 2>&1 | tail -n +3
 ```
 
 <!-- !test out compile instances -->
 ```
 true
 
->>> true
+>>> 34
+>>> 16
 >>> 
 ```
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -625,3 +625,33 @@ exit $exit_code
 ```
 error: --seed must be a big integer, found: NotANumber
 ```
+
+### OK on compile imports
+
+<!-- !test in compile imports -->
+```
+echo "init" | quint -r ../examples/language-features/imports.qnt::G 2>&1 | tail -n +3
+```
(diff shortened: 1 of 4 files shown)


quint/io-cli-tests.md
```diff
@@ -625,3 +625,33 @@ exit $exit_code
...
+
+>>> true
+>>> 
+```
+
+### OK on compile instances
+
+<!-- !test in compile instances -->
+```
+echo "A2::g(false)" | quint -r ../examples/language-features/instances.qnt::Instances 2>&1 | tail -n +3
```
konnov, line 648: `g` has the same behavior in `A1` and `A2`. Would it make sense to test `f` too, which has different behaviors in `A1` and `A2`?

quint/src/flattening.ts
```diff
@@ -134,14 +148,15 @@ type DefAfterFlattening = (
 ) & WithOptionalDoc
 
 function isFlattened(def: QuintDef): def is DefAfterFlattening {
-  return def.kind !== 'instance' && def.kind !== 'import'
+  return def.kind !== 'instance' && def.kind !== 'import' && def.kind !== 'export'
```
konnov, line 151: I am a bit confused by this name. So any definition different from an instance, `import`, and `export` will be considered flattened, even before flattening. Maybe we should rename the method to `isFlat`?

quint/test/flatenning.test.ts
```diff
@@ -134,4 +134,85 @@ describe('flatten', () => {
...
+  describe('export instance with qualifier', () => {
+    const baseDefs = [
+      'const N: int',
+      'val x = N + 1',
+    ]
+
+    const defs = [
+      'val myN = 1',
+      'import A(N = myN) as A1',
+      'export A1 as B',
```
konnov, line 203: That's interesting. What happens, if we have `var y: int` in the module? My expectation is that it should not appear twice, but it is somehow should be visible both under `A1::y` and `B::y`.

konnov (approved):
Thanks for fixing this! I still have to have a careful look at the overall flattening process. I like that you have added more tests. This is probably the way for us to understand what we actually expect from `import` and `export`.

PR #800, opened 2023-04-13
The first implementation of `import ... from`

### LINKED ISSUE #8: Support for multiple files
As outlined in [ADR001](https://github.com/informalsystems/tnt/blob/main/doc/adr001-transpiler-architecture.md#33-resolve), we have to implement the task "resolve", which would resolve module imports.

konnov:
Closes #8. This is the first batch of support for `import .. from '<filename>'`. I had to keep balance between hacking it quickly and making it not too hacky. ~This PR only has happy path tests.~ The construct itself is working now. ~But we would need a follow up PR that adds tests for cyclic imports and other error processing~.

~Now it is working with `parse` but will probably fail with `run`, `test`, and REPL.~

It is working with `parse`, `run`, `test`, and vscode. REPL requires some googling #829.

Follow-up issues: #829, #830, #831, #832.


quint/src/quintParserFrontend.ts
```diff
@@ -128,6 +129,101 @@ export function parsePhase1(
...
+/**
+ * Phase 1b of the Quint parser. Go over each definition of the form
+ * `import ... from '<path>'`, do the following:
+ * 
+ *  - parse the modules that are refferred by each path,
+ *  - add the parsed modules.
+ * 
+ * Cyclic dependencies among different files are reported as errors.
+ */
+export function parsePhase1b(
```
bugarela, line 141: I got a sense of what this is doing, but how/why it works is not completely clear to me. It's 85 lines of imperative code, and I find it hard to follow. I'd much rather see this broken down into functions so I can reason about cycle detection, sorting and recursion separately.

I don't want to block this since we are just trying to get this working to unblock other work, but I do want to note the problem.

quint/src/sourceResolver.ts
```diff
@@ -0,0 +1,166 @@
...
+  load: (lookupPath: SourceLookupPath) => Either<string, string>
+
+  /**
+   * Extract the resolver-specific stem from a path, e.g., the directory name
+   * if path is a path in a filesystem.
+   *
+   * @param lookupPath a source lookup path created via `mkLookupPath`.
+   * @returns the stem of a path, e.g., the parent directory
+   */
+  stempath: (lookupPath: SourceLookupPath) => string
```
bugarela, line 72: What is `stem` here? I don't remember seeing it before and couldn't find the meaning with a couple of google searches.

quint/src/sourceResolver.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+/**
+ * A special version of `fileSourceResolver` that replaces source names
+ * with a user-defined string. This is needed for testing the parser,
+ * which otherwise would produce absolute file names.
+ * 
+ * Perhaps, we could have a better solution in the future.
+ */
+export const fileSourceResolverForTests =
```
bugarela, line 112: To avoid this duplication, we could have the `replacer` be an optional parameter to `fileSourceResolver` (with the identity function as its default value) and apply the `replacer` at the original `load call`, getting rid of `fileSourceResolverForTests` entirely.

quint/src/sourceResolver.ts
```diff
@@ -0,0 +1,166 @@
...
+      return contents
+        ? right(contents)
+        : left(`Source not found: '${lookupPath.normalizedPath}'`)
+    },
+
+    stempath: (lookupPath: SourceLookupPath): string => {
+      return dirname(lookupPath.normalizedPath)
+    }
+  }
+}
```
bugarela, line 166:
```suggestion
}

```

quint/test/sourceResolver.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    assert.deepEqual(result,
+      left(`Source not found: '/does-not-exist'`))
+  })
+
+  it('stemname', () => {
+    const r = stringSourceResolver(staticTable)
+    const result = r.stempath(r.lookupPath('/baz', './baz.qnt'))
+    assert.deepEqual(result, '/baz')
+  })
+})
```
bugarela, line 99:
```suggestion
})

```

bugarela (approved):
That's a lot of things you got working in a short amount of time :rocket:. I have some comments, but don't want to block this.

Also, could you add `from` to the list of keywords in our editor plugins?

quint/src/sourceResolver.ts
konnov: It is the directory relative to the path of the importing file. I wanted to call it `basepath` but that is also confusing.

quint/src/quintParserFrontend.ts
konnov: Yeah, it's not always easy to express the same thing in functional code. This code is using the [Worklist algorithm](https://en.wikipedia.org/wiki/Reaching_definition#Worklist_algorithm). The challenge here is to also detect which modules should go first, hence, it assigns ranks to modules. I think I could rewrite it into functional code, but it would be even less obvious than the imperative code.

quint/src/cliCommands.ts
```diff
@@ -9,12 +9,12 @@
...
 import { existsSync, readFileSync, writeFileSync } from 'fs'
 import JSONbig from 'json-bigint'
-import { basename, resolve } from 'path'
+import { dirname, basename, resolve } from 'path'
 import { cwd } from 'process'
 import chalk from 'chalk'
 
 import {
-  ErrorMessage, Loc, compactSourceMap, parsePhase1, parsePhase2
+  ErrorMessage, Loc, compactSourceMap, parsePhase1, parsePhase1b, parsePhase2
```
shonfeder, line 17: can we give the parse phases descriptive names? `1b` doesn't mean anything. how about `loadFiles` or something?

examples/cosmos/ics23/ics23.qnt
```diff
@@ -208,8 +208,8 @@ module ics23 {
     // actualPrefix = len(inner.prefix) - 33
     length(inner.prefix) >= ChildSize,
     // getPosition(0) returns 0
-    val from = length(inner.prefix) - ChildSize
-    inner.prefix.slice(from, from + ChildSize) == EmptyChild
```
shonfeder, line 212: This shouldn't be necessary iiuc? `from` seems to only be parsed out of import statements?

quint/src/quintParserFrontend.ts
```diff
@@ -128,6 +129,101 @@ export function parsePhase1(
   }
 }
 
+/**
+ * Phase 1b of the Quint parser. Go over each definition of the form
+ * `import ... from '<path>'`, do the following:
+ * 
+ *  - parse the modules that are refferred by each path,
```
shonfeder, line 136:
```suggestion
 *  - parse the modules that are referenced by each path,
```

quint/src/quintParserFrontend.ts
shonfeder: If you've referenced that algo in your implementation, would help to add a comment citing that.

quint/src/quintParserFrontend.ts
```diff
@@ -128,6 +129,101 @@ export function parsePhase1(
...
+  sourceResolver: SourceResolver,
+  mainPath: SourceLookupPath,
+  mainPhase1Result: ParserPhase1
+): ParseResult<ParserPhase1> {
+  // we accumulate the source map over all files here
+  let sourceMap = new Map(mainPhase1Result.sourceMap)
+  // The list of modules that have not been been processed yet.
+  // Each element of the list carries the module to be processed and the trail
+  // of sources that led to this module.
+  let worklist: [QuintModule, SourceLookupPath[]][] =
```
shonfeder, line 152: It doesn't look like you reassign this variable. iiuc, this should be a const. it is a constant reference to a mutable structure: not a mutable reference. afaik, you'll still be able to push into it.

```suggestion
  const worklist: [QuintModule, SourceLookupPath[]][] =
```

shonfeder (changes requested):
Quick turnaround here! :muscle: 

I'm part way thru, but would like to have the first baths of questions/suggestions addressed before merging.

quint/src/cliCommands.ts
```diff
@@ -182,7 +183,13 @@ export function load(args: any): CLIProcedure<LoadedStage> {
 export function parse(loaded: LoadedStage): CLIProcedure<ParsedStage> {
   const { args, sourceCode, path } = loaded
   const parsing = { ...loaded, stage: 'parsing' as stage }
-  return parsePhase1(newIdGenerator(), sourceCode, path)
+  const idgen = newIdGenerator()
+  return parsePhase1(idgen, sourceCode, path)
+    .chain(phase1Data => {
+      const resolver = fileSourceResolver()
```
bugarela, line 189: Could we instantiate this resolver inside `parsePhase1b`? It would be great if the top-level calls are as simple as possible since we will have to replicate this in the language server.

examples/cosmos/ics23/ics23.qnt
konnov: The ANTLR grammar introduces  a new token called 'from'. Once we have it, `from` should be a reserved keyword.

shonfeder:
Posted my last batch of pending comments and removed my change request.

shonfeder:
Not to block, but to record my uncertainty for the record, and maybe so I'd learn the motivation for the approach here:

The general thing I don't understand about this PR's approach is why we cannot approach the problem here in a shape a bit more like this:

- Do a pass to recursively phase1 parse all files imported by file reference
- in the process, rewrite relative file imports paths to absolute paths
- Extend the phase1 result with a mapping `absFilePath -> IR`
- Use this structure for dependency ordering (this part likely using the imperative algorithm, but only for benign state, without need for I/O. This is really just a special case of topological sorting iiuc?).
- Pass this ordered collection of modules on for module resolution.

The main benefits of this approach to my mind are

- Easier to reason about the separate operations.
- Don't need to mix file I/O into the relatively subtle dep resolution code

quint/src/quintParserFrontend.ts
```diff
@@ -128,6 +129,101 @@ export function parsePhase1(
...
+ * Cyclic dependencies among different files are reported as errors.
+ */
+export function parsePhase1b(
+  idGen: IdGenerator,
+  sourceResolver: SourceResolver,
+  mainPath: SourceLookupPath,
+  mainPhase1Result: ParserPhase1
+): ParseResult<ParserPhase1> {
+  // we accumulate the source map over all files here
+  let sourceMap = new Map(mainPhase1Result.sourceMap)
```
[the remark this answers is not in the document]
shonfeder: but why? Anything that has a reference to the map thru this variable will have it's ref mutated anyhow, so why mutate the reference and create a copy of a mutable structure instead of just mutating the mutable structure? Does something at some point keep a reference to the previous map?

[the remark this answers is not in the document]
konnov: I do not want to mutate the output of phase1 directly. Phase 1 is called multiple times, once per file. So instead of collecting the source information in the source map of the main file, I am making a copy here.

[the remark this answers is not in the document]
shonfeder: sorry if i'm being dense, but i'm afraid I don't follow: you are already copying the Phase1 output into a map via the `new Map` call, no? Each time you get a new Phase1 result you are adding it to this map. Whether you mutate this map or mutate the ref to add a new map, you are not touching the phase1 result in directly either way, are you?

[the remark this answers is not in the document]
shonfeder: To be clear, this isn't a blocking objection, but i just don't see the benefit to constantly creating copies if you are mutating the state anyhow.

quint/src/sourceResolver.ts
```diff
@@ -0,0 +1,166 @@
...
+   * Produce the source name that will be shown in errors and stored in source maps.
+   * @returns a string representation of the path that is written in source maps
+   */
+  toSourceName: () => string
+}
+
+/**
+ * A general interface for resolving sources.
+ */
+export interface SourceResolver {
```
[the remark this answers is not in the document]
konnov: I've introduced this abstraction, as it has been quite an issue in SANY. We do not have to deal with filename, in order to resolve relative paths. In case of tests, we could use a map of strings. For a web interface, we could use a map of names to URIs. It's true that the main use case is filenames right now.

quint/src/sourceResolver.ts
```diff
@@ -0,0 +1,166 @@
...
+ */
+export const fileSourceResolver = (): SourceResolver => {
+  return {
+    lookupPath: (basepath: string, importPath: string) => {
+      return {
+        normalizedPath: posix.join(basepath, importPath),
+        toSourceName: () => {
+          return posix.join(basepath, importPath)
+        },
+      }
```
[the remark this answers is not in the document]
konnov: toSourceName produces a platform-independent path, since it is using `posix.join`, the paths will look the same on Linux and Windows. I actually had this issue with tests.

quint/src/cliCommands.ts
konnov: Yeah, I have renamed all phases

quint/src/cliCommands.ts
konnov: I agree with the intention. However, the resolver is a very important part here. I can imagine that we would like to load some source files differently very soon, e.g., the standard library. Most likely, this would require a different resolver. So it would be good to propagate the resolver that is constructed elsewhere.

shonfeder (approved):
As per our convo, approving to make clear I’m fine with this being merged to unblock work, my other concerns notwithstanding.

quint/src/sourceResolver.ts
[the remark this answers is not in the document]
shonfeder: I see. That makes sense. Thanks.

quint/src/sourceResolver.ts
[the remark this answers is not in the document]
shonfeder: But why the same computation twice , just once in a thunk? You already have the result, so why ever call `toSourceName`?

quint/src/quintParserFrontend.ts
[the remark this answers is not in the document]
konnov: > sorry if i'm being dense, but i'm afraid I don't follow: you are already copying the Phase1 output into a map via the `new Map` call, no? Each time you get a new Phase1 result you are adding it to this map. Whether you mutate this map or mutate the ref to add a new map, you are not touching the phase1 result in directly either way, are you?

oh, that's what you mean. This is how I am extending this map:

```js
        sourceMap = new Map([...sourceMap, ...parseResult.value.sourceMap])
```

quint/src/sourceResolver.ts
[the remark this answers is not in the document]
konnov: Actually, I just have hit this problem with the VSCode plugin. @bugarela and I agreed to handle only `file://` atm, but we should check how it is done with URIs in general, postponed for #831.

konnov:
I have addressed most of the comments, also renamed the parsing phases. It took me a bit longer than expected. But imports seem to work as expected now. Very exciting feature 🔥

The only annoying thing is that REPL does not use the current shell directory to load imports. Postponed for #829.

Merging this PR to unblock @bugarela and @thpani on the IBC work.

PR #806, opened 2023-04-13
suppress comments in REPL

### LINKED ISSUE #413: Support comments in REPL
We should support and discard comments entered in the repl. This is usual
behavior, and allows copying in blocks of code verbatim.

Currently I am getting

```sh
>>> // foo
syntax error: <input>:1:1 - error: mismatched input '/' expecting {'module', '{', 'const', 'var', 'assume', 'type', 'import', 'val', 'def', 'pure', 'action', 'run', 'temporal', '[', 'all', 'any', 'if', 'nondet', '_', STRING, BOOL, INT, 'and', 'or', 'iff', 'implies', 'Set', 'List', 'Map', '-', '(', IDENTIFIER}
1: // foo
```

### LINKED ISSUE #414: REPL input does not support multiline value definitions
``` sh
>>> val kettleState = {
syntax error: <input>:1:20 - error: no viable alternative at input 'valkettleState={'
1: val kettleState = {
```

We should use the same input mode switch as we do for `def`'s

konnov:
Closes #413. The feature that has been requested for some time. REPL simply ignores:

 - one line comments such as `// aaaa`
 - doc comments such as `/// aasdsdafa`
 - multiline comments such as:
   ```js
         /* multi
              line
                comment
         */
   ``` 


bugarela (approved):
LGTM :sparkles:

PR #813, opened 2023-04-14
Tests for exploring different channel setups in ICS20

bugarela:
Hello :octocat: 

ICS20 assumes that channel names are unique per chain. In this PR, I'm adding some PBT tests for two different scenarios: when the channel setup for the chains is random and when it satisfies the assumption.

When running a simulation scenario with a random channel setup, the test finds an error. For example. with seed `--seed=0xfd4eb918724e7`, the first execution that breaks the assertion has this channel setup:
```
Map("A" -> Map("A" -> "channel-2", "B" -> "channel-2", "C" -> "channel-2"), "B" -> Map("A" -> "channel-2", "B" -> "channel-2", "C" -> "channel-3"), "C" -> Map("A" -> "channel-2", "B" -> "channel-2", "C" -> "channel-2"))
```

(command for reproducing: `quint test --main=randomChannelsPBT --verbosity=3 --max-samples=36 --seed=0xfd4eb918724e7 ../examples/cosmos/ics20/denomTrace.qnt`)

Where we can see that chain `"A"` has the same `"channel-2"` for both "B" and "C". With this setup, the denom trace is not updated correctly according to the fixed transfer sequence, so the token would be not recovered properly by its origin chain.

When running the same simulation scenario with a random chain setup that **satisfies** the unique per chain assumption, then the assertion is always satisfied (at least for the sample I've tried): `quint test --main=properChannels
PBT --verbosity=3 --max-samples=36 ../examples/cosmos/ics20/denomTrace.qnt`

Closes https://github.com/informalsystems/apalache-team-services/issues/3


examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -67,6 +67,20 @@ module denomTrace {
...
     }
   }
+
+  pure def isProperChannelSetup(setup: str -> str -> str): bool =
+    pure val chainNames = Set("A", "B", "C") // should be channelSetup.keys(), but that's breaking the effect checker somehow (see issue #808)
+    and {
+      // Check the domain for the second level of chains
+      chainNames.forall(chain => setup.get(chain).keys() == chainNames),
+      tuples(chainNames, chainNames).forall((fromChain, toChain) =>
+        fromChain != toChain implies // we dont care about channels from chain to itself
```
bugarela, line 77: I couldn't think of a way to define the set of all possible channel setups without having channels from chains to themselves in it. If you can think of a way, please let me know. My solution was to ignore them in here, but it's annoying that they appear in the output (i.e. `Map("A" -> Map("A" -> "thing-I-dont-care-about" ...) ...)`

examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -173,6 +188,87 @@ module denomTraceTest {
...
+
+  const channelPredicate: (str -> str -> str) => bool
+
+  pure val channelNames = Set("channel-1", "channel-2", "channel-3")
+
+  /// All possible combination of channels by chain satisfying `channelPredicate`
+  pure val possibleChannelSetups = chains.setOfMaps(chains.setOfMaps(channelNames)).filter(channelPredicate)
+
+  var trace: DenomTrace
+  var channelSetup: str -> str -> str
```
bugarela, line 212: I had to make this a variable in order to pick one value from the `possiblChannelSetups` and make it fixed for the entire run. If you can think of a better way, lmk.

konnov: Yeah, currently, it is the best way to do it. Perhaps, we could add something similar to `--cinit` of Apalache

examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -173,6 +188,87 @@ module denomTraceTest {
...
+  run simulationTest = init.then(step.repeated(6)).then((all {
+    def show(a) = true { show(channelSetup) },
+    assert(trace.path == []),
+    unchanged,
+  }))
+}
+
+// Fails because the channel setup is not proper
+// The aim is to show that it is, in fact, necessary to enforce `isProperChannelSetup`
+module randomChannelsPBT {
```
konnov, line 257:
```suggestion
module randomChannelsTests {
```

Not everyone knows what PBT means, and it's also less important that it's about tests

examples/cosmos/ics20/denomTrace.qnt
josef-widder: In the synopsis of the [connection](https://github.com/cosmos/ibc/tree/main/spec/core/ics-003-connection-semantics) spec it says a connection is between two separate chains. As channels are linked to connections, I guess self-loops are ruled out here. Although I didn't find the check in the connection spec. But it might make sense to dig there.

thpani: [ICS 4](https://github.com/cosmos/ibc/blob/8d9d3b6fe7309b034df8457760b3bbd11d24b8e1/spec/core/ics-004-channel-and-packet-semantics/README.md?plain=1#L42) (bold emphasis is mine):

> A *channel* is a pipeline for exactly-once packet delivery between specific modules on **separate blockchains**, which has at least one end capable of sending packets and one end capable of receiving packets.

examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -67,6 +67,21 @@ module denomTrace {
       { path: [hopDestination].concat(denomTrace.path), baseDenom: denomTrace.baseDenom }
     }
   }
+
+  /// `isProperChannelSetup` is true iff channel names are unique by chain
+  pure def isProperChannelSetup(setup: str -> str -> str): bool =
```
thpani, line 72: I was unable to infer the meaning of the `setup` parameter without reading the code.
I think the comment here should refer to and explain what's encoded in the `str -> str -> str` map.

examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -173,6 +188,87 @@ module denomTraceTest {
   }))
 }
 
+module denomTraceSimulation {
```
thpani, line 191: I think it would help to include a high-level intro comment of what's achieved by this module.

Since we don't have additional documentation, maybe also include the commands used to run it from the PR description?

thpani (commented):
This has already been merged, but I was reviewing it to bring myself up to speed:

The code looks good, but I think it could have used a lot more comments – I essentially had to read everything twice to make sense of it.

PR #825, opened 2023-04-17
Remove `notin` operator

### LINKED ISSUE #814: `notin` is in a quantum superposition of existence
As @cason noted in our internal slack, `notin` doesn't work in the REPL, but it is documented and works in the LSP.

@konnov thought `notin` was removed, but I don't think this is so. We demoted it from an infix operator, but it is still defined as a builtin:

```
$  rg notin
src/types/builtinSignatures.ts
47:  { name: 'notin', type: '(a, Set[a]) => bool' },

src/definitionsCollector.ts
40:    { kind: 'def', identifier: 'notin' },

src/ToIrListener.ts
473:        if (name === 'in' || name === 'notin' ||

src/effects/builtinSignatures.ts
134:  { name: 'notin', effect: standardPropagation(2) },

src/builtin.qnt
114:  /// `e.notin(s)` is true when the element `e` is not in the set `s`.
121:  /// assert(4.notin(Set(1, 2, 3)))
122:  /// assert(not(1.notin(Set(1, 2, 3))))
124:  pure def notin(a, s): (a, Set[a]) => bool

testFixture/SuperSpec.qnt
197:  val notin_2_empty = 2.notin(Set())
```

I have also proposed removing it as a builtin in #388, but I don't think this was put into place. 

Where should we land with this operator?

1. Add support in the REPL
2. Remove it from everywhere else

I'm in favor of (2), if we offer this operator, it should be defined IMO.

shonfeder:
Closes #814

-------


konnov (approved):
Thanks for taking care of it!

PR #827, opened 2023-04-18
Add spells as a precursor of the standard library

konnov:
As we have discussed many times, it would be great to have the standard library. We are still lacking infrastructure for introducing such a library, e.g., we would have to import from the Quint package somehow.

For the moment, l have introduced three modules in `./examples/spells`: basic spells, common spells, and rare spells.


examples/README.md
```diff
@@ -21,6 +21,11 @@ yet. To set your expectations right, check the dashboard below first.
  - [puzzles](./puzzles). These are logical puzzles. Some people find them nice
    for learning new languages. If you are one of these people, check the
    puzzles.
+
+ - [spells](./spells). These are nice small definitions that make specification
+   writing easier. We collect them here. One day some of them will become the
+   standard library. If you think you have invented a nice spell that would
+   help other, [contribute your spell](./spells/contribute-your-spell.md).
```
shonfeder, line 28:
```suggestion
   help others, [contribute your spell](./spells/contribute-your-spell.md).
```

examples/spells/basicSpells.qnt
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * This module collects definitions that are ubiquitous.
+ * One day they will become the standard library of Quint.
+ */
+module basicSpells {
+  /**
+   * An annotation for writing preconditions.
+   * @param __cond condition to check
+   * @returns true if and only if __cond evaluates to true
+   */
```
bugarela, line 11: When I was writing `builtin.qnt`, I proposed we used a rust-like doc comment format, with triple slashes (`///`) and markdown inside. Currently, that sort of comment is parsed as a special comment and it's contents are assigned as a definition's documentation (that can be converted to markdown and visualized in overlays in VSCode).


I think this other JSDoc like format has some pros and cons against the triple-slash format. My question here is: do we want to introduce a new documentation format now? Do we want to switch from the one we already have?

I'm ok with changing the format, I'd just like to be consistent here (i.e. this file vs. `builtin.qnt`).

konnov: ah, that's what I was looking for. Yes, let's use `///`. Would it work with `@param` etc.?

konnov: I have changed the formatting to `///`. Afaik, it does not work in front of a `module`, right?

bugarela: > ah, that's what I was looking for. Yes, let's use `///`. Would it work with `@param` etc.?

Not really, it assumes the input is simple markdown. We can have some special rules there if we want.

bugarela: > I have changed the formatting to `///`. Afaik, it does not work in front of a `module`, right?

Not yet :disappointed:

konnov:
> LGTM
> 
> One spell I've seen appear over and over is `Map.getOrDefault`, to treat a map as a total function.

Oh yes, that's a good point. I will add `getOrElse`

konnov:
I have also added the definition of `has`, which is quite handy too:

```
  pure def has(__map: a -> b, __key: a): bool = {
    __map.keys().contains(__key)
  }
```

PR #834, opened 2023-04-18
fix imports in repl

### LINKED ISSUE #829: Use the current working directory in REPL when importing files
A follow-up of #800. `import A from "./A"` in REPL should use the current working directory. Node's `getcwd()` returns the directory where the Quint script is located. We have to find a way to use the shell cwd instead.

konnov:
Closes #829. There was actually nothing wrong with `cwd()`. I just did not use it in the right place. Fixed.


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `quint repl` is printing the same version number as returned by `quint --version` (#804)
 - add the command `.seed` in REPL (#812)
 - fix `quint run` to output compile errors again (#812)
+- fix `import from` in REPL (#834)
```
shonfeder, line 20: I think We don’t need a release not for this because the break wasn’t in a release?

shonfeder (approved):

PR #837, opened 2023-04-19
cache pure val

### LINKED ISSUE #807: Module-level `pure val` definitions are not cached
We should cache those definitions in the compiler to avoid recomputing them every time.

This is an issue for me because it significantly affects performance of my tests, since a large set is recomputed for each sample.

konnov:
Closes #807. This PR adds caching of top-level `pure val`, similar to caching of nested `pure val`. There is not much we can test here.


bugarela (approved):
Looking forward to see the performance improvement on `properChannelTests` in `denomTrace.qnt` :rocket:

PR #839, opened 2023-04-20
fix #838: the order of keys in a record constructor must be irrelevant

### LINKED ISSUE #838: Record equality depends on the order of the keys in REPL
```
>>> {a:1, b:2} == {b:2, a:1}
false
```

I'd expect this to be `true`.

konnov:
Closes #838. I expected `OrderedMap` to do it automatically. Unfortunately, it does not. Hence, sorting the map by keys when constructing it.


CHANGELOG.md
```diff
@@ -31,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Proper errors are now reported when a lambda returns an operator (#811)
+- Fix `quint run` to output compile errors again (#812)
+- Fix the order constructor, so the key order does not matter (#839)
```
bugarela, line 34:
```suggestion
- Fix the record constructor, so the key order does not matter (#839)
```

PR #840, opened 2023-04-20
ICS20 spec integrating `bank` and `denomTrace`

bugarela:
Hello :octocat: 

This adds a first complete version of the ICS20 spec, integrating with the existing `bank.qnt` and `denomTrace.qnt` specs. I've also extracted some common definitions to `base.qnt`.

The commit history is a mess because I'm afraid of rebasing this (since I've started working on top of two existing branches), but the diff is correct.

I'll do some other iterations on this spec, adding invariants and some non-happy-path examples.

Closes https://github.com/informalsystems/apalache-team-services/issues/4

### DIFF
--- a/examples/cosmos/ics20/ics20.qnt
+++ b/examples/cosmos/ics20/ics20.qnt
@@ -47,14 +47,6 @@ module ics20 {
     channelEscrowAddresses: Channel -> Address,
   }
 
-  /// `onRecvPacket` should return an acknowledgment, but it also has to update
-  /// the chain state, so the return type is the updated value for the chain
-  /// state and the acknowledgment
-  type ReceiveTokenPacketResult = {
-    chainState: ChainState,
-    acknowledgement: FungibleTokenPacketAcknowledgement
-  }
-
   /// The counterparty for a channel `C` in a chain is the channel identifier of
   /// the channel `C` connects to, in the other chain.
   pure def getCounterparty(chainState: ChainState, sourceChannel: Channel): Channel = {
@@ -93,38 +85,33 @@ module ics20 {
     }
   }
 
-  pure def onRecvPacket(chainState: ChainState, packet: Packet): ReceiveTokenPacketResult = {
+  /// `onRecvPacket` should return an acknowledgment, but it also has to update
+  /// the chain state, so the return type is the updated value for the chain
+  /// state and the acknowledgment
+  pure def onRecvPacket(chainState: ChainState, packet: Packet): (ChainState, FungibleTokenPacketAcknowledgement) = {
     pure val data = packet.data
-    if (movingBackAlongTrace({port: packet.sourcePort, channel: packet.sourceChannel}, data.denom)) {
+    pure val result = if (movingBackAlongTrace({port: packet.sourcePort, channel: packet.sourceChannel}, data.denom)) {
       // unescrow tokens to receiver
       pure val escrowAccount = chainState.channelEscrowAddresses.get(packet.destChannel)
       pure val receiverDenom = updateTrace(data.denom, packet.sourcePort, packet.sourceChannel, packet.destPort, packet.destChannel)
-
+      
       pure val bankResult = chainState.bank.TransferCoins(escrowAccount, data.receiver, receiverDenom, data.amount)
-
-      pure val newChainState = chainState.with("bank", bankResult.accounts)
-      pure val ack = if (bankResult.success) {
-        { success: true, errorMessage: "" }
-      } else {
-        { success: false, errorMessage: "transfer coins failed" }
-      }
-
-      { chainState: newChainState, acknowledgement: ack }
+      (bankResult, "transfer coins failed")
     } else {
       // mint vouchers to receiver
       pure val newDenom = updateTrace(data.denom, packet.sourcePort, packet.sourceChannel, packet.destPort, packet.destChannel)
 
       pure val bankResult = chainState.bank.MintCoins(data.receiver, newDenom, data.amount)
-
-      pure val newChainState = chainState.with("bank", bankResult.accounts)
-      pure val ack = if (bankResult.success) {
-        { success: true, errorMessage: "" }
-      } else {
-        { success: false, errorMessage: "mint coins failed" }
-      }
-
-      { chainState: newChainState, acknowledgement: ack }
+      (bankResult, "mint coins failed")
     }
+    pure val newChainState = chainState.with("bank", result._1.accounts)
+    pure val ack = if (result._1.success) {
+      { success: true, errorMessage: "" }
+    } else {
+      { success: false, errorMessage: result._2 }
+    }
+
+    (newChainState, ack)
   }
 }
 
@@ -135,10 +122,10 @@ module ics20Test {
   /// Map from chain identifiers to their state
   var chainStates: str -> ChainState
 
-  /// Map from chain identifiers to their connections.
-  /// A connection is a map from counterparty chain identifiers to the local chain's channel identifier.
-  /// For example, chain A connects to chain B through channel "channelToB".
-  pure val connections = Map(
+  /// Map from chain identifiers to a map of the chain identifiers it can
+  /// communicate with and the channel to be used to send packets. For example,
+  /// chain A connects to chain B through channel "channelToB".
+  pure val channelTopology = Map(
     "A" -> Map(
       "B" -> "channelToB",
       "C" -> "channelToC"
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `quint run` returns a non-zero exit code on invariant violation and errors (#793)
 - Support for imports from files: `import ... from <filename>` (#800)
 - suppress comments in REPL (#806)
+- Support for imports from files: `import ... from <filename>` (#800)
```
shonfeder, line 17: is this picked up from an earlier commit? maybe this PR needs a rebase?

shonfeder (approved):

bugarela:
Weird, I ran integration tests a bunch of times locally and no issue. Good thing I have a seed, will take a look later.

CHANGELOG.md
bugarela: ah weird! I must have messed this up in merging. I tried to avoid rebasing because I'm not sure how it will behave since I created this branch on top of Thomas' and then merged Igor's WIP branch on it before starting to work. Perhaps it will be easy to rebase if I just drop all of Igor's commits, as the ics20 files shouldn't have a lot of conflicts. I'll try. If it's to messy, I'll just remove this and squash merge :stuck_out_tongue:

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,308 @@
...
+  var moduleStates: str -> ModuleState
+  var inFlightPackets: Set[Packet]
+
+  /// All channels available in a module. Should be extracted to a channels
+  /// quint module.
+  pure def channelsForModule(mod: str): Set[Channel] = {
+    pure val channels = Map(
+      "A" -> Set("channelToB", "channelToC"),
+      "B" -> Set("channelToA", "channelToC"),
+      "C" -> Set("channelToA", "channelToB")
```
thpani, line 165: I don't think the same channel can be used by different modules.
Did you check?

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,308 @@
...
+      channelEscrowAddresses: Set("channelToA", "channelToB", "channelToC").mapBy(_ => "escrow_account"),
+    }).setBy("A", st => st.with("bank", Map("alice" -> Map(toDenom("atom") -> 100)))),
+    inFlightPackets' = Set()
+  }
+
+  /// A set of all possible combinations of module, account and denom with
+  /// ammount greater than 0, to be used to send a valid transfer. If we just
+  /// pick these values at random, it is most likely that the bank module will
+  /// refuse the transfer for lack of funds, and we won't be able to find a
+  // valid example in the random simulator.
```
thpani, line 199:
```suggestion
  /// valid example in the random simulator.
```

thpani (commented):
I'm a bit confused by the module/channel handling in the stateful part of the spec. There's a lot of inline `val`s defining module/channel mappings inside the pure defs. It seems like they should be lifted into a single, module-level `val` or `const`?

I stopped around line 170, because I couldn't figure out the chain/module/channel topology used inside the test.

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,308 @@
...
+    moduleState: ModuleState,
+    inFlightPackets: Set[Packet]
+  }
+
+  pure def sendFungibleTokens(moduleName: str, moduleState: ModuleState, packets: Set[Packet], // this could be an `sendPacket` operator when higher-order operators are supported
+                              denomination: DenomTrace, amount: UINT256,
+                              sender: Address, receiver: Address,
+                              sourcePort: str, sourceChannel: Channel,
+                              timeoutHeight: Height,
+                              timeoutTimestamp: uint64): SendFungibleTokensResult = {
```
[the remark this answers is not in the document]
thpani: Then `GlobalState` can eventually have a `Chain -> ModuleState` map of all the local chain contexts.

CHANGELOG.md
bugarela: Rebasing seems to have worked fine!

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,308 @@
...
+    moduleState: ModuleState,
+    acknowledgement: FungibleTokenPacketAcknowledgement
+  }
+
+  pure def onRecvPacket(moduleState: ModuleState, packet: Packet): ReceiveTokenPacketResult = {
+    val data = packet.data
+    if (movingBackAlongTrace({port: packet.sourcePort, channel: packet.sourceChannel}, data.denom)) {
+      // unescrow tokens to receiver
+      val escrowAccount = moduleState.channelEscrowAddresses.get(packet.destChannel)
+      val receiverDenom = updateTrace(data.denom, packet.sourcePort, packet.sourceChannel, packet.destPort, packet.destChannel)
```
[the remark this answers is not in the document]
bugarela: Yeah, this bothers me but I TBH also don't like either of your options.
Returning a pair from an `update` is weird, and might me complicated to
understand, and inlining the logic is something I really want to avoid as that
would be a duplication.

I thought of different option, but I'm not completely sure about it: We could
change `updateTrace` so it receives the boolean value for the result of
`movingBackAlongTrace` instead of calling it. Then we could also remove the
`sourcePort` and `sourceChannel` params of `updateTrace` since those are only
used to call `movingBackAlongTrace`. WDYT?

examples/cosmos/ics20/ics20.qnt
bugarela: I couldn't find anything saying we can't, and the `generateIdentifier` example
given in the [ICS 4
spec](https://github.com/cosmos/ibc/tree/main/spec/core/ics-004-channel-and-packet-semantics#opening-handshake)
makes me think that we could have a `channel-1` in each chain.

If your question was related to the fact that I was calling those "modules"
instead of "chains", than that's fixed.

thpani: > If your question was related to the fact that I was calling those "modules" instead of "chains", than that's fixed.

That's what threw me off, it's resolved :)

examples/cosmos/ics20/ics20.qnt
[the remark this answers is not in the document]
thpani: > We could change `updateTrace` so it receives the boolean value for the result of `movingBackAlongTrace` instead of calling it.

I like this option!

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,242 @@
...
+      val bankResult = chainState.bank.MintCoins(data.receiver, newDenom, data.amount)
+
+      val newChainState = chainState.with("bank", bankResult.accounts)
+      val ack = if (bankResult.success) {
+        { success: true, errorMessage: "" }
+      } else {
+        { success: false, errorMessage: "mint coins failed" }
+      }
+
+      { chainState: newChainState, acknowledgement: ack }
```
thpani: This logic is the same as in the `if`-branch, except for the `errorMessage` literal.

I'm wondering if we should fold de-duplicate this by moving this chunk outside the if-then-else. We can disambiguate the error message based on the result of `movingBackAlongTrace`.

thpani (approved):
Looks great!

I think also the test is really easy to follow now!

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,242 @@
...
+
+      chainState
+        .with("bank", bankResult.accounts)
+        .with("outPackets", chainState.outPackets.union(Set(packet)))
+    } else {
+      chainState
+    }
+  }
+
+  pure def onRecvPacket(chainState: ChainState, packet: Packet): ReceiveTokenPacketResult = {
```
josef-widder, line 96: I wonder whether instead of `ReceiveTokenPacketResult` returning a pair `(Chainstate, FungibleTokenPacketAcknowledgement)` wouldn't be more readable. It becomes more clear that the intention is to return and Ack and have a side effect. What do you think?

examples/cosmos/ics20/ics20.qnt
```diff
@@ -0,0 +1,242 @@
...
+  import base.* from "./base"
+  import ics20.*
+
+  /// Map from chain identifiers to their state
+  var chainStates: str -> ChainState
+
+  /// Map from chain identifiers to their connections, which are maps from
+  /// counterparty chain identifiers to channel identifiers. For example, chain
+  /// A connects to chain B through channel "channelToB".
+  pure val connections = Map(
```
josef-widder, line 141: I am a bit confused here. The IBC standards have the concept of connection (in the stack between clients and channels), but I guess this is not what is captured here? It seems that this is more abstract an captures the network of chains (the network "topology")? 
If this interpretation is right, I would rename `connections` to `network`?

examples/cosmos/ics20/ics20.qnt
bugarela: Makes sense! Will do

examples/cosmos/ics20/ics20.qnt
bugarela: Done!

PR #842, opened 2023-04-24
update the current spells

konnov:
Updating the cheatsheet to include the spells, so people would be able to discover them faster


thpani (approved):
LGTM

it would be great to have the cheatsheet in a more accessible/searchable format; any plans?
(@ivan-gavran also mentioned this as feedback some time ago)

konnov:
True. No idea of how to do that yet. I like that I can quickly fix it in keynote and produce a nice-looking result at the same time.

PR #844, opened 2023-04-25
fix the bug 843

### LINKED ISSUE #843: Nested `pure val` that refers to params caches wrong value
Consider the following example:
```
module test {
  def foo(x) = {
    pure val bar = x + 1
    bar
  }

  var t: int

  run aTest = (t' = foo(1)).then(t' = foo(2)).then(all{
    assert(t == 3),
    t' = t,
  })

}
```

The pure value `bar` here refers to `x`, which is parameter to `foo`, so the value for `bar` cannot be cached (since it changes if `x` changes). Currently, it is cached, resulting in this behavior:

```
  test
    1) aTest failed after 1 test(s)

  1 failed

  1) aTest:
      /home/gabriela/projects/quint/examples/test.qnt:10:5 - error: Assertion failed
      10:     assert(t == 3),

    [Frame 0]
    foo(1) => 2

    [Frame 1]
    foo(2) => 2

    Use --seed=0x84c63e40f06ed --match=aTest to repeat.

/home/gabriela/projects/quint/examples/test.qnt:10:5 - error: Assertion failed
10:     assert(t == 3),
        ^^^^^^^^^^^^^^

error: Tests failed
```

see how in `Frame 1`, `foo(2) => 2`, which should be `3`.

PS: I couldn't reproduce this by calling `foo()` in the REPL with different values. The bug appeared only in the simulator.

konnov:
Closes #843. Fixed the bug with spurious caching of nested `pure val`s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix `quint run` to output compile errors again (#812)
 - Fix the record constructor, so the key order does not matter (#839)
 - `quint test` shows compile errors, if they occur (#841)
+- Fix the bug that occurred when caching `pure val` (#844)
 
 ### Security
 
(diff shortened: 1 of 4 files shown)


bugarela (approved):
Thanks for fixing this so quickly!

PR #847, opened 2023-04-25
Update type maps at flattening with new generated ids

### LINKED ISSUE #846: Inferred types/effects maps are not updated with flatenning
Reported by Igor here https://github.com/informalsystems/quint/pull/845#issuecomment-1521530318

We should do something like this:
https://github.com/informalsystems/quint/blob/12bc9472150fa848a19885eeba88b40135b8a83e/quint/src/flattening.ts#L339-L343

For those maps

bugarela:
Hello :octocat: 

This closes #846 by adding a type map update every time we generate a new id for an expression, the same way we did for the source map. Effects are not handled yet since they don't even make it to this step of compilation (the map is not available, so there's nothing to update).


konnov (approved):
Looks good to me. Thanks for fixing this!

PR #852, opened 2023-04-26
Cleanup CLI

### LINKED ISSUE #850: Support `--with-lookup` option on `typecheck` command
Required for https://github.com/informalsystems/apalache/issues/2542#issuecomment-1521848863

Support for this flag was not added to the typecheck command when it was introduced in #639 

https://github.com/informalsystems/quint/blob/871f7a246905e0642c8b4bf1dbd443f574420f86/quint/src/cli.ts#L24-L55

shonfeder:
Closes #850

This closes #850 by removing needless filtration of our machine readable outputs
and the needless CLI flag introduced to counteract this filtration.

Any command that can set `--out` will now supply all useful data onto the given
JSON file.

Also does a light refactor on the CLI options.


konnov (approved):
Thanks for refactoring `--out`. I think we should somehow warn the users that the json output is not very much backwards compatible, as we are emitting all internals now, which may change often.

shonfeder:
Afaik, we are not exploding more internals than we had been via `typecheck —out`. So I’m not sure what additional breakage risk is introduced here that wasn’t before. Are you worried about some particular breakage case? In particular, we still filter which fields are output, we just don’t add a second level filter for particular sub commands. Do you have in mind a reason why we need that last part to protect stability?

bugarela (approved):

shonfeder:
I’ll follow up with Igor on possible concerns or warnings needed re: the data provided.

thanks for the reviews!

Issue #848, opened 2023-04-26
Remove `repeated` in favor of `reps`


konnov:
We should remove `repeated` for the reasons explained in #788. We will give the users a bit of time to transition to `reps` first.

PR #855, opened 2023-04-27
Introduce `reps` instead of `repeated`

### LINKED ISSUE #575: Operator for runs over indexes
Runs are great for guiding simulation without the necessity of defining new state variables just for that. Recently, I have tried to define runs over some data (i.e. a list containing which info to use at each step of the run), but I had to define at least one state variable for that: a counter

It would be great to either have a new operator or change `repeated` so it yields a counter:
```scala
((i) => MyAction(parameters[i])).repeated(10)
```

Here's my specific situation, in case it helps with context:

```scala
  run test =
    pure val errors: str -> List[bool] = Map(
      "A" -> [false, false, true, false, false, false, true, false, true, false, false],
      "B" -> [false, false, true, true, false, true, true, false, true, false, false]
    )
    pure val expectedOutput = [
      { terminal: "B", sent: 1, error: false, altr: true, altt: true },
      { terminal: "A", sent: 1, error: false, altr: false, altt: true },
      { terminal: "B", sent: 2, error: false, altr: false, altt: false },
      { terminal: "A", sent: 2, error: false, altr: true, altt: false },
      { terminal: "B", sent: 3, error: true, altr: false, altt: true },
      { terminal: "A", sent: 2, error: true, altr: true, altt: true },
      { terminal: "B", sent: 3, error: false, altr: true, altt: true },
      { terminal: "A", sent: 3, error: true, altr: true, altt: true },
      { terminal: "B", sent: 3, error: false, altr: true, altt: false },
      { terminal: "A", sent: 3, error: false, altr: false, altt: true },
      { terminal: "B", sent: 4, error: false, altr: false, altt: false },
      { terminal: "A", sent: 4, error: true, altr: false, altt: false },
      { terminal: "B", sent: 4, error: true, altr: false, altt: true },
      { terminal: "A", sent: 4, error: true, altr: false, altt: false },
      { terminal: "B", sent: 4, error: false, altr: false, altt: true },
      { terminal: "A", sent: 4, error: false, altr: true, altt: false },
      { terminal: "B", sent: 5, error: true, altr: false, altt: true },
      { terminal: "A", sent: 4, error: true, altr: true, altt: true },
      { terminal: "B", sent: 5, error: false, altr: true, altt: true },
      { terminal: "A", sent: 5, error: false, altr: false, altt: true },
      { terminal: "B", sent: 6, error: false, altr: false, altt: false },
      { terminal: "A", sent: 6, error: false, altr: true, altt: false },
    ]
    Init.then((all{
      SendMsg("B", errors.get("B")[counter]),
      output' = output,
      counter' = counter,
    }).then(all{
      ReceiveMsg("A", errors.get("A")[counter]),
      counter' = counter,
    }).then(all{
      SendMsg("A", errors.get("A")[counter]),
      output' = output,
      counter' = counter,
    }).then(all{
      ReceiveMsg("B", errors.get("B")[counter]),
      counter' = counter + 1,
    }).repeated(11)).then(all {
      assert(output == expectedOutput),
      output' = output,
      counter' = counter,
      altt' = altt,
      altr' = altr,
      storedMsgs' = storedMsgs,
      fileNumber' = fileNumber,
    })
```

### LINKED ISSUE #788: `repeated` should not be left-associative
As noticed by @lasarojc, the operator `repeated` binds its arguments opposite to what is intuitively expected from it:

```bluespec
>>> var y: int

>>> y' = 0
true
>>> (y' = y+1).repeated(10)
true
>>> y
10
>>> (y' = 0).then(y' = y + 1)
true
>>> y
1
>>> (y' = 0).then(y' = y + 1).repeated(10)
true
>>> y
1
>>> (y' = 0).then((y' = y + 1).repeated(10))
true
>>> y
10
```

In the third example, @lasarojc expected the output to be the same as in the fourth example. Indeed, we normally expect repetition (e.g., `*` in regular expressions) to be right-associative, whereas `.` is left-associative.

We should revisit the operator `repeated`, since it produces counterintuitive behavior.

In the first version of this operator, we called it `n.times(A)` instead of `A.repeated(n)`. Notice that the order of arguments was inverse to what we have in `repeated`, so left-associativity of `.` would not be a problem. We changed the operator, as `times` sounds like Cartesian product (this is the operator name in TLA+).

The candidate forms are:

 - `n.iterations(A)`
 - `n.repetitions(A)`.
 - `n.instants(A)`.

konnov:
Closes #788. Closes #575. Since we have found that `repeated` has usability issues, this PR introduces its left-associative version called `reps`. In addition to changing the order of arguments, `reps` also passes the iteration number to the action, as requested in #575.

Removal of `repeated` is planned in #848.


doc/lang.md
```diff
@@ -1781,8 +1785,36 @@ that start with `Init` and evaluate `Next` four times in a row.
 
 *Mode:* Run.
 
+#### Reps
+
+The expressions `n.reps(i => A(i))` or `n.reps(A)` repeats `n` times an action `A`.
+The iteration number, starting with 0, is passed as an argument of `A`.
+As actions are usually not parameterized by the iteration number,
+the most common form looks like: `n.reps(i => A)`.
```
bugarela, line 1793:
```suggestion
the most common form looks like: `n.reps(_ => A)`.
```

doc/lang.md
```diff
@@ -1781,8 +1785,36 @@ that start with `Init` and evaluate `Next` four times in a row.
...
+ - When `n <= 0`, this operator does not change the state.
+ - When `n = 1`, `n.reps(A)` is equivalent to `A(0)`.
+ - When `n > 1`, `n.reps(A)`, is equivalent to
+   `A(0).then((n - 1).reps(i => A(1 + i)))`.
+
+##### Example
+  
+```bluespec
+var x: int
+run test = (x' = 0).then(3.reps(i => x' = x + 1)).then(assert(x == 3))
```
bugarela, line 1806:
```suggestion
run test = (x' = 0).then(3.reps(_ => x' = x + 1)).then(assert(x == 3))
```

examples/language-features/counters.qnt
```diff
@@ -58,35 +58,31 @@ module counters {
...
   // execute Init and then Next four times
   run run3 = {
-    Init.then(Next.repeated(4))
+    Init.then(4.reps(_ => Next))
   }
 
   // execute Init and then Next ten times, while asserting n < 5
   run run4 = {
-    Init.then((all {
+    Init.then(10.reps(i => all {
```
bugarela, line 66:
```suggestion
    Init.then(10.reps(_ => all {
```

examples/language-features/counters.qnt
```diff
@@ -58,35 +58,31 @@ module counters {
...
       Next,
-    }).repeated(10))
+    }))
   }
 
   // this test should not fail
   run passingTest = {
-    Init.then(
-      all {
+    Init.then(10.reps(i => all {
```
bugarela, line 74:
```suggestion
    Init.then(10.reps(_ => all {
```

examples/language-features/counters.qnt
```diff
@@ -58,35 +58,31 @@ module counters {
...
-      }.repeated(10)
-    )
+    }))
   }
 
   // this test should fail
   run failingTest = {
-    Init.then(
-      all {
+    Init.then(10.reps(i => all {
```
bugarela, line 82:
```suggestion
    Init.then(10.reps(_ => all {
```

quint/test/runtime/compile.test.ts
```diff
@@ -1015,7 +1022,16 @@ describe('compiling specs to runtime values', () => {
...
         `)
 
-      assertVarAfterCall('n', '4', input)
+      assertVarAfterCall('n', '4', 'run1', input)
+    })
+
+    it('reps', () => {
+      const input = dedent(
+        `var n: int
+        |run run1 = (n' = 1).then(3.reps(i => n' = n + 1))
```
bugarela, line 1031:
```suggestion
        |run run1 = (n' = 1).then(3.reps(_ => n' = n + 1))
```

bugarela: Also: can you add a test where `i` is actually used?

bugarela (approved):
I'm very glad we will have the yielded indexes: I was actually missing those a bit in my ICS20 runs :smile: 

I'm not sure about the `reps` name. My experience with ruby makes me wish this was called `times` as we discussed initially, but I think `reps` is ok as well. I'd like to get @shonfeder's opinion on this before merging.

konnov:
> I'm not sure about the `reps` name. My experience with ruby makes me wish this was called `times` as we discussed initially, but I think `reps` is ok as well. I'd like to get @shonfeder's opinion on this before merging.

I also like `times` in Ruby, but it is too close to `\times` in TLA+, which has completely different meaning.

quint/test/runtime/compile.test.ts
```diff
@@ -67,19 +67,25 @@ function assertVarExists(kind: ComputableKind, name: string, input: string) {
...
     if (!key) {
-      return left('No callable found')
+      return left(`${callee} not found`)
     }
-
     const run = ctx.values.get(key)
     if (!run) {
-      return left(`${key} not found`)
+      return left(`${callee} not found via ${key}`)
     }
```
shonfeder, line 89: The control flow here is pretty convoluted, iiuc:

Whether or not we can do anything in this function seems to depend on `ctx.main`: but that's not apparent unless the reader first reasons out the mutable update to `key` in combination with the early return.

Using early exits consistently  for the error conditions would clean this up a lot IMO:

```typescript
    if (!ctx.main) {
      return left('main module not found')
    }

    const def = ctx.main.defs.find(def => def.kind === 'def' && def.name === callee)
    if (!def) {
      return left(`${callee} deff not found`)
    }

    const key = kindName('callable', def.id)
    if (!key) {
      return left(`${callee} not found`)
    }

    const run = ctx.values.get(key)
    if (!run) {
      return left(`${callee} not found via ${key}`)
    }
```

Alternatively, one of the points of the Either data type is to sequence error prone computations that need to return some data on error. Since this is already returning an `Either`, we could make use of that here with a helper like:

```typescript
function definedOrErr<T>(x: T | undefined, err: string): Either<string, NonNullable<T>> {
  return x == null ? left(err) : right(x);
}
```

To pipeline the error handling and make the data dependencies very clear:

```typescript
function evalVarAfterCall(
  varName: string,
  callee: string,
  input: string
): Either<string, string> {
  // use a combination of Maybe and Either.
  // Recall that left(...) is used for errors,
  // whereas right(...) is used for non-errors in sweet monads.
  const callback = (ctx: CompilationContext): Either<string, string> => {
    return definedOrErr(ctx.main, "main module not defined")
      .chain((main) =>
        definedOrErr(
          main.defs.find((def) => def.kind === "def" && def.name === callee),
          `def for ${callee} not found`
        )
      )
      .chain((def) =>
        definedOrErr(kindName("callable", def.id), `${callee} not found`)
      )
      .chain((key) =>
        definedOrErr(ctx.values.get(key), `${callee} not found via ${key}`)
      )
      .chain((run) =>
        run
          .eval()
          .map((res) => {
            if ((res as RuntimeValue).toBool() === true) {
              // extract the value of the state variable
              const nextVal = (
                ctx.values.get(kindName("nextvar", varName)) ?? fail
              ).eval();
              // using if-else, as map-or-unwrap confuses the compiler a lot
              if (nextVal.isNone()) {
                return left(`Value of the variable ${varName} is undefined`);
              } else {
                return right(
                  expressionToString(nextVal.value.toQuintEx(idGen))
                );
              }
            } else {
              const s = expressionToString(res.toQuintEx(idGen));
              const m = `Callable ${callee} was expected to evaluate to true, found: ${s}`;
              return left<string, string>(m);
            }
          })
          .or(just(left(`Value of ${callee} is undefined`)))
          .unwrap()
      );
  };

  return evalInContext(input, callback);
}
```

shonfeder (changes requested):

shonfeder:
I also agree that `reps` > `times` just by virtue of the possible TLA confusion. `reps` is also relevant for runs that **exercise** the spec :stuck_out_tongue_closed_eyes:

quint/test/runtime/compile.test.ts
konnov: Yep, added

doc/lang.md
```diff
@@ -8,77 +8,81 @@ This document presents language constructs in the same order as the [summary of
...
-      - [Fairness](#fairness)
-      - [Other temporal operators](#other-temporal-operators)
-    - [Unbounded quantifiers](#unbounded-quantifiers)
-  - [Instances](#instances)
-    - [Common case 1](#common-case-1)
-    - [Common case 2](#common-case-2)
-    - [The general case](#the-general-case)
-    - [No anonymous instances](#no-anonymous-instances)
-    - [Discussion](#discussion)
+<!-- Created by [gh-md-toc](https://github.com/ekalinin/github-markdown-toc.go) -->
```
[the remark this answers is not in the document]
konnov: That's true. I am not sure I know how to do one from the top of my head. I have opened an issue on that: https://github.com/informalsystems/quint/issues/857

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - add the precursor of the standard library in the form of spells (#827)
 - cache top-level `pure val` (#837)
 - support for higher-order operators in REPL/simulator (#845)
+- introduce `reps` as an alternative to `repeated` (#855)
```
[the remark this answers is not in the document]
konnov: Good point. I have added it under the "Deprecated" section.

quint/test/runtime/compile.test.ts
shonfeder: Yeah, the either chaining doesn’t add a lot of value here afaict :)

shonfeder (approved):

konnov:
> I also agree that `reps` > `times` just by virtue of the possible TLA confusion. `reps` is also relevant for runs that **exercise** the spec 😝

Yeah, `reps` come from gym exercises, so there is a precedence for this abbreviation :)

konnov:
@shonfeder, I think I have addressed your comments.

shonfeder:
Looks good to me, just looks like you need to run the generation command :)

quint/testFixture/_1028importCycleC.json
```diff
@@ -1 +1 @@
-{"stage":"parsing","warnings":[],"errors":[{"explanation":"Cyclic imports: 'mocked_path/testFixture/_1028importCycleC.qnt' imports 'mocked_path/testFixture/_1026importCycleA.qnt' imports 'mocked_path/testFixture/_1027importCycleB.qnt' imports 'mocked_path/testFixture/_1028importCycleC.qnt'","locs":[{"source":"mocked_path/testFixture/_1027importCycleB.qnt","start":{"line":1,"col":2,"index":20},"end":{"line":1,"col":45,"index":63}}]}]}
\ No newline at end of file
+{"stage":"parsing","warnings":[],"errors":[{"explanation":"extraneous input 'from' expecting {'}', 'const', 'var', 'assume', 'type', 'val', 'def', 'pure', 'action', 'run', 'temporal', 'import', DOCCOMMENT}","locs":[{"source":"mocked_path/testFixture/_1028importCycleC.qnt","start":{"line":1,"col":20,"index":38},"end":{"line":1,"col":23,"index":41}}]}]}
```
bugarela, line 1: @konnov seems like your `update-fixtures` command picked up an old version of quint without `from` and `export`

konnov: yeah, I have run `npm run generate` again, which uses the `npm` version

Issue #851, opened 2023-04-26
Constants that are operators are broken in the simulator


bugarela:
```
module A {
  const N: (int) => bool
  pure def N2 = (x) => x > 0
}


module test {
  import A(N = (x) => x > 0) as A1

  run myTest = all {
    assert(A1::N(1).fail()),
    assert(A1::N2(1)),
  }
}
```

This simple operator works as a pure def but not as a constant. This used to work, as I was using something like this in `denomTrace.qnt`, but not it is breaking (my bad for not having added it to the integration tests).

Might have something to do with caching @konnov

PR #856, opened 2023-04-27
Improve ICS20 denom trace interface

bugarela:
Hello :octocat: 

Closes https://github.com/informalsystems/apalache-team-services/issues/9. This improves `movingBackAlongTrace` and `updateTrace` interface so we don't have to call `movingBackAlongTrace` twice in `ics20.qnt`. 

I also had trouble getting the tests to work again because of #851. I've added a workaround for that.

And channel "setup" was renamed to "topology" as suggested by @josef-widder.


examples/cosmos/ics20/ics20.qnt
```diff
@@ -90,20 +90,22 @@ module ics20 {
   /// state and the acknowledgment
   pure def onRecvPacket(chainState: ChainState, packet: Packet): (ChainState, FungibleTokenPacketAcknowledgement) = {
     pure val data = packet.data
-    pure val result = if (movingBackAlongTrace({port: packet.sourcePort, channel: packet.sourceChannel}, data.denom)) {
+    pure val movingBack = movingBackAlongTrace(data.denom, { port: packet.sourcePort, channel: packet.sourceChannel })
+    pure val newDenom = updateTrace(movingBack, data.denom, { port: packet.destPort, channel: packet.destChannel })
+
+    pure val result = if (movingBack) {
```
josef-widder, line 96: Can we add a type annotation for `result`. I found this hard to read. Perhaps also add the suggested comment?
```suggestion
    // result is a pair of 1. a BankResult and 2. an error message that is used in case BankResult.success = false 
    pure val result = if (movingBack) {
```

josef-widder (commented):
Looks great.

Could we also add the [transfer test](https://github.com/cosmos/ibc-go/blob/a25f0d421c32b3a2b7e8168c9f030849797ff2e8/modules/apps/transfer/transfer_test.go#L44) from IBC-go? Perhaps in a parameterized way so that we can show that we can easily generate many more test cases?

examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -88,20 +86,27 @@ module denomTraceTest {
     path: [],
   }
 
-  /// Finds the correct channels for a transfer according to the chain setup
-  /// described above
-  pure def getChannelsForTransfer(sourceChain: str, destChain: str): { sourceChannel: str, destChannel: str } =
-    pure val channels = Map(
-      "A" -> "channelToA",
+  /// Map from chain identifiers to their stored channels, which are maps from
```
thpani, line 89: Not sure what "stored" means in this context, but I think the possessive pronoun is enough?

```suggestion
  /// Map from chain identifiers to their IBC channels, which are maps from
```

examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -24,17 +24,15 @@ module denomTrace {
   /// as the source zone. When the token is sent back to the chain it previously
   /// received from, the prefix is removed. This is a backwards movement in the
   /// token's timeline and the sender chain is acting as the sink zone."
-  pure def movingBackAlongTrace(hopSource: HopElement, denomTrace: DenomTrace): bool =
+  pure def movingBackAlongTrace(denomTrace: DenomTrace, hopSource: HopElement): bool =
```
thpani, line 27: Could we also mention in the comment above that the `HopElement` passed here describes the source end of the channel?

examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -88,20 +86,27 @@ module denomTraceTest {
...
-      // If this is a transfer from A -> B, then the source channel is `channelToB`
-      sourceChannel: channels.get(destChain),
-      // If this is a transfer from A -> B, then the dest channel is `channelToA`
-      destChannel: channels.get(sourceChain),
-    })
+  )
+
+  action Transfer(sourceChain, destChain) =
+    val movingBack = movingBackAlongTrace(trace, { port: "transfer", channel: channelTopology.get(sourceChain).get(destChain) })
+    trace' = updateTrace(movingBack, trace, { port: "transfer", channel: channelTopology.get(destChain).get(sourceChain) })
```
thpani, line 109: One thing I repeatedly find confusing is that the same IBC channel is sometimes referred to by its local channel name, and sometimes by its remote name.

Should we introduce a type alias
```bluespec
type Channel = { local: HopElement, remote: HopElement }
```
and a def that constructs this representation from a `ChannelTopology`?
We could then update `movingBackAlongTrace` and `updateTrace` to simply take a `Channel`. The record field names would also make it more obvious which channel end is referred to.

examples/cosmos/ics20/denomTrace.qnt
```diff
@@ -47,17 +45,17 @@ module denomTrace {
     }
   }
 
-  /// `isProperChannelSetup` is true iff channel names are unique by chain
-  pure def isProperChannelSetup(setup: str -> str -> str): bool =
-    pure val chainNames = Set("A", "B", "C") // should be setup.keys(), but that's breaking the effect checker somehow (see issue #808)
+  /// `isProperChannelTopology` is true iff channel names are unique by chain
+  pure def isProperChannelTopology(topology: ChannelTopology): bool =
```
thpani, line 49: If this is actually the only test, I'd prefer to name the predicate according to what it determines, e.g.,

```suggestion
  /// `channelNamesUniquePerChain` is true iff channel names are unique by chain
  pure def channelNamesUniquePerChain(topology: ChannelTopology): bool =
```

This way, I don't have to think what "proper" means when I see a call to `isProperChannelTopology`.

examples/cosmos/ics20/ics20.qnt
thpani: I agree, pairs are hard to read in general.
I think we should either use records over pairs, or immediately unpack the elements of a returned pair into their own named bindings.

thpani (approved):
Looks great! Nice refactoring towards an even clearer exposition!

I have two smallish ideas for refactoring in the comments below:

examples/cosmos/ics20/denomTrace.qnt
bugarela: I actually find it important to only handle source channel ends on `movingBackAlongTrace` and destination channel ends on `updateTrace` since that is what is happening in fact: the check on `movingBackAlongTrace` compares one side of the hop against the other side of the hop that was registered in the trace (by `updateTrace`). This was something that helped me see things clearer and I'd like to avoid generalizing this parameters (to receive more information than what's actually being used).

Also, what's local and what's remote is not really part of this spec scope IMO. Locality of channel information is not relevant here.

thpani: Maybe this is something we should follow-up on outside of this PR, to get it merged.

bugarela: But source and destination is an absolute truth, it is not relative to the perspective of each chain, as `local` and `remote` might be.

But I agree, we can revisit and iterate later :smile: will merge.

PR #858, opened 2023-04-28
Extend ICS-20 spec with failed acks

thpani:
Towards https://github.com/informalsystems/apalache-team-services/issues/8

Handle failed acknowledgements in the ICS-20 spec:
- Add `refundTokens`, `onAcknowledgePacket`, `onTimeoutPacket` definitions from ICS 20
- Update the `receivePacket` action to produce an acknowledgement
- Add a `receiveAck` action to receive an in-flight acknowledgement on the source chain
- Add a test provoking a failed acknowledgement

This PR is interspersed with various refactorings, but they are isolated to principled separate commits.

Since this has grown a lot, I will add the timeout action + test in a separate PR.

### DIFF
--- a/examples/cosmos/ics20/ics20.qnt
+++ b/examples/cosmos/ics20/ics20.qnt
@@ -192,7 +192,6 @@ module ics20Test {
   import base.* from "./base"
   import bank.getBalance from "./bank"
   import bank.getBalances from "./bank"
-  import basicSpells.has from "../../spells/basicSpells"
   import ics20.*
 
   /****************************************************************************
--- a/examples/cosmos/ics20/ics20.qnt
+++ b/examples/cosmos/ics20/ics20.qnt
@@ -235,12 +235,22 @@ module ics20Test {
     }).setToMap()
   })
 
+  /// Return true iff there is a channel from `sourceChain` to `destChain` defined in `CHANNEL_TOPOLOGY`.
+  ///
+  /// In particular, this returns false if `sourceChain == destChain`.
+  pure def hasChannelBetween(sourceChain: str, destChain: str): bool = all {
+    // ICS 4: "a channel is a pipeline [...] between specific modules on **separate** blockchains"
+    sourceChain != destChain,
+    // there is a channel between `sourceChain` and `destChain`
+    CHANNEL_TOPOLOGY.has(sourceChain),
+    CHANNEL_TOPOLOGY.get(sourceChain).has(destChain),
+  }
+
   /// Return true iff there is a channel from `sourceChain` to `destChain` defined in `CHANNEL_TOPOLOGY`
   /// that corresponds to the channel endpoints of `packet`.
   pure def isTravelingBetween(packet: Packet, sourceChain: str, destChain: str): bool = all {
     // there is a channel between `sourceChain` and `destChain`
-    CHANNEL_TOPOLOGY.has(sourceChain),
-    CHANNEL_TOPOLOGY.get(sourceChain).has(destChain),
+    hasChannelBetween(sourceChain, destChain),
     // packet endpoints correspond to channel ends
     channelCounterparties.get(sourceChain).has(packet.sourceChannel),
     channelCounterparties.get(sourceChain).get(packet.sourceChannel) == packet.destChannel
@@ -264,11 +274,8 @@ module ics20Test {
                     denom: DenomTrace, amount: UINT256,
                     sender: Address, receiver: Address): bool = all {
     // (1) Pre-condition:
-    // ICS 4: "a channel is a pipeline [...] between specific modules on **separate** blockchains"
-    sourceChain != destChain,
     // there is a channel between `sourceChain` and `destChain`
-    CHANNEL_TOPOLOGY.has(sourceChain),
-    CHANNEL_TOPOLOGY.get(sourceChain).has(destChain),
+    hasChannelBetween(sourceChain, destChain),
 
     // (2) Send the packet using `sendFungibleTokens`:
     val sourceChainState = chainStates.get(sourceChain)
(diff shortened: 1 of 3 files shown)


examples/cosmos/ics20/ics20.qnt
```diff
@@ -146,90 +219,243 @@ module ics20Test {
...
+  /// - Call the `onRecvPacket` callback on it.
+  /// - Update the chain states to record the packet as received.
+  /// - Record (but not receive) the acknowledgement produced by `onRecvPacket`
+  ///   on `destChain`.
+  action receivePacket(sourceChain: str, destChain: str): bool = all {
+    // (1) Pre-condition:
+    // ICS 4: "a channel is a pipeline [...] between specific modules on **separate** blockchains"
+    sourceChain != destChain,
+
+    // (2) Non-deterministically pick a packet from `sourceChain` to receive on `destChain`:
```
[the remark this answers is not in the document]
thpani: Good catch! I believe I fixed it in the commits below, PTAL

examples/cosmos/ics20/ics20.qnt
```diff
@@ -146,90 +219,266 @@ module ics20Test {
...
+  /// of `amount`-many tokens of `denom` from `sender` to `receiver`.
+  action sendPacket(sourceChain: str, destChain: str,
+                    denom: DenomTrace, amount: UINT256,
+                    sender: Address, receiver: Address): bool = all {
+    // (1) Pre-condition:
+    // ICS 4: "a channel is a pipeline [...] between specific modules on **separate** blockchains"
+    sourceChain != destChain,
+    // there is a channel between `sourceChain` and `destChain`
+    CHANNEL_TOPOLOGY.keys().contains(sourceChain),
+    CHANNEL_TOPOLOGY.get(sourceChain).keys().contains(destChain),
```
bugarela, line 270: I really like `isTravelingBetween`. Can we also have a `hasChannelBetween` and call
`CHANNEL_TOPOLOGY.hasChannelBetween(sourceChain, destChain)`?

examples/cosmos/ics20/ics20.qnt
```diff
@@ -146,90 +219,266 @@ module ics20Test {
...
     all {
-      chainStates' = chainStates
-        .set(destChain, result._1)
-        .set(sourceChain, sourceChainState.with("outPackets", sourceChainState.outPackets.exclude(Set(packet))))
+      // (1) Pre-condition:
+      // ICS 4: "a channel is a pipeline [...] between specific modules on **separate** blockchains"
+      sourceChain != destChain,
+      // there is a channel between `sourceChain` and `destChain`
+      CHANNEL_TOPOLOGY.keys().contains(sourceChain),
+      CHANNEL_TOPOLOGY.get(sourceChain).keys().contains(destChain),
```
bugarela, line 298: This checks are already done by `isTravelingBetween` in line 300. Can we remove the redundancy or do you think it is important somehow?
```suggestion
```

examples/cosmos/ics20/ics20.qnt
```diff
@@ -146,90 +219,266 @@ module ics20Test {
...
+      // there is an unreceived packet traveling from `sourceChain` to `destChain`
+      sourceChainState.outPackets.exists(packet => packet.isTravelingBetween(sourceChain, destChain)),
+
+      // (2) Non-deterministically pick a packet from `sourceChain` to receive on `destChain`:
+      nondet packet = sourceChainState.outPackets.oneOf()
```
bugarela, line 303: Just a note, you might want to change this or not: It is more efficient for the random simulator if you filter the set before calling `oneOf` (as oppose to having the `isTravelingBetween` condition being enforced on the picked value). What will happen here is that, if there are too many packets in `outPackets`, the simulator will try to pick values a few times and then fail the action if the randomly picked values are not the correct ones to satisfy the condition. Filtering makes it that the the simulator always gets this right.

This problem might not be relevant for the simulations we are currently doing, but I had issues with this in the `denomTrace` spec and decided do make sure I'm always filtering my sets before calling `oneOf` on them.

bugarela (approved):
LGTM! Just a few optional suggestions

examples/cosmos/ics20/ics20.qnt
```diff
@@ -112,26 +128,83 @@ module ics20 {
...
+      } else {
+        // The (failed) transfer attempt was moving the token further along its denom trace.
+        // For this, we originally escrowed tokens in `sendFungibleTokens`, so unescrow them back.
+        pure val escrowAccount = chainState.channelEscrowAddresses.get(packet.srcChannel)
+        chainState.bank.TransferCoins(escrowAccount, data.sender, data.denom, data.amount)
+      }
+    if (bankResult.success) {
+      chainState.with("bank", bankResult.accounts)
+    } else {
+      // TODO: ICS 20 does not specify what happens if the bank modules return an error.
```
josef-widder, line 167: We shouldn't forget about that. Shall we open an issue on the IBC spec repo?

josef-widder (commented):
Looks great!

I find it amazing how readable the onTimeout and onAck functions are.

josef-widder (commented):
Looks great!

I find it amazing how readable the onTimeout and onAck functions are.

examples/cosmos/ics20/ics20.qnt
thpani: Will do! I've filed https://github.com/informalsystems/apalache-team-services/issues/14 to follow up on.

examples/cosmos/ics20/ics20.qnt
```diff
@@ -146,90 +219,266 @@ module ics20Test {
...
+  /// Return true iff there is a channel from `sourceChain` to `destChain` defined in `CHANNEL_TOPOLOGY`
+  /// that corresponds to the channel endpoints of `packet`.
+  pure def isTravelingBetween(packet: Packet, sourceChain: str, destChain: str): bool = all {
+    // there is a channel between `sourceChain` and `destChain`
+    CHANNEL_TOPOLOGY.keys().contains(sourceChain),
+    CHANNEL_TOPOLOGY.get(sourceChain).keys().contains(destChain),
+    // packet endpoints correspond to channel ends
+    channelCounterparties.get(sourceChain).keys().contains(packet.sourceChannel),
+    channelCounterparties.get(sourceChain).get(packet.sourceChannel) == packet.destChannel
+  }
```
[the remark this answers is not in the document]
thpani: Oh that's great! I like how it aligns the code 🤩 [9a162b2](https://github.com/informalsystems/quint/pull/858/commits/9a162b2dc8c2934ac2770a51ce40a68d5ab3ee37)

examples/cosmos/ics20/ics20.qnt
bugarela: I feel like it belongs on the `isTravelingBetween` as a sanity check, not here. My interpretation is that if there is a packet traveling between A and B, then surely there is a connection between A and B. This check seems more like a sanity check on the compatibility of the packet with the topology, and since `isTravelingBetween` is the operator acting on the packet here, seems like sanity checks should be done there.

But this is not very relevant to me, feel free to keep the check here if you think its better.

examples/cosmos/ics20/ics20.qnt
konnov: Right. This may show some difference in the simulator, as it would reject some traces, if the condition is checked after `oneOf`. If you filter a set upfront, before calling `oneOf`, then the simulator would not be able to make an infeasible choice. This would not matter for Apalache though.

examples/cosmos/ics20/ics20.qnt
thpani: Then let's keep the redundance, since we both believe it's relevant at a different point :)

examples/cosmos/ics20/ics20.qnt
thpani: Aaah yes, this also looks nicer! 0dba89c
I actually tried to write a predicated `oneOf` before writing this. Minimality of the core language aside, would it make sense to have `oneOf` optionally take a filter?

PR #859, opened 2023-05-03
Language server performance boost & support for multiple files

bugarela:
Hello :octocat: 

This is a huge update on the language server. For the users, the main impacts should be:
- support for multiple files in the existing features (signature help, go to definition, overlay documentation)
- less lag

The changes for supporting multiple files were done in commit https://github.com/informalsystems/quint/commit/a8c874e015272fcc0a63a0ca85ac73340f5a3d57. I was not happy with performance while testing this commit, so I made other changes pursuing that:
1. The server is now a class instead of a bunch of global constants. This makes it easier to track state changes (as we can see what begins with `this.`) and prevents issues with `async` functions.
2. Instead of running `QuintAnalyzer` for every character the user types, we only run it after 1 second without any changes, so we don't pile up a bunch of runs and cause lag. Parsing is still done right away since it is fast enough and we depend too much on the AST for our functionality.
3. From (1), `server.ts` was already with a huge diff and looked too messy, so I also extracted some pure fragments of it to different files.


quint/src/docs.ts
```diff
@@ -50,6 +50,29 @@ export function produceDocs(quintModule: QuintModule): Map<string, Documentation
   return new Map<string, DocumentationEntry>(compact(entries))
 }
 
+/**
+ * Produces a documentation map for the modules' definitions
+ *
+ * @param quintModule the module for which documentation should be produced
+ * @returns a map of definition ids to their documentation
+ */
+export function produceDocsById(quintModule: QuintModule): Map<bigint, DocumentationEntry> {
```
bugarela, line 59: This is a copy from the function above, but with the result indexed by id instead of name. We should probably stick with only this one in the future.

vscode/quint-vscode/server/src/hover.ts
bugarela, line 1: Unchanged behavior, only extracted

vscode/quint-vscode/server/src/parsing.ts
bugarela, line 1: Unchanged, only extracted

vscode/quint-vscode/server/src/server.ts
```diff
@@ -26,319 +26,260 @@ import {
...
+    // when the text document first opened or when its content has changed.
+    documents.onDidChangeContent(async(change) => {
+      parseDocument(this.idGenerator, change.document).then((result) => {
+        this.parsedDataByDocument.set(change.document.uri, result)
+      }).catch(diagnostics => {
+        this.connection.sendDiagnostics({ uri: change.document.uri, diagnostics })
+      })
 
-    const [loc, type] = typeResult
+      this.scheduleAnalysis(change.document)
```
bugarela, line 95: This is a relevant peace: schedule analysis instead of doing it right away.

vscode/quint-vscode/server/src/server.ts
```diff
@@ -26,319 +26,260 @@ import {
...
+      const start = text.search(new RegExp(unqualifiedName))
+      return [{
+        // The range for the name being hover over
+        originSelectionRange: locToRange(parsedData.sourceMap.get(nameId)!),
+        targetUri: uri,
+        // The range for the entire definition (including the qualifier and body)
+        targetRange: range,
+        // The range for the definition's name
+        targetSelectionRange: { ...range, start: { ...range.start, character: range.start.character + start } },
+      }]
```
bugarela, line 184: We now return this `DefinitionLink` instead of a simple location, which enables the client to open or navigate to a different file if the definition is there.

vscode/quint-vscode/server/src/definitions.ts
```diff
@@ -0,0 +1,36 @@
...
+  nameId: bigint,
+  name: string,
+  definitionId?: bigint,
+}
+
+export function findDefinition(
+  parsedData: ParserPhase3, uri: string, position: Position
+): QuintDefinitionLink | undefined {
+  const { modules, sourceMap, table } = parsedData
+  const results: [Loc, bigint][] = [...sourceMap.entries()].map(([id, loc]) => [loc, id])
```
konnov, line 16: Would the following piece copy the whole array (presumably, a large one), even though we need just a few entries?

```ts
[...sourceMap.entries()]
```

konnov (approved):
Looks great!

vscode/quint-vscode/server/src/definitions.ts
bugarela: Yeah, I think I tried to generalize things to early here, we shouldn't need to convert those maps into lists of tuples at all. Changing this will require some meaningful refactor tho, as this `results` format is being used across different functions. I'll add an issue for this refactor.

PR #845, opened 2023-04-25
support higher-order operators in the simulator/REPL

konnov:
Closes #221. This PR adds support for higher-order operators in the simulator/REPL. This is the third attempt at implementing it. This time it worked.

This feature required more changes to the representation of runtime values and `Computable`s than I expected. Basically, a lambda compiles to a runtime value that can be evaluated later again, when provided arguments.

I still have to stare at the diff to see, whether the code could be improved.

- [x] Tests added for any new code
- [ ] Entries added to the respective `CHANGELOG.md` for any new functionality
- [ ] Feature table on [`README.md`](../README.md#roadmap) updated for any listed functionality

PR #841, opened 2023-04-20
handle compilation errors that occur in `quint test`

konnov:
Closes #828. This PR fixes the problem of compilation errors not being shown by `quint test`.

- [x] Tests added for any new code
- [x] Entries added to the respective `CHANGELOG.md` for any new functionality

PR #861, opened 2023-05-04
Release v0.10.0

bugarela:
### Added

- Modules and definitions can now be exported (#771)
- Non-deterministic tests are run multiple times similar to {Quick,Scala}check (#786)
- `quint run` returns a non-zero exit code on invariant violation and errors (#793)
- Support for imports from files: `import ... from <filename>` (#800)
- suppress comments in REPL (#806)
- `quint repl` is printing the same version number as returned by `quint --version` (#804)
- add the command `.seed` in REPL (#812)
- fix `quint run` to output compile errors again (#812)
- add the precursor of the standard library in the form of spells (#827)
- cache top-level `pure val` (#837)
- support for higher-order operators in REPL/simulator (#845)
- introduce `reps` as an alternative to `repeated` (#855)
- More data is now output in most commands when `--out` is provided (#852)
- Mock command `verify` for running Apalache in the future (#787)

### Changed

- The ITF export adds output similar to Apalache (#780)

### Deprecated

- `repeated` is deprecated in favor of `reps` (#855)

### Removed

- `notin` is no longer a builtin operator (#814)
- `--with-lookup` command (#850)

### Fixed

- Proper errors are now reported when a lambda returns an operator (#811)
- Fix `quint run` to output compile errors again (#812)
- Fix the record constructor, so the key order does not matter (#839)
- `quint test` shows compile errors, if they occur (#841)
- Fix the bug that occurred when caching `pure val` (#844)


bugarela:
@shonfeder I had a [problem](https://github.com/informalsystems/quint/actions/runs/4880901338/jobs/8709145136?pr=861) because `version.ts` was not updated, since we don't run `npm run compile` in the preparation script. I forced pushed the version change on the same release commit, so I think this won't be a problem for this release. And then we have to update the preparation script for next time. Seems good?

shonfeder (approved):
Makes sense to me!

PR #865, opened 2023-05-05
Add timeout handling to ICS-20 spec

thpani:
Closes https://github.com/informalsystems/apalache-team-services/issues/8

ℹ️ based on #858 – review that one first 😉 

Add 2 actions, one for timing out packets, and one for processing the timeout on the originating chain.
Add a test for proper timeout handling.

### DIFF
--- a/examples/cosmos/ics20/ics20.qnt
+++ b/examples/cosmos/ics20/ics20.qnt
@@ -1,7 +1,7 @@
 // -*- mode: Bluespec; -*-
 
 /**
- * A specification for the ICS20 fungible token transfer protocol:
+ * A specification of the ICS20 fungible token transfer protocol:
  * https://github.com/cosmos/ibc/tree/main/spec/app/ics-020-fungible-token-transfer
  *
  * Gabriela Moreira and Thomas Pani, Informal Systems, 2023
@@ -45,10 +45,12 @@ module ics20 {
   type ChainState = {
     bank: Accounts,
     channels: ChannelCounterparties,
-    outPackets: Set[Packet],
-    receivedButUnacknowledgedPackets: Set[Packet],
-    inAcknowledgements: Set[FungibleTokenPacketAcknowledgement],
     channelEscrowAddresses: Channel -> Address,
+    // IBC packet handling:
+    outPackets: Set[Packet],                        // outgoing unreceived packets
+    receivedButUnacknowledgedPackets: Set[Packet],  // received but unacknowledged packets
+    inAcknowledgements: Set[FungibleTokenPacketAcknowledgement],  // incoming acknowledgements
+    inTimeouts: Set[Packet],                                      // incoming timeouts
   }
 
   /****************************************************************************
@@ -158,7 +160,7 @@ module ics20 {
       } else {
         // The (failed) transfer attempt was moving the token further along its denom trace.
         // For this, we originally escrowed tokens in `sendFungibleTokens`, so unescrow them back.
-        pure val escrowAccount = chainState.channelEscrowAddresses.get(packet.srcChannel)
+        pure val escrowAccount = chainState.channelEscrowAddresses.get(packet.sourceChannel)
         chainState.bank.TransferCoins(escrowAccount, data.sender, data.denom, data.amount)
       }
     if (bankResult.success) {
@@ -274,9 +276,11 @@ module ics20Test {
       bank: if (chain == "A") Map("alice" -> Map(ATOM -> 100)) else Map(),
       channels: channelCounterparties.get(chain),
       channelEscrowAddresses: channelCounterparties.get(chain).keys().mapBy(_ => ESCROW_ACCOUNT),
+      // Start without any unprocessed packets/acknowledgements/timeouts
       outPackets: Set(),
       receivedButUnacknowledgedPackets: Set(),
-      inAcknowledgements: Set()
+      inAcknowledgements: Set(),
+      inTimeouts: Set()
     })
   }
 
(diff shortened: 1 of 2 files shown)


bugarela (approved):
:sparkles:

examples/cosmos/ics20/ics20.qnt
```diff
@@ -45,10 +45,12 @@ module ics20 {
   type ChainState = {
     bank: Accounts,
     channels: ChannelCounterparties,
-    outPackets: Set[Packet],
-    receivedButUnacknowledgedPackets: Set[Packet],
-    inAcknowledgements: Set[FungibleTokenPacketAcknowledgement],
     channelEscrowAddresses: Channel -> Address,
+    // IBC packet handling:
+    outPackets: Set[Packet],                        // outgoing unreceived packets
```
konnov, line 50: what does it mean to be outgoing and unreceived? I guess, a sender chain would not immediately learn that its packet has been received, until the moment it receives an acknowledgment. Would something like `sentNoAckPackets` describe this set better?

examples/cosmos/ics20/ics20.qnt
```diff
@@ -45,10 +45,12 @@ module ics20 {
...
+    receivedButUnacknowledgedPackets: Set[Packet],  // received but unacknowledged packets
```
konnov, line 51: how about `rcvdToAckPackets`? I am just worried about very long names, which are hard to type.

Also, is this set on the sender side or the receiver side? Kind of a silly question, but it would be nice to understand.

examples/cosmos/ics20/ics20.qnt
```diff
@@ -306,7 +308,30 @@ module ics20Test {
                       .set(sourceChain, newSourceChainState)
   }
 
-  /// Receive an acknowledgement for `packet` on `chain`:
+  /// Receive a packet sent from `sourceChain` to `destChain` and consider it timed out.
+  action timeoutPacket(sourceChain: str, destChain: str): bool = all {
```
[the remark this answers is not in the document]
konnov: Is not `destChain` used in the condition on line 315?

examples/cosmos/ics20/ics20.qnt
```diff
@@ -325,6 +350,25 @@ module ics20Test {
...
 
+  /// Receive a timeout on `chain`:
+  /// - Pick an arbitrary, previously unreceived timeout on `chain`.
+  /// - Call the `onTimeoutPacket` callback on it.
+  /// - Update the chain state to record the timeout as received.
+  action receiveTimeout(chain: str): bool = all {
+    // (1) Compute updated chain state:
+    // non-deterministically pick an acknowledgement
+    val chainState = chainStates.get(chain)
+    nondet timedoutPacket = chainState.inTimeouts.oneOf()
```
konnov, line 361: `oneOf` requires the set to be non-empty. So you have to add a test for it before calling `oneOf`. It's a bit inconvenient.

examples/cosmos/ics20/ics20.qnt
```diff
@@ -325,6 +350,25 @@ module ics20Test {
...
+    // call the `onTimeout` callback and
+    val timedoutChainState = onTimeoutPacket(chainState, timedoutPacket)
+    // remove `timedoutPacket` from incoming timeouts
+    val newInTimeouts = chainState.inTimeouts.exclude(Set(timedoutPacket))
+    val newChainState = timedoutChainState.with("inTimeouts", newInTimeouts)
```
konnov, line 366: Is it possible in IBC to both timeout a packet and acknowledge it? I do not see why this should not be possible.

konnov (approved):
Looks good. I've left a few questions

examples/cosmos/ics20/ics20.qnt
josef-widder: We discussed synchronously. Igor sees now ;-)

examples/cosmos/ics20/ics20.qnt
thpani: Yes, we have to do this everywhere. I wanted to introduce it separately, together with a `step` action. Currently it does not matter much, because when writing tests, we know when the action can succeed (i.e., when the underlying set is non-empty).

examples/cosmos/ics20/ics20.qnt
thpani: > Would something like `sentNoAckPackets` describe this set better?

I don't see how this helps, it's really receival that's decisive here, because producing and consuming the ack is asynchronous.

examples/cosmos/ics20/ics20.qnt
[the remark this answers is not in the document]
bugarela: Yes, but what I mean is: currently this action describes a "timeout any packet coming from source chain" (and not "timeout any packet coming from source chain and going to dest chain"), so it seems misleading that we receive `destChain` as a parameter.

IMO, the check on line 315 only makes sense if we are actually using `destChain` for something else.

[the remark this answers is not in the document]
thpani: Gabriela's right, this should only be predicated on the source chain. bacf020

examples/cosmos/ics20/ics20.qnt
thpani: We could probably separate this out into a separate `RelayerState`. However, it's not obvious where to draw the line between on-chain state and relayer state, where they overlap.

examples/cosmos/ics20/ics20.qnt
thpani: See https://github.com/informalsystems/quint/pull/865#discussion_r1187325862. The distinguishing feature is receival, not acknowledgement.

> I am just worried about very long names, which are hard to type.
Why? 🙃

examples/cosmos/ics20/ics20.qnt
thpani: Fixed in 4530c46

examples/cosmos/ics20/ics20.qnt
thpani: We have a separate issue for addressing this, so we can follow up there: https://github.com/informalsystems/apalache-team-services/issues/7

PR #866, opened 2023-05-05
User stories for interaction with traces

konnov:
We have a stream of issues that are all related to the theme of interacting with the current state and traces. This PR accumulates 3 user stories and a preliminary REPL UX, in order to address these issues.

None of these stories are implemented. I did it backwards, in order to have a coherent approach to this feature.


doc/story008-repl-traces.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+**Proposed interface**. The simplest yet powerful interface is to introduce the
+system state as a record that is accessible via a standard name. We decided to
+use the name `q::state`. Here is an example:
+
+```bluespec
+>>> q::state
+{
+  beeping: false,
```
thpani, line 78: This is the first time I see something living in the `q::` toplevel.

Wouldn't it be more consistent to use a REPL command `.state` for this?
That has the added benefit that I can "discover" it through `.help`.

doc/story008-repl-traces.md
```diff
@@ -0,0 +1,262 @@
...
+```bluespec
+>>> q::state.fieldNames()
+Set("beeping", "heatingOn", "temperature")
+
+>>> q::state.temperature
+20
+```
+
+**Shortcut.** Since typing `q::state` often may be tedious, we offer a shortcut
+*`q::s` for `q::state`.
```
thpani, line 95: Please don't.

doc/story008-repl-traces.md
```diff
@@ -0,0 +1,262 @@
...
+true
+>>> depressButton
+true
+```
+
+**Current interface**. Assuming that `q::trace` from Story 2 is implemented,
+we should be able to see the intermediate states that are produced by the above
+sequence of actions. However, if we want to evaluate state invariants against
+the intermediate states, we have to replay the actions from start, e.g., by
+executing `init`, then `pressButton`, then `heat`.
```
thpani, line 182: Why is this solved by jumping around the trace, rather than giving a better interface to how we handle checking of invariants?

doc/story008-repl-traces.md
```diff
@@ -0,0 +1,262 @@
...
+
+**Proposed interface**. We introduce several actions for navigating the trace.
+
+### 3.1. Jump
```
thpani, line 186: I'd advocate for holding off on implementing this – it kind of introduces REPL state on top of a stateful spec. It's hard to make this explicit in terms of output, so I'd expect it to cause more confusion than help.

I'd rather have people be explicit about where they are in a trace.

What if a newcomer enters `q::jump(2)` without knowing what it means? They will have a horrible time figuring out why they're not getting the expected output on their subsequent commands.

thpani (commented):
1. I like `q::state` and `q::trace`, though I'm torn about the naming.
It feels weird that these occupy some magic, top-level like namespace.
More detailed comments below.

2. I'm not convinced by the more complex functionality described here.
It feels like you're trying to implement a GUI (where one can naturally navigate and interact with a trace) on the CLI, and I'm concerned about the (unintended) complexity that could be introduced through this.

doc/story008-repl-traces.md
lasarojc: +1 for .state and not risking colliding with any user defined namespaces.

+1 for it being a regular record.

how about an intermediate approach, using an empty namespace, that is, `::state`?

What if I have multiple modules imported? Will `q::state` be a record of records?

If `::state` is used for the top-level namespace, then it becomes almost intuitive to use `M::state` for a module imported as `import bla as M from file`

doc/story008-repl-traces.md
```diff
@@ -0,0 +1,262 @@
...
+
+The following REPL session highlights the expected interface:
+
+```bluespec
+>>> init.then(pressButton).then(heat).then(depressButton)
+true
+>>> q::jump(2)
```
lasarojc, line 193: While I agree with @thpani that these features may be confusing for newcomers, I would love these features. I'd volte for implementing and clearly document it rather than not.

On the interest of clearly documenting, I would do `::jump(1)` here so that when the history is modified below the history is shorter, showing very clearly that something was thrown away.

konnov:
Since this PR has been around for long time, I will just merge it. There are useful comments, which we might want to refer to in the future.

PR #874, opened 2023-05-08
ICS20 test based on `transfer_test.go`

bugarela:
Hello :octocat: 

Closes https://github.com/informalsystems/apalache-team-services/issues/11 by adding a concrete test based on the test in the go implementation. Does not add a parametrization on top of this yet.

The comments on the go test are outdated (seems like they updated the test and forgot about the comments). I'll open a PR fixing that. For now, if you are going to review this while looking at the code, I suggest ignoring their comments since that can be very confusing.

Perhaps is a good idea o hold on reviewing this until I have the comments fixed there.

**UPDATE**: here's the test with updated comments and other improvements
https://github.com/cosmos/ibc-go/blob/ce6ee2661f068a36fc112b29aa81b6a9c79685e8/modules/apps/transfer/transfer_test.go#L46

### DIFF
--- a/examples/cosmos/ics20/ics20.qnt
+++ b/examples/cosmos/ics20/ics20.qnt
@@ -257,7 +257,7 @@ module ics20Test {
 
   /* Helper operators to manipulate bank state */
   /// Sets the balance of `account` in `chain` to `amount` of `denom`.
-  action SetBalanceIn(chain, account, denom, amount) = {
+  action setBalanceIn(chain, account, denom, amount) = {
     chainStates' = chainStates.setBy(chain, state =>
       state.with("bank", state.bank.setBy(account, balances => balances.set(denom, amount)))
     )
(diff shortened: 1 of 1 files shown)


examples/cosmos/ics20/ics20.qnt
```diff
@@ -480,5 +480,59 @@ module ics20Test {
...
+    pure val amount = 9223372036854775808
+
+    // NOTE: fungible token is prefixed with the full trace in order to verify the packet commitment
+    pure val denomInA = ATOM
+    pure val denomSentFromAToB = denomInA.with("path", [{ port: "transfer", channel: "channelToA" }].concat(denomInA.path))
+    pure val denomSentFromBToC = denomSentFromAToB.with("path", [{ port: "transfer", channel: "channelToB" }].concat(denomSentFromAToB.path))
+    init.then(
+      // Make sure that Alice has enough tokens on chain A. I did not understand
+      // how this is done in the go test.
+      chainStates' = chainStates.setBy("A", state => state.with("bank", state.bank.setBy("alice", balances => balances.set(ATOM, amount))))
```
konnov, line 495: `chainStates.setBy("A", state => state.with("bank", state.bank.setBy("alice", balances => balances.set(ATOM, amount))))`

I feel like this expression is asking for a helper definition that make this much shorter and understandable, similar to TLA+'s `EXCEPT` actually:

```
pure val setBalanceIn(chain, account, denom, amount) = {
  chainStates.setBy(chain, state =>
    state.with("bank", state.bank.setBy(account, balances => balances.set(denom, amount)))
  )
}
```

It's actually interesting to see how these calls compose. The functional style seems right to me.

examples/cosmos/ics20/ics20.qnt
```diff
@@ -480,5 +480,59 @@ module ics20Test {
...
+    ).then(
+      sendTransfer("A", "B", ATOM, amount, "alice", "bob")
+    ).then(all {
+      // check that voucher exists on chain B
+      val balanceB = chainStates.get("B").bank.getBalances("bob").getBalance(denomSentFromAToB)
```
konnov, line 500:
```
chainStates.get("B").bank.getBalances("bob").getBalance(denomSentFromAToB)
```

Similar to the above comment,  this is asking for a helper definition:

```
pure def getBalanceIn(chain, account, denom) = {
  chainStates.get(chain).bank.getBalances(account).getBalance(denom)
}
```

konnov (approved):
Looks great. I suggest adding two helper definitions. Our goal is to improve readability. I think the goals of the Golang engineers are a bit different. That's why we see a lot of verbose code in their tests.

josef-widder (commented):
Looks great. I wonder what parameterization we should put on-top?

- obviously the amount can be a parameter
- similarly the chains
- We could have several repetitions of this scenario with non-deterministically chosen amounts (or with the same amount) with parameterized assertions 

Not sure what would be the best to demonstrate the power of parameterization of tests. @konnov, what do you think?

thpani:
@bugarela Could you please rebase? I just merged the parent PR, and it messed up the diff here.

bugarela:
> @bugarela Could you please rebase? I just merged the parent PR, and it messed up the diff here.

Done!

examples/cosmos/ics20/ics20.qnt
```diff
@@ -255,6 +255,19 @@ module ics20Test {
     channelCounterparties.get(sourceChain).get(packet.sourceChannel) == packet.destChannel
   }
 
+  /* Helper operators to manipulate bank state */
+  /// Sets the balance of `account` in `chain` to `amount` of `denom`.
+  action SetBalanceIn(chain, account, denom, amount) = {
```
thpani, line 260: so far, all our actions are lower camel case, this is upper camel case.
shall we use the same spelling on all of them?

thpani (approved):
Looks great!

Thematically, I think this should come immediately before or after `ABCACBATest`, because both follow a sequence of successful transfers. Should we move this one up?

examples/cosmos/ics20/ics20.qnt
bugarela: For some reason, I thought we were using upper camel case for actions, but we are not. Will fix.

bugarela:
> Looks great. I wonder what parameterization we should put on-top?
> 
> * obviously the amount can be a parameter
> * similarly the chains
> * We could have several repetitions of this scenario with non-deterministically chosen amounts (or with the same amount) with parameterized assertions
> 
> Not sure what would be the best to demonstrate the power of parameterization of tests. @konnov, what do you think?

I've created an issue to follow up on this https://github.com/informalsystems/apalache-team-services/issues/16

PR #880, opened 2023-05-09
Introduce spread syntax for record updates

### LINKED ISSUE #816: Improve syntax for record updates and access
I think we should remove the `field` operator, because, as @lasarojc noted, it's type signature is a lie.

```
>>> val r = {a: 2}

>>> val f = "a"

>>> r.field(f)
static analysis error: error: [QNT000] Record field name must be a string expression but is name: f
Generating record constraints for r,f
Generating constraints for f

r.field(f)
^^^^^^^^^^


2
```

either `f` doesn't have type `string`, or the type signature for `field` is incorrect.

For the same reason, we should replace `with` with an update syntax that doesn't pretend to take strings :)

E.g., `{r with a : 42}`.

We could introduce the ability to pass around labels, perhaps indexed by union types, around. But this would be significantly more involved.

konnov:
Closes #816. This PR simply introduces the syntax form `{ f1: e1, fN: eN, ...record }`, which is translated to `record.with("f1", e1).with("fN", eN)`.


doc/lang.md
```diff
@@ -1260,8 +1260,10 @@ field(r, "fld") // r is an identifier, "fld" is a string literal
 r.fieldNames()
 fieldNames(r)
 // record update: [r EXCEPT !.f = e]
-r.with(f, e)
-with(r, f, e)
+r.with("f", e)
+with(r, "f", e)
```
shonfeder, line 1264: Shall we also hide this from users? It has the same problem as `field`: it is not a typeable operator, because we have no type for record labels. In fact it was in an attempt to supply record field names by reference that Lasaro reported the unecpected behavior reported in #816.

quint/src/ToIrListener.ts
```diff
@@ -591,19 +600,66 @@ export class ToIrListener implements QuintListener {
...
+      // field: expr
       const id = this.idGen.nextId()
       this.sourceMap.set(id, this.loc(ctx))
+      const nameEx: QuintEx = {
+        id,
+        kind: 'str',
+        value: ctx.IDENTIFIER()?.text!,
+      }
+      // the 'Tup' id is not needed, as the tuple is unwrapped
+      this.exprStack.push({
```
shonfeder, line 616: Answers to these questions would be helpful to add to the comment: 

- Why is the tup unwrapped? 
- What is a tup doing here in the firs place?

quint/testFixture/SuperSpec.qnt
```diff
@@ -179,6 +179,7 @@ module withConsts {
   val test_list_nth = [2, 3, 4][2]
   val test_record = { name: "igor", year: 1981 }
   val test_record2 = Rec("name", "igor", "year", 1981)
+  val test_record3 = { name: "quint", year: 2023, ...test_record }
   val test_set = Set(1, 2, 3)
```
shonfeder, line 183: It looks like the grammar supports multiple splats, like `{...r1, ...r2, f:2}`. I see we error out in that case, but don't see a test for that. Could you add one?

quint/src/ToIrListener.ts
```diff
@@ -591,19 +600,66 @@ export class ToIrListener implements QuintListener {
...
+    const spreads = elems.filter(e => e.kind === 'app' && e.args.length === 1)
+    const pairs = elems.map(e => (e.kind === 'app' && e.args.length === 2 ? e.args : [])).filter(es => es.length > 0)
+
+    if (spreads.length === 0) {
+      // { field1: value1, field2: value2 }
+      this.pushApplication(ctx, 'Rec', pairs.flat())
+    } else if (spreads.length > 1) {
+      // error
+      const msg = 'QNT012: ... may be used once in { ...record, <fields> }'
+      this.pushError(ctx, msg)
```
shonfeder, line 646: If we were using the `{r with f1: n}` syntax this would be made impossible by the grammar.

shonfeder (changes requested):
I see one test missing (unless I am missing it) and think comments could be improved a bit. Otherwise LGTM :)

doc/lang.md
konnov: We can remove it from the manual. However, it's important to know about this operator, as the spread operator is immediately translated into `with`. So it may happen that the users would see a message that mentions `with`, e.g., in REPL.

konnov: > It has the same problem as field: it is not a typeable operator, because we have no type for record labels. In fact it was in an attempt to supply record field names by reference that Lasaro reported the unecpected behavior reported in https://github.com/informalsystems/quint/issues/816.

I think these are two separate questions:
 - 1. Can we specify the generic type of `field` and `with` in our type system?
 - 2. Can we specify a concrete type of `field` and `with` when they are called with string literals?

The answer to the first question is "no". The answer to the second question is "yes". In principle, I don't see how it would be different from the case of `Set(...)`, which requires varargs. There are a few special operators, and we are trying to keep their number to minimum. Perhaps, it's an issue of documentation.

quint/testFixture/SuperSpec.qnt
konnov: Yeah, I will add another test

quint/src/ToIrListener.ts
konnov: This is true. However, it makes no difference for the end user.

konnov:
@shonfeder, I have addressed your comments

doc/lang.md
shonfeder: The problem is that the type we give it is incorrect: it does not take a value of type string (as the field argument). It only takes a string literal. This means users may try to use value of type string to access a field value, which should be correct according to the typing given, but instead we error out. The problem I see is not that type is not generic, but that the type is incorrect: the operator does not work over
Values of the string.

shonfeder (approved):
LGTM!

doc/lang.md
konnov: Do you mean the type in the documentation or elsewhere?

PR #884, opened 2023-05-11
save ITF traces in tests

### LINKED ISSUE #288: Extend `_lastTrace` when evaluating actions
Currently, `_lastTrace` is populated only when `_test` or `_testOnce` is called. We should also extend the trace when the user is manually invoking actions, as in e.g.:

```
>>> ...
>>> Init
>>> Next
>>> // expect _lastTrace to contain two states
```

### LINKED ISSUE #877: Optionally save ITF traces for passing and failing tests
In the context of demonstrating #874 to the IBC community, it would be great, if we could emit an ITF trace for the tests. Currently, only the `run` command is emitting ITF traces, but it is not able to run tests. It should be easy to emit traces in `quint test` too.

konnov:
Closes #877. This PR adds the option `--out-itf` to `quint test`, which specifies the filename prefix. When this option is specified, `quint test` saves the best trace in ITF for every test, passing or failing.

Saving to ITF works. However, all traces generated by tests are empty now, since we have to fix #288. I will check today, how fast it could be fixed.

### DIFF
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -596,6 +596,32 @@ rm out-itf-example.itf.json
 ]
 ```
 
+### Test outputs ITF
+
+TODO: output states after fix: https://github.com/informalsystems/quint/issues/288
+
+<!-- !test in test itf -->
+```
+quint test --out-itf=coin_ \
+  ../examples/solidity/Coin/coin.qnt
+cat coin_sendWithoutMintTest.itf.json | jq '.vars'
+rm coin_sendWithoutMintTest.itf.json
+cat coin_mintSendTest.itf.json | jq '.vars'
+rm coin_mintSendTest.itf.json
+```
+
+<!-- !test out test itf -->
+```
+[
+  "minter",
+  "balances"
+]
+[
+  "minter",
+  "balances"
+]
+```
+
 ### OK REPL tutorial
 
 The REPL tutorial is reproducible in REPL.
--- a/quint/src/cli.ts
+++ b/quint/src/cli.ts
@@ -104,6 +104,10 @@ const testCmd = {
         desc: 'name of the main module (by default, computed from filename)',
         type: 'string',
       })
+      .option('out-itf', {
+        desc: 'specify the name prefix of test traces in the Informal Trace Format',
+        type: 'string',
+      })
       .option('max-samples', {
         desc: 'the maximum number of successful runs to try for every randomized test',
         type: 'number',
--- a/quint/src/cliCommands.ts
+++ b/quint/src/cliCommands.ts
@@ -312,6 +312,17 @@ export async function runTests(prev: TypecheckedStage): Promise<CLIProcedure<Tes
       maxSamples: testing.args.maxSamples,
       rng,
       verbosity: verbosityLevel,
+      onTrace: (name: string, status: string, vars: string[], states: QuintEx[]) => {
+        if (testing.args.outItf) {
+          const trace = toItf(vars, states)
+          if (trace.isRight()) {
+            const jsonObj = addItfHeader(prev.args.input, status, trace.value)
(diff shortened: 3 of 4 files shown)


quint/io-cli-tests.md
```diff
@@ -596,6 +596,32 @@ rm out-itf-example.itf.json
 ]
 ```
 
+### Test outputs ITF
+
+TODO: output states after fix: https://github.com/informalsystems/quint/issues/288
+
+<!-- !test in test itf -->
+```
+quint test --out-itf=coin_ \
```
thpani, line 605:
```suggestion
quint test --out-itf=coin_ \
```

Maybe `--output-prefix` (I assume we would also use this prefix for any other additional file output)?
And drop the `_` from the parameter and add it programmatically instead?

IMO that would also be more uniform with our treatment in Apalache.

quint/src/runtime/testing.ts
```diff
@@ -84,6 +85,13 @@ export function compileAndTest(
   const recorder = newTraceRecorder(options.verbosity, options.rng)
   const ctx = compile(modules, sourceMap, lookupTable, types, main.name, recorder, options.rng.next)
 
+  const saveTrace = (name: string, status: string) => {
+    // save the best traces are reported by the recorder
+    const states =
+      recorder.getBestTrace().args.map(e => e.toQuintEx(zerog))
```
thpani, line 91: I think the concept of "best" needs to be explained here.

thpani (commented):
Cool!

> `quint test` saves the best trace in ITF for every test

What does "best" mean, in this context? It's not obvious to me.

Also, you probably saw this coming, but I have naming suggestions for the flag 😆

konnov:
I have fixed the output option to `--prefix`

quint/src/cli.ts
```diff
@@ -248,6 +253,20 @@ const docsCmd = {
   handler: (args: any) => load(args).then(chainCmd(docs)).then(outputResult),
 }
 
+const validate = (argv: any) => {
+  if (argv.output && typeof argv.output === 'string') {
+    const output = argv.output
+    if (!output.endsWith('.itf.json')) {
+      throw new Error(`Unexpected format in --output: ${output}`)
+    }
+    if (!output.includes('{}') && !output.includes('{#}')) {
```
thpani, line 262: Shouldn't we always require `{}`?

quint/src/cliCommands.ts
```diff
@@ -312,6 +315,18 @@ export async function runTests(prev: TypecheckedStage): Promise<CLIProcedure<Tes
       maxSamples: testing.args.maxSamples,
       rng,
       verbosity: verbosityLevel,
+      onTrace: (index: number, name: string, status: string, vars: string[], states: QuintEx[]) => {
+        if (outputTemplate && outputTemplate.endsWith('.itf.json')) {
+          const filename = outputTemplate.replaceAll('{}', name).replaceAll('{#}', index)
```
thpani, line 320: So if there's no `{#}`, the file will be overwritten with the latest run.
What if there's no `{}` and multiple tests?

thpani (approved):
Approving so you can merge, but I think `{}` should be non-optional.

quint/src/cli.ts
konnov: if there is no `{}` and there is no `{#}`, the arguments parser will throw an error. I am assuming that both `{}` and `{#}` are unique.

quint/src/cliCommands.ts
konnov: The arguments validator is checking for that. Initially, the check was in `runTests`, but I've decided to move it in the args parser.

quint/src/cli.ts
konnov: I can imagine a case, where test names are less relevant than their indices. For example, when we just want to process all of them.

PR #887, opened 2023-05-11
editor-plugins/vim: improve examples of how to adopt the provided syntax highlighting

cason:
Two minor contributions with this PR, both of them in the syntax file provided for Vim text editor. Each commit refers to one proposed change:

1. The `autogroup` setup can be much simple, 1 line instead of 4 on `~/.vimrc`. Also, using `runtime` is more adequate considering arbitrary setups. In any case, the previous version didn't work, as `~/vim` was used instead of the default `~/.vim`.
2. Syntax highlighting can be configured per file, using `modeline`. Namely, by adding `// vim: syntax=quint` to the end of a file, we instruct Vim to load the Quint syntax for the file. This could be a good practice for published Quint files.


konnov (approved):
Looks good. Thank you for fixing this!

Issue #886, opened 2023-05-11
When `run` is evaluated, it should emit a trace


konnov:
A follow up of #877. We should pass a trace to the recorder, whenever `run` is evaluated.

PR #890, opened 2023-05-14
Save ITF traces in `test`

konnov:
This a relatively quick feature for the demo. It collects traces in `quint test`. We should revisit this code later and see if it could be done better e.g. in the context of #886.

**update:** fixed the 3x slowdown in `ics20`, which was caused by massive transformations in the testing framework.


quint/io-cli-tests.md
```diff
@@ -621,13 +621,28 @@ exit $exit_code
...
+  [
+    "charlie",
+    0
+  ],
+  [
+    "eve",
+    0
+  ],
+  [
+    "null",
```
bugarela, line 643: I don't think this null should be here

quint/src/runtime/impl/runtimeValue.ts
```diff
@@ -244,6 +245,68 @@ export const rv = {
...
+        }
+
+        default:
+          // no other case should be possible, produce the set of all integers
+          return rv.mkInfSet('Int')
+      }
+
+    default:
+      // no other case should be possible, produce the set of all integers
+      return rv.mkInfSet('Int')
```
bugarela, line 306: Should we introduce some error reporting instead of this arbitrary defaults?

quint/src/runtime/testing.ts
```diff
@@ -118,11 +120,24 @@ export function compileAndTest(
         for (; nsamples <= options.maxSamples; nsamples++) {
           // record the seed value
           seed = options.rng.getState()
-          // run the test
+          // reset the trace
           recorder.onRunCall()
```
bugarela, line 124: Perhaps this comment belongs on the next line
```suggestion
          recorder.onRunCall()
          // reset the trace
```

quint/io-cli-tests.md
konnov: "null" is an account name here. This is probably confusing :))

quint/src/runtime/impl/runtimeValue.ts
konnov: Yep, fixed that

quint/src/runtime/impl/runtimeValue.ts
```diff
@@ -244,6 +245,74 @@ export const rv = {
...
+          return just(rv.mkRecord(pairs))
+        }
+
+        default:
+          // no other case should be possible
+          return none()
+      }
+
+    default:
+      // no other case should be possible, produce the set of all integers
```
bugarela, line 311:
```suggestion
      // no other case should be possible
```

bugarela (approved):

PR #899, opened 2023-05-19
Implement import... from for instances

### LINKED ISSUE #891: Instantiating a module from a file is not supported by the grammar
What is the right syntax for instantiating a module from a file?

- `import MyModule(param = 0) as MM from "./mymodule"` 
- `import MyModule(param = 0) from "./mymodule" as MM` 

whatever it is, it is not expressed in the grammar and implemented, according to @konnov.

konnov:
Closes #891. This PR add the missing case of `import ... from` for instances.


quint/testFixture/_1030const.json
bugarela, line 1: Seems like some fixtures were added but are not being tested

bugarela (approved):

quint/testFixture/_1030const.json
konnov: True. Fixed that.

Issue #897, opened 2023-05-19
VSCode plugin hovers are too noisy


bugarela:
Specially when hovering over defs, the hover information is not very useful.

![image](https://github.com/informalsystems/quint/assets/18356186/47597ee1-7a6a-4431-9ca4-7eae26c35c84)

PR #900, opened 2023-05-19
Prepare quint for LSP hover improvements

bugarela:
Hello :octocat: 

I'm working on #897 and these are some changes made on the compiler side to make improvements on the LSP possible. I'm opening this as a separate PR to facilitate reviewing.

1. `nondet` defs have a separate parsing rule and didn't have the proper loc for the definition part of the let-in. This was messing up hovering on `nondet` statements.
2. Types are probably the most important information to show on hovers. However, some types are really hard to read, such as long records. I'm now taking advantage of the recently implemented pretty printer to show those types in the plugin :sparkles:. For that, I had to add some extra functions to the pretty printer.
3. Hovering over definitions shows way too much text. `IRFinder` can now find definitions so the LSP can get access to the `QuintDef` for the definition and display only relevant information. 

  - None of this should be user facing before the plugin changes. 


quint/src/IRprinting.ts
```diff
@@ -35,8 +37,16 @@ export function moduleToString(quintModule: QuintModule): string {
...
+export function definitionToString(
+  def: QuintDef,
+  includeBody: boolean = true,
+  type: TypeScheme | undefined = undefined
+): string {
+  const typeAnnotation = isAnnotatedDef(def)
+    ? `: ${typeToString(def.typeAnnotation)}`
+    : type
+    ? `: ${typeSchemeToString(type)}`
+    : ''
```
konnov, line 49: Could you add some parentheses here? I am not sure about the priorities

quint/src/graphics.ts
```diff
@@ -123,6 +121,99 @@ export function prettyQuintEx(ex: QuintEx): Doc {
   }
 }
 
+export function prettyQuintDef(
+  def: QuintDef,
+  includeBody: boolean = true,
+  type: TypeScheme | undefined = undefined
```
konnov, line 127:
```suggestion
  type?: TypeScheme
```

I think this is an equivalent but shorter parameter declaration :)

quint/src/graphics.ts
```diff
@@ -123,6 +121,99 @@ export function prettyQuintEx(ex: QuintEx): Doc {
...
+): Doc {
+  const typeAnnotation = isAnnotatedDef(def)
+    ? [text(': '), prettyQuintType(def.typeAnnotation)]
+    : type
+    ? [text(': '), prettyTypeScheme(type)]
+    : []
```
konnov, line 133: Could you add parentheses here?

quint/src/graphics.ts
```diff
@@ -123,6 +121,99 @@ export function prettyQuintEx(ex: QuintEx): Doc {
...
+        richtext(chalk.magenta, qualifierToString(def.qualifier)),
+        text(' '),
+        richtext(chalk.blue, def.name),
+        ...typeAnnotation,
+      ])
+      return includeBody ? group([header, text(' = '), prettyQuintEx(def.expr)]) : header
+    }
+    default:
+      // TODO, not used for now
+      return text(definitionToString(def))
```
konnov, line 146: would it be better to return something like `<error>`, since this is an unexpected case?

quint/src/graphics.ts
```diff
@@ -123,6 +121,99 @@ export function prettyQuintEx(ex: QuintEx): Doc {
...
+  }
+}
+
+export function prettyTypeScheme(scheme: TypeScheme): Doc {
+  const [vars, type] = canonicalTypeScheme(scheme)
+  const varsDoc =
+    vars.length > 0 ? group([text('∀ '), docJoin([text(','), line()], vars.map(text)), text(' . ')]) : text('')
```
konnov, line 153: I am not sure many people would understand the role of the quantifier here. Would not it be better to use some form of a template-like syntax, e.g., `[x, y, z]. type`?

quint/src/types/printing.ts
```diff
@@ -28,6 +29,13 @@ export function constraintToString(c: Constraint): string {
  * @returns a string with the formatted type scheme
  */
 export function typeSchemeToString(t: TypeScheme): string {
+  const [vars, type] = canonicalTypeScheme(t)
+
+  const varsString = vars.length > 0 ? `∀ ${vars.join(', ')} . ` : ''
```
konnov, line 34: Also here, I would propose to use a more conventional syntax like `[x, y, z]. type`

konnov (approved):
Looks good! I've left a few comments on types syntax

quint/src/IRprinting.ts
bugarela: Actually, I can't. `prettier` removes it. I've added a comment to clarify priorities instead.

quint/src/graphics.ts
bugarela: same as https://github.com/informalsystems/quint/pull/900#discussion_r1200396964

quint/src/graphics.ts
bugarela: I don't see how people would recognize/understand templates more than universal quantification.

Thinking further, I think we should have two forms of printing signatures as we have to forms of defining signatures in our syntax. The `∀ a . (a) => a` like form should be used for adopters of the haskell-like signatures, while for the rest, we could try to mimic rust/java/ts-like generics: `foo<a>(param: a): a`

bugarela: I've opened an [issue](https://github.com/informalsystems/quint/issues/901) to discuss this further. I'll not block this PR on your comment as this was not introduced by the PR itself.

PR #902, opened 2023-05-22
Fix problem with docstring parsing

bugarela:
Hello :octocat: 

I noticed some docstrings were not being properly parsed while playing with the plugin. Investigating, I found two issues:
1. We were only considering docstrings on `opdef`s when building the IR
2. For nested definitions (let-ins), the docstring was being added to the innermost definition, while the grammar only allows docstrings for the outtermost.

This fixes both.


quint/src/generated/Quint.g4
```diff
@@ -13,8 +13,8 @@ grammar Quint;
 // entry point for the parser
 modules : module+ EOF;
```
konnov, line 14: Shall we introduce `documentedModule` similar to `documentedUnit`?

konnov (approved):
Looks good!

Issue #888, opened 2023-05-11
Handle all verification errors


shonfeder:
Ensure we have test cases and useful error reporting for all error types returned by Apalache.

shonfeder:
Thomas has taken care of this in the course of his work.

PR #912, opened 2023-05-26
Add ofItf function

shonfeder:
Adds a function to convert ITF traces into trades represented as an array of
`QuintEx`s.

Required for #888

### DIFF
--- a/quint/src/itf.ts
+++ b/quint/src/itf.ts
@@ -175,7 +175,7 @@ export function toItf(vars: string[], states: QuintEx[]): Either<string, ItfTrac
 
 export function ofItf(itf: ItfTrace): QuintEx[] {
   // Benign state to synthesize ids for the Quint expressions
-  var nextId = 0n
+  let nextId = 0n
   // Produce the next ID in sequence
   const getId = (): bigint => {
     const id = nextId
@@ -184,23 +184,23 @@ export function ofItf(itf: ItfTrace): QuintEx[] {
   }
 
   const ofItfValue = (value: ItfValue): QuintEx => {
-    const withId = { id: getId() }
+    const id = getId()
     if (typeof value === 'boolean') {
-      return { ...withId, kind: 'bool', value }
+      return { id, kind: 'bool', value }
     } else if (typeof value === 'string') {
-      return { ...withId, kind: 'str', value }
+      return { id, kind: 'str', value }
     } else if (typeof value === 'number') {
-      return { ...withId, kind: 'int', value: BigInt(value) }
+      return { id, kind: 'int', value: BigInt(value) }
     } else if (Array.isArray(value)) {
-      return { ...withId, kind: 'app', opcode: 'List', args: value.map(ofItfValue) }
+      return { id, kind: 'app', opcode: 'List', args: value.map(ofItfValue) }
     } else if (isBigint(value)) {
-      return { ...withId, kind: 'int', value: BigInt(value['#bigint']) }
+      return { id, kind: 'int', value: BigInt(value['#bigint']) }
     } else if (isTup(value)) {
-      return { ...withId, kind: 'app', opcode: 'Tup', args: value['#tup'].map(ofItfValue) }
+      return { id, kind: 'app', opcode: 'Tup', args: value['#tup'].map(ofItfValue) }
     } else if (isSet(value)) {
-      return { ...withId, kind: 'app', opcode: 'Set', args: value['#set'].map(ofItfValue) }
+      return { id, kind: 'app', opcode: 'Set', args: value['#set'].map(ofItfValue) }
     } else if (isUnserializable(value)) {
-      return { ...withId, kind: 'name', name: value['#unserializable'] }
+      return { id, kind: 'name', name: value['#unserializable'] }
     } else if (isMap(value)) {
       const args = value['#map'].map(([key, value]) => {
         const k = ofItfValue(key)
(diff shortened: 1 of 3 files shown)


quint/src/itf.ts
```diff
@@ -90,18 +119,34 @@ export function toItf(vars: string[], states: QuintEx[]): Either<string, ItfTrac
...
-            return merge(ex.args.map(exprToItf)).mapRight(pairs => {
-              return { '#map': pairs.map(p => p['#tup']) }
-            })
+            return merge(
+              // Convert all the entries of the map
+              ex.args.map(exprToItf)
+            ).chain(pairs =>
+              merge(
+                // Quint represents map entries as tuples, but in ITF they are 2 element arrays,
+                // so we unpack all the ITF tuples into arrays
```
bugarela, line 141: Should this be
```suggestion
                // so we unpack all the quint tuples into arrays
```
?

quint/src/itf.ts
```diff
@@ -126,6 +173,60 @@ export function toItf(vars: string[], states: QuintEx[]): Either<string, ItfTrac
   })
 }
 
-// export function ofItf(itf: any[]): Either<string, QuintEx[]> {
-//   const stateToExpr = (any: )
-// }
+export function ofItf(itf: ItfTrace): QuintEx[] {
```
bugarela, line 176: I'm far from being a specialist in English propositions, but perhaps `fromItf` makes more sense here

quint/src/itf.ts
```diff
@@ -126,6 +173,60 @@ export function toItf(vars: string[], states: QuintEx[]): Either<string, ItfTrac
...
+  // Benign state to synthesize ids for the Quint expressions
+  var nextId = 0n
```
bugarela, line 178:
```suggestion
  let nextId = 0n
```

quint/src/itf.ts
```diff
@@ -126,6 +173,60 @@ export function toItf(vars: string[], states: QuintEx[]): Either<string, ItfTrac
...
+      }
+    } else if (typeof value === 'object') {
+      // Any other object must represent a record
+      // For each key/value pair in the object, form the quint expressions representing
+      // the record field and value
+      const args = Object.keys(value)
+        .filter(key => key !== '#meta') // Must be removed from top-level ojects representing states
+        .map(f => [{ id: getId(), kind: 'str', value: f }, ofItfValue(value[f])] as [QuintStr, QuintEx])
+        .flat() // flatten the converted pairs of fields into a single array
+      return { ...withId, kind: 'app', opcode: 'Rec', args }
```
bugarela, line 224: Very minor aesthetic thing: you can do this
```suggestion
    const id = getId()
    if (typeof value === 'boolean') {
      return { id, kind: 'bool', value }
    } else if (typeof value === 'string') {
      return { id, kind: 'str', value }
    } else if (typeof value === 'number') {
      return { id, kind: 'int', value: BigInt(value) }
    } else if (Array.isArray(value)) {
      return { id, kind: 'app', opcode: 'List', args: value.map(ofItfValue) }
    } else if (isBigint(value)) {
      return { id, kind: 'int', value: BigInt(value['#bigint']) }
    } else if (isTup(value)) {
      return { id, kind: 'app', opcode: 'Tup', args: value['#tup'].map(ofItfValue) }
    } else if (isSet(value)) {
      return { id, kind: 'app', opcode: 'Set', args: value['#set'].map(ofItfValue) }
    } else if (isUnserializable(value)) {
      return { id, kind: 'name', name: value['#unserializable'] }
    } else if (isMap(value)) {
      const args = value['#map'].map(([key, value]) => {
        const k = ofItfValue(key)
        const v = ofItfValue(value)
        return { id: getId(), kind: 'app', opcode: 'Tup', args: [k, v] } as QuintApp
      })
      return {
        id,
        kind: 'app',
        opcode: 'Map',
        args,
      }
    } else if (typeof value === 'object') {
      // Any other object must represent a record
      // For each key/value pair in the object, form the quint expressions representing
      // the record field and value
      const args = Object.keys(value)
        .filter(key => key !== '#meta') // Must be removed from top-level ojects representing states
        .map(f => [{ id: getId(), kind: 'str', value: f }, ofItfValue(value[f])] as [QuintStr, QuintEx])
        .flat() // flatten the converted pairs of fields into a single array
      return { id, kind: 'app', opcode: 'Rec', args }
```

bugarela (approved):
Looks great!

quint/src/itf.ts
shonfeder: Confusing, right :) But this actually unpacking ITF tuples. First all the arguments are sent to the toItf converter, which produces a list of itf tuples, but then we need to convert those int pairs of itf values, since itf doesn’t represent maps as a list of tuples, but ad a list of 2 element lists.

the logic could maybe be improved here, but I was trying not to get too far into fiddling with the previous code.

quint/src/itf.ts
shonfeder: Oh yes. Thanks. I forgot JS has two different mutable variables, one of which is not to be used lol. Thanks.

quint/src/itf.ts
shonfeder: That is nicer. Thanks :)

quint/src/itf.ts
bugarela: Ah I see it now, thanks! I was not following the code on this part before, but from the previous comment line, this seemed wrong. Makes sense now!

quint/src/itf.ts
shonfeder: I think either can work grammatically. I don't have a strong preference here, but assuming you don't either from the approval and tone of comment :)

If I'm wrong and you end up finding this confusing, let me know and i'll fix it in a followup!

PR #913, opened 2023-05-26
fix trace selection in the simulator

konnov:
This PR fixes a bug in reporting by `quint run`, which I have introduced before the presentation in a rush in #890. The simulator was quickly finding a violation but it was reporting the trace as no-violation. Also, added a test to check that the found trace is reported properly.

This is a critical fix actually. Can we cut a new release after this is merged?

### DIFF
--- a/quint/cli-tests.md
+++ b/quint/cli-tests.md
@@ -319,6 +319,6 @@ fi
 
 <!-- !test exit 1 -->
 <!-- !test check lottery - Run -->
-    quint run --max-samples=10000 --max-steps 10 --seed=0x29f8e8021fae9 \
-      --invariant noBuyInDrawingInv --main lotteryMempool \
+    quint run --max-samples=10000 --max-steps=10 --seed=0x29f8e8021fae9 \
+      --invariant=noBuyInDrawingInv --main=lotteryMempool \
       ../examples/solidity/icse23-fig7/lottery.qnt
--- a/quint/cli-tests.md
+++ b/quint/cli-tests.md
@@ -314,3 +314,11 @@ fi
 
 <!-- !test check bug843pureValCache - Syntax/Types & Effects/Unit tests -->
     quint test ./testFixture/bug843pureValCache.qnt
+
+### OK on run lottery
+
+<!-- !test check lottery - Run -->
+    quint run --max-samples=10000 --max-steps 10 --seed=0x29f8e8021fae9 \
+      --invariant noBuyInDrawingInv --main lotteryMempool \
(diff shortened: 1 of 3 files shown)


quint/cli-tests.md
```diff
@@ -314,3 +314,11 @@ fi
...
 <!-- !test check bug843pureValCache - Syntax/Types & Effects/Unit tests -->
     quint test ./testFixture/bug843pureValCache.qnt
+
+### OK on run lottery
+
+<!-- !test exit 1 -->
+<!-- !test check lottery - Run -->
+    quint run --max-samples=10000 --max-steps 10 --seed=0x29f8e8021fae9 \
+      --invariant noBuyInDrawingInv --main lotteryMempool \
+      ../examples/solidity/icse23-fig7/lottery.qnt
```
bugarela, line 324: Shall we use `=` either on all options or none?
```suggestion
    quint run --max-samples=10000 --max-steps=10 --seed=0x29f8e8021fae9 \
      --invariant=noBuyInDrawingInv --main=lotteryMempool \
      ../examples/solidity/icse23-fig7/lottery.qnt
```

bugarela:
```suggestion
    quint run --max-samples 10000 --max-steps 10 --seed 0x29f8e8021fae9 \
      --invariant noBuyInDrawingInv --main lotteryMempool \
      ../examples/solidity/icse23-fig7/lottery.qnt
```

quint/src/runtime/trace.ts
```diff
@@ -255,28 +256,30 @@ export const newTraceRecorder = (verbosityLevel: number, rng: Rng) => {
...
-          bestTraceSeed = runSeed
-          bestTrace.result = outcome
-          bestTrace.args = trace
-        }
+      const notOk = fromResult(outcome)
+      const prevNotOk = fromResult(bestTrace.result)
+
+      // override the best trace only if:
+      //  - there is an error, the new trace is shorter, or the old trace is non-error;
+      //  - there is no error, the new trace is longer, and there was no error before.
```
bugarela, line 273: I find the old comments about the end-goal of this very important. Can we add that here as well?
```suggestion
      // Prefer short traces for error, and longer traces for non error.
      // Therefore, override the best trace only if:
      //  - there is an error, the new trace is shorter, or the old trace is non-error;
      //  - there is no error, the new trace is longer, and there was no error before.
```

bugarela (approved):
LGTM! I can cut a release after this is merged.

PR #914, opened 2023-05-27
Print counterexamples on variant violations and deadlocks

shonfeder:
Towards #888

This adds handling for variant violations and deadlocked errors reported by Apalache.

There are also tests added for CLI configuration via `--step` and `--init` flags.

### DIFF
--- a/quint/src/cliCommands.ts
+++ b/quint/src/cliCommands.ts
@@ -36,12 +36,13 @@ import { QuintError, quintErrorToString } from './quintError'
 import { TestOptions, TestResult, compileAndTest } from './runtime/testing'
 import { newIdGenerator } from './idGenerator'
 import { SimulatorOptions, compileAndRun } from './simulation'
-import { toItf } from './itf'
+import { ofItf, toItf } from './itf'
 import { printExecutionFrameRec, printTrace, terminalWidth } from './graphics'
 import { verbosity } from './verbosity'
 import { Rng, newRng } from './rng'
 import { fileSourceResolver } from './sourceResolver'
 import { verify } from './quintVerifier'
+import { ExecutionFrame } from './runtime/trace'
 
 export type stage = 'loading' | 'parsing' | 'typechecking' | 'testing' | 'running' | 'documentation'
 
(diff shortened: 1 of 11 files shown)


quint/src/cliCommands.ts
```diff
@@ -558,10 +563,22 @@ export async function verifySpec(prev: TypecheckedStage): Promise<CLIProcedure<V
...
       .map(_ => ({ ...verifying, status: 'ok', errors: [] } as VerifiedStage))
-      .mapLeft(err => ({
-        msg: err.explanation,
-        stage: { ...verifying, status: 'failure', errors: err.errors },
-      }))
+      .mapLeft(err => {
+        const trace: QuintEx[] | undefined = err.traces ? ofItf(err.traces[0]) : undefined
+        if (trace !== undefined) {
+          // Always print the conterexample, unless the output is being directed to one of the outfiles
+          const verbosityLevel = prev.args.out || prev.args.outItf ? 0 : 2
```
thpani, line 570: Meaning of the magic constants 0 and 2 is not clear to me.
Should we have names for these?

quint/src/cliCommands.ts
```diff
@@ -558,10 +563,22 @@ export async function verifySpec(prev: TypecheckedStage): Promise<CLIProcedure<V
...
     res
       .map(_ => ({ ...verifying, status: 'ok', errors: [] } as VerifiedStage))
-      .mapLeft(err => ({
-        msg: err.explanation,
-        stage: { ...verifying, status: 'failure', errors: err.errors },
-      }))
+      .mapLeft(err => {
+        const trace: QuintEx[] | undefined = err.traces ? ofItf(err.traces[0]) : undefined
+        if (trace !== undefined) {
+          // Always print the conterexample, unless the output is being directed to one of the outfiles
```
thpani, line 569: To double-check my understanding:
This is because on verify, we always expect a counterexample to be printed, right?

thpani (approved):
Looks great!

quint/src/cliCommands.ts
shonfeder: Probably! This mysterious numberings is pervasive throughout this file unfortunately. I’ll make a followup to note the need to apply some cleanup to the verbosity system.

quint/src/cliCommands.ts
shonfeder: That’s my assumption, yep! This is following the “spec” Igor wrote in the docs for the command: https://github.com/informalsystems/quint/blob/main/doc/quint.md#command-verify

quint/src/cliCommands.ts
shonfeder: Created an issue to follow up! https://github.com/informalsystems/quint/issues/915

shonfeder:
Thanks for the review!

quint/src/cliCommands.ts
konnov: We should use `defaultLevel` of `verbosity.ts`

PR #917, opened 2023-05-31
Incremental analysis

bugarela:
Hello :octocat:

This PR makes analysis incremental on new expressions evaluated in the REPL. When a REPL input is parsed as an expression, we compile it with a new function called `compileExpr` that uses a previous `CompilationState` stored in the `ReplState`.

Towards #618 


quint/src/repl.ts
```diff
@@ -668,3 +651,54 @@ function countBraces(str: string): [number, number, number] {
 
   return [nOpenBraces, nOpenParen, nOpenComments]
 }
+
+function evalExpr(
```
bugarela, line 655: This was not changed, I just extracted the evaluation part to its own function

quint/src/repl.ts
```diff
@@ -362,44 +376,41 @@ function loadFromFile(out: writer, state: ReplState, filename: string): boolean
...
 
       // save the definition history
       newState.defsHist = defsAndExprs[0]
+
+      const isError = !tryEval(out, newState, 'val __loaded = true')
+      if (isError) {
+        return false
+      }
+
+      // remove 'val __loaded = true' from the definiton history
```
konnov, line 392:
```suggestion
      // reset the definiton history to the one loaded from the file
```

quint/src/repl.ts
```diff
@@ -362,44 +376,41 @@ function loadFromFile(out: writer, state: ReplState, filename: string): boolean
...
     // whether an error occurred at some step
-    let isError = false
     newState.moduleHist = modulesAndRepl[0]
     if (modulesAndRepl.length > 1) {
       // found a REPL session
       const defsAndExprs = modulesAndRepl[1].split(replEnd)
 
       // save the definition history
       newState.defsHist = defsAndExprs[0]
+
```
konnov, line 386:
```suggestion
     // evaluate a dummy definition, in order to compile the definitions from the file
```

Shall we use `val _ = true` in the future? I guess, `_` would not work as expected now, but we should make it work at some point.

konnov (approved):
Great improvement! It looks like the same approach could be applied to top-level definitions too, except maybe imports

PR #930, opened 2023-06-07
Improve interfaces of precompilation phases

bugarela:
Hello :octocat: 

This is preparatory work towards incremental analysis. It changes the interfaces from several precompilation procedures to receive a list of definitions instead of a whole module, and also allows for an initial state to be set at the constructor. To avoid complex handling of states, the statefulness of `QuintAnalyzer` is now hidden under `analyzeModules` and `analyzeInc` functions.

Flattening is also wrapped under a new function called `flattenModules` which flattens all the modules handling dependencies between them. This complexity was extracted from `compile`. We also start to distinct unflattened modules from flattened modules in the type system.


konnov (approved):
Nice refactoring! I like the changes to the API. It looks like you have found a good way to do that without increasing complexity too much.

PR #939, opened 2023-06-14
expand on ponzi

konnov:
This PR extends the Ponzi example with interesting properties and tests.


Kukovec (approved):
LGTM

PR #944, opened 2023-06-15
Support imports & module instances in `quint verify`

### LINKED ISSUE #936: `verify` fails on `BinSearch10.qnt`
Using the example in #935. I do the following:

In tab 1:

```sh
$ cd $APALACHE_HOME
$ apalache-mc server --write-intermediate=true
```

In tab 2:

```
$ export APALACHE_DIST=$APALACHE_HOME/target/universal/apalache-0.40.3-SNAPSHOT/
$ cd $QUINT_HOME/examples/classic/sequential/BinSearch
$ quint verify --init=Init --step=Next --invariant=Postcondition BinSearch10.qnt
```

This is what I see in tab 1:

```
Error: internal error: unhandled verification error at pass SanyParser
    at handleVerificationFailure (/Users/igor/devl/informal/quint/quint/dist/src/quintVerifier.js:66:19)
    at Object.check (/Users/igor/devl/informal/quint/quint/dist/src/quintVerifier.js:83:47)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) 
```

This is what I see in tab 2:

```
  > Error parsing file StringSource(Qnt)                          E@09:37:46.988
  > Input was not a valid representation of the QuintIR: Definition has invalid `kind` field: LinkedHashMap(id -> 17, kind -> "instance", qualifiedName -> "I", protoName -> "BinSearch", overrides -> [[{"id":14,"name":"INPUT_SEQ"},{"id":11,"kind":"app","opcode":"List","args":[{"id":1,"kind":"int","value":1},{"id":2,"kind":"int","value":2},{"id":3,"kind":"int","value":3},{"id":4,"kind":"int","value":4},{"id":5,"kind":"int","value":5},{"id":6,"kind":"int","value":6},{"id":7,"kind":"int","value":8},{"id":8,"kind":"int","value":9},{"id":9,"kind":"int","value":10},{"id":10,"kind":"int","value":33}]}],[{"id":15,"name":"INPUT_KEY"},{"id":12,"kind":"int","value":7}],[{"id":16,"name":"INT_WIDTH"},{"id":13,"kind":"int","value":16}]], identityOverride -> false, fromSource -> "./BinSearch") E@09:37:46.989
```

thpani:
Flatten modules in `quint verify`, to support imports and const instantiation.

Closes #936

### DIFF
--- a/quint/src/cliCommands.ts
+++ b/quint/src/cliCommands.ts
@@ -554,7 +554,28 @@ export async function verifySpec(prev: TypecheckedStage): Promise<CLIProcedure<V
   const args = verifying.args
   // TODO error handing for file reads and deserde
   const loadedConfig = args.apalacheConfig ? JSON.parse(readFileSync(args.apalacheConfig, 'utf-8')) : {}
-  const parsedSpec = jsonStringOfOutputStage(pickOutputStage(prev))
+
+  // Flatten modules, replacing instances, imports and exports with their definitions
+  const { flattenedModules, flattenedTable, flattenedAnalysis } = flattenModules(
+    verifying.modules,
+    verifying.table,
+    verifying.idGen,
+    verifying.sourceMap,
+    verifying
+  )
+
+  // Pick main module, we only pass this on to Apalache
+  const mainArg = prev.args.main
+  const mainName = mainArg ? mainArg : basename(prev.args.input, '.qnt')
+  const main = flattenedModules.find(m => m.name === mainName)
+
+  if (!main) {
+    return cliErr(`module ${mainName} does not exist`, { ...verifying, errors: [], sourceCode: '' })
+  }
+
+  const veryfiyingFlat = { ...verifying, ...flattenedAnalysis, modules: [main], table: flattenedTable }
+  const parsedSpec = jsonStringOfOutputStage(pickOutputStage(veryfiyingFlat))
+
   // We need to insert the data form CLI args into thier appropriate locations
   // in the Apalache config
   const config = {
--- a/README.md
+++ b/README.md
@@ -153,9 +153,9 @@ completely implementing every pass.
 | [Records][]                       | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x:                |
 | [Discriminated unions][]          | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x: [244][]        | :x: [539][]        | :x:                | :x:                |
 | [Tuples][]                        | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :green_circle:     | :white_check_mark: |
-| [Imports][]                       | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x:                | :white_check_mark: |
+| [Imports][]                       | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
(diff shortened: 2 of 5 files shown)


quint/src/cliCommands.ts
```diff
@@ -554,7 +554,36 @@ export async function verifySpec(prev: TypecheckedStage): Promise<CLIProcedure<V
...
-  const parsedSpec = jsonStringOfOutputStage(pickOutputStage(prev))
+
+  // Flatten modules, replacing instances, imports and exports with their definitions
+  const [analysisErrors, analysisOutput] = analyzeModules(verifying.table, verifying.modules)
+
+  if (analysisErrors.length > 0) {
+    const errorLocator = mkErrorMessage(verifying.sourceMap)
+    const errors = Array.from(analysisErrors, errorLocator)
+    return cliErr('verification failed', { ...verifying, errors })
+  }
```
bugarela, line 565: I'm not sure if this is ready for review yet since it is in draft mode, but anyway, here's one comment:

You shouldn't need to run analysis again, as `prev` is a `TypecheckedStage` and therefore should have the analysis output in it's properties. irrc, there's no `analysisOutput` property itself, but there should be `types`, `effects` and `modes`, so you can construct the `AnalysisOutput` object from those.

konnov: I propose we first make things work in this PR and then refactor them to make code reuse

quint/src/cliCommands.ts
```diff
@@ -554,7 +554,36 @@ export async function verifySpec(prev: TypecheckedStage): Promise<CLIProcedure<V
...
+
+  const { flattenedModules, flattenedTable, flattenedAnalysis } = flattenModules(
+    verifying.modules,
+    verifying.table,
+    verifying.idGen,
+    verifying.sourceMap,
+    analysisOutput
+  )
```
shonfeder, line 573: iiuc, you can give the `flattenModules` the `verifying` object as its last argument:

```suggestion
  const { flattenedModules, flattenedTable, flattenedAnalysis } = flattenModules(
    verifying.modules,
    verifying.table,
    verifying.idGen,
    verifying.sourceMap,
    verifying
  )
```

It'd be nice if `flattenModules` just took a (sub-record) or the `TypecheckedStage`, so unpacking the other fields here weren't needed. But that's nothing to fuss with here ofc. :)

quint/src/cliCommands.ts
shonfeder: I don't think there's any reuse to be obtained here besides avoiding rerunning analysis.

shonfeder (approved):
Once the extra analysis is removed, it all looks good to me :)

quint/src/cliCommands.ts
bugarela: > I had naively assumed flattening would already be taken care of in the processing pipeline up to this point.

Currently we just call flattening at the REPL/compiler since we only need it for that stage. And it is not possible to just pre-process everything before the compiler because of its interactivity (i.e. if an import statement is written into the REPL, it will need to re-process a lot of stuff, including reflattening).

So this:

> I don't think there's any reuse to be obtained here besides avoiding rerunning analysis.

is correct.

On the topic of further refactor after this PR, I was thinking about introducing some sort of structure to aggregate results between "compilation" (parsing/analysis/flattening/compilation) stages, and I only now see this is pretty much what we have with these CLI stages. I wonder whether it's a good idea to use these stages beyond the CLI implementation, and then, as Shon mentions in his [comment](https://github.com/informalsystems/quint/pull/944#discussion_r1230986373) about arguments, avoid having a lot of arguments being propagated and use a single object with the aggregated information. You'll see in my next PR how cumbersome managing all those arguments can be.

thpani: Straightened out the implementation in [05f7c0b](https://github.com/informalsystems/quint/pull/944/commits/05f7c0b8d35e7dcb5402447d3ca4d72be6b238f2)

PR #945, opened 2023-06-15
add gradual ponzi

konnov:
One more example that may be useful for the workshops.


examples/solidity/GradualPonzi/gradualPonzi.qnt
```diff
@@ -0,0 +1,225 @@
...
+    if (amount > evm.balances.get(investor)) {
+      error: "Insufficient funds",
+      evm: evm,
+      ponzi: ponzi,
+     } else if (amount < MINIMUM_INVESTMENT) {
+      error: "Investment below minimum",
+      evm: evm,
+      ponzi: ponzi,
+    } else
+      pure val investorCut = amount / length(ponzi.investors)
```
bugarela, line 62: Can this integer division be changing the total supply?

bugarela (approved):
LGTM, just a few comments

examples/solidity/GradualPonzi/gradualPonzi.qnt
konnov: I have added `noLeftoversInv`, which shows that the contract may accumulate leftover tokens due to rounding. It's actually quite nice.

bugarela: Awesome, makes sense to me!

PR #952, opened 2023-06-15
Incremental REPL definitions

### LINKED ISSUE #932: Use incremental analysis on new REPL definitions
After #917, analysis is done incrementally for new expressions in the REPL, but not for new definitions

bugarela:
Hello :octocat: 

This PR makes it so new definitions typed in the REPL are evaluated incrementally :sparkles:. Notice that this definition can be an import/instance/export statement, so this requires some extra steps (than compiling expressions), as re-flattening and re-resolving names. Closes #932 

For that, the two main changes are:
1. The flattening module now has a new function to flatten a single def and include it in a flat module. The reason why I combined the responsibility of a. flattening a def and b. adding  it to the module in a single function is that (a) cannot really be done in isolation, as we need all of the module information (dependent modules, lookup table, etc) in order to do flattening, so I thought it made sense to create an `addDefToFlatModule` interface.
2. To initialize the REPL, either from scratch or from a file, we can no longer use the hack of evaluating a dummy definition, as that is incremental. So I've defined a function called `tryEvalHistory` that will trigger compilation for the whole module. The way to handle simulator built-ins was also updated to make sure they are properly defined and compiled.


quint/src/flattening.ts
```diff
@@ -72,349 +64,412 @@ export function flattenModules(
...
-          qualifier: 'pureval',
-          name: param.name,
-          expr,
-          typeAnnotation: constDef.typeAnnotation,
-          id: param.id,
-        },
-      ]
-    })
-  )
+  private flattenInstance(def: QuintInstance): FlatDef[] {
```
bugarela, line 185: This file from here on is pretty much unchanged. The only change is that we now use `this` instead of a `ctx` parameter`.

quint/src/repl.ts
```diff
@@ -478,41 +487,65 @@ function loadShadowVars(state: ReplState, context: CompilationContext): void {
...
 // In the future, we will declare them in a separate module.
-const simulatorBuiltins = `val ${lastTraceName} = [];
-def q::test(q::nruns, q::nsteps, q::init, q::next, q::inv) = false;
-def q::testOnce(q::nsteps, q::init, q::next, q::inv) = false;
-`
+function simulatorBuiltins(compilationState: CompilationState): FlatDef[] {
+  return [
+    parseDefOrThrow(compilationState, `val ${lastTraceName} = []`),
+    parseDefOrThrow(compilationState, `def q::test = (q::nruns, q::nsteps, q::init, q::next, q::inv) => false`),
+    parseDefOrThrow(compilationState, `def q::testOnce = (q::nsteps, q::init, q::next, q::inv) => false`),
```
konnov, line 494: Nice! Maybe we should introduce a whole module for that in the future?

quint/src/repl.ts
```diff
@@ -478,41 +487,65 @@ function loadShadowVars(state: ReplState, context: CompilationContext): void {
...
-function tryEval(out: writer, state: ReplState, newInput: string): boolean {
-  // output errors to the console in red
-  function printErrors(moduleText: string, context: CompilationContext) {
-    printErrorMessages(out, 'syntax error', newInput, moduleText, context.syntaxErrors)
-    printErrorMessages(out, 'static analysis error', newInput, moduleText, context.analysisErrors, chalk.yellow)
-    printErrorMessages(out, 'compile error', moduleText, newInput, context.compileErrors)
-    out('\n') // be nice to external programs
+function parseDefOrThrow(compilationState: CompilationState, text: string): FlatDef {
+  const result = parseExpressionOrUnit(text, '<builtins>', compilationState.idGen, compilationState.sourceMap)
+  if (result.kind === 'toplevel' && isFlat(result.def)) {
```
konnov, line 500: Does it mean that we would not be able to write nested definitions in REPL anymore?

konnov (approved):
This is amazing! Thanks for making definitions in REPL incremental!

quint/src/repl.ts
bugarela: Not at all, this is only used for parsing the builtins. Also, `toplevel` here is just the kind returned by `parseExpressionOrUnit` when it finds a unit, so that would also work for let-ins (assuming that's what you mean by nested definitions).

PR #958, opened 2023-06-16
WIP: Automate examples dashboard

### LINKED ISSUE #957: Automate examples dashboard generation
We currently have a md table in `./examples/README.md` used to list all our examples and which checks they are passing. However it is maintained by hand, and easily falls out of date.

We can instead automate this, so we ensure new examples are tested and that they are kept up to date with how the actual results turn out.

shonfeder:
Closes #957


examples/run-examples.sh
```diff
@@ -0,0 +1,24 @@
...
+export APALACHE_DIST=../quint/_build/apalache
+# Start the apalache server
+"$APALACHE_DIST"/bin/apalache-mc server &> /dev/null &
+
+# We recursively glob all the .qnt files, and run ./run-example.sh on each.
+#
+# Since the output of each run is a row in the README.md table,
+# sorting the outputs lexographically gets the results in the
+# right order.
+parallel ./run-example.sh ::: **/*.qnt | sort
```
shonfeder, line 24:
```suggestion
parallel ./run-example.sh ::: **/*.qnt | sort -f
```

shonfeder:
The sort is not matching perfectly, I think because I need to set the local, as per https://stackoverflow.com/a/29315/1187277

bugarela (approved):
This is super cool! At first I thought we should have this the other way around: generating integration tests based on the dashboard, so only things that are expected to succeed are checked. 

The way you did it makes a bit less convenient to spot failing tests, since it requires looking at the dashboard diff instead of having red lines printed in the CLI. However, we get the amazing benefit of seeing when a code change enables examples to go through one step further in the pipeline, so I think that's worth it!

One thing I'm not sure about is how this interacts with our existing integration tests. There is a lot of intersection on what the commands ran here vs the integration test, as well as the outcome (if this fails, integration tests will probably also fail and vice-versa). But replacing integration tests with this might affect significantly our development cycle, since we would 1. have to run one more local test command and 2. keep checking the dashboard file for git diffs and re-running commands to find failures.

The only concrete suggestion I have for now is unrelated to all of this :sweat_smile:. Should we move the `.sh` files and `README-text.md` into a new folder? They might confuse/distract people looking at the `examples` dir.

Awesome work :rocket:

konnov (approved):
That's nice! One feature that I am missing though is category breakdown. Maybe it is not as needed as I thought, but most views go to the cosmos and solidity examples.

shonfeder:
> One feature that I am missing though is category breakdown. Maybe it is not as needed as I thought, but most views go to the cosmos and solidity examples

I think the category is reflected in the directory in the path? It's a different representation, but simplifies the script quite a bit and i hope is just as easy to identify?

shonfeder:
Thanks for the reviews. I've moved the scripts into a separate directory and improved the documentation a bit.

shonfeder:
Oh whoops. This test was not marked as required but I had auto-merge in, so I accidentally merged this in with a broken check. I'll fix it!

examples/.scripts/README-text.md
```diff
@@ -0,0 +1,41 @@
+This directory contains examples of protocol specifications written in Quint.
```
thpani, line 1: I was confused why this is here. I think other people might stumble upon this file (e.g., via search) and have the same problem.

Can we somewhere note that this is a template, at least in the filename?

thpani (commented):
I'm late to the party, but one quick comment regarding the template .md in `.scripts`.

examples/.scripts/run-examples.sh
```diff
@@ -0,0 +1,25 @@
...
+export APALACHE_DIST=../quint/_build/apalache
+# Start the apalache server
+"$APALACHE_DIST"/bin/apalache-mc server &> /dev/null &
+
+# We recursively glob all the .qnt files, and run ./run-example.sh on each.
+#
+# Since the output of each run is a row in the README.md table,
+# sorting the outputs lexographically gets the results in the
+# right order.
+parallel "${SCRIPT_DIR}/run-example.sh" ::: **/*.qnt | env LC_ALL=C sort --ignore-case
```
thpani, line 25: Does this really work? We still have global mutable state in Apalache.

thpani (commented):
One more question

examples/.scripts/run-examples.sh
shonfeder: Do you have in mind where? I have been tearing the server via parallel calls (tho not rigorously) and have generally not had any trouble since I made the parsing threadsafe, as far as I can recall. But I have just started to see signs of worry with the river spec.

shonfeder: For context: all the txm tests we currently use for running the `verify` command run concurrently.

thpani: I was just thinking, in general, of Scala singletons

shonfeder: It had been working, but I’m seeing issues with the river spec just added, so maybe the others were working only because of how simple they were.

I’ll need to debug when I have a chance, and maybe we’ll end up needing to sequence this.

PR #960, opened 2023-06-16
Ensure entities are flattened in mode checker

### LINKED ISSUE #808: Effect checker unpacking(?) propagating wrong effects
Minimal example:
```
module test {
  pure def isProperChannelSetup(channelSetup: Set[str]): bool =
    tuples(Set("A", "B", "C"), channelSetup).forall((fromChain, toChain) =>
        Set(toChain, fromChain) == channelSetup
    )
}
```

I'm getting a:
```
error: [QNT200] pure def operators may not interact with state variables, but operator `isProperChannelSetup` performs temporal operations over variables v19,v21. Use temporal instead.
```

The effect for the lambda seems correct (a pure arrow effect), but then the result of the `forall` application results in a `Read & Temporal` effect which is not right.

bugarela:
Hello :octocat: 

It's a Friday miracle, I could finally fix #808 :tada: 

I thought unpacking was doing something wrong, then I though arrow effect simplification was doing something wrong... but in the end, it was not even an effect inference issue, but a problem in mode checking!

For mode checking arrow effects, we want to see if there are entities in the result that are not present on the parameters. When mode checking an effect like `(Read[v0, v1] & Temporal[v2, v3]) => Read[v0, v1] & Temporal[v2, v3]`, we were not properly handling unions (i.e. `[v0, v1]`), so we were saying that `v2 != [v2, v3]` and `v3 != [v2, v3]`, therefore `v2` and `v3` were new temporal entities and the overall mode had to be temporal, which of course is wrong.

So I fixed the mode checker to properly recurse over unions and produce a flat list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 ### Removed
 ### Fixed
+
+- Fixed a bug where sometimes static analysis would flag a mode error where
+  there isn't one (#960)
+
 ### Security
 
 ## v0.11.2 -- 2023-06-15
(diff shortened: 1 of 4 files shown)


shonfeder (approved):
Great debugging and lovely, precise fix!

Happy Friday :)

PR #962, opened 2023-06-19
Treat .qnt files as bluespec

thpani:
Add GH linguist override to treat `.qnt` files as Bluespec in `.gitattributes`.

This should work for all files with `.qnt` extension, similar to a `// -*- mode: Bluespec; -*-` mode line in each file.

For details, see https://github.com/github-linguist/linguist/blob/master/docs/overrides.md


konnov (approved):
Nice!

PR #964, opened 2023-06-19
Fix some examples

shonfeder:
This fixes some of the low hanging fruit in failing examples. It also fixes an oversight in the runner script, which left the apalache server running after the script ended.

I'm not sure if the approach to running tests with specs containing constants is agreeable to everyone. Afaik, the only alternative currently is to just fail any spec that runs tests and is parameterized on constants. Interested to know what y'all think here.

This also makes us ignore the new `river.qnt` spec in she examples for the moment. There is some sort of non-deterministic failure I'm trying to debug, and don't want it making main look broken in the meantime.


examples/classic/distributed/ClockSync/clockSync3.qnt
```diff
@@ -139,3 +139,11 @@ module ClockSync3 {
         }
     }
 }
+
+// The default instance of the clockSync spec
```
shonfeder, line 143: This is the only approach I’m aware of to make constant parameterized tests runnable. Please let me know if I’m missing a better approach.

examples/classic/sequential/BinSearch/BinSearch.qnt
```diff
@@ -106,8 +106,7 @@ module BinSearch {
     }
 
     // Computation step (lines 5-16)
-    action Next = any {
-        LoopIteration,
+    action step = any {LoopIteration,
```
bugarela, line 109:
```suggestion
    action step = any {
        LoopIteration,
```

examples/classic/distributed/ClockSync/clockSync3.qnt
bugarela: I think this is currently the only way

bugarela (approved):
LGTM

examples/classic/sequential/BinSearch/BinSearch.qnt
```diff
@@ -44,7 +44,7 @@ module BinSearch {
               -(MAX_UINT - wrapped)
 
     // Initialization step (lines 2-3)
-    action Init = all {
+    action init = all {
```
thpani, line 47: This is really confusing... the action names below are upper case, but `init` and `next` are lower case?

I understand that you need standardized names for the runner script, but I think consistent style is important for code examples.

quint/apalache-tests.md
```diff
@@ -35,51 +35,14 @@ error: Configuration error (see the manual): Operator invalidInit not found (use
 
 ## Successful verification
 
-### Can verify `../examples/language-features/booleans.qnt`
-
-<!-- !test check can check booleans.qnt -->
-```
-quint verify --init Init --step Next ../examples/language-features/booleans.qnt
```
thpani, line 42: Why are we removing these?

afaict the examples script runs the commands in a subshell. Does it fail if one of the commands fail?

thpani (commented):
Again, I'm late, but I have two questions/comments

quint/apalache-tests.md
shonfeder: It will fail in CI if the outcome is different then we expect.

examples/classic/sequential/BinSearch/BinSearch.qnt
shonfeder: Makes sense. I’ll fix the other operator names in a followup.

quint/apalache-tests.md
shonfeder: I’m open to rethinking the approach here ofc. If you foresee problems.

but, as noted in my PRs, my current thought is that we move tests that the examples Work or not per se into the examples dashboard, and only use the integration tests for checking specific functionality.

Issue #690, opened 2023-03-09
The scope of constraints and substitutions in the type inferrer is arbitrary


bugarela:
Just like with effects in #669, we also apply type substitutions to all previously inferred types, which doesn't scale. This is not impacting performance too much on our examples right now, but it is a big potential issue.
 
https://github.com/informalsystems/quint/blob/edeadf4357b5856ca6c6056d7eab7d3993af3e4d/quint/src/types/constraintGenerator.ts#L237-L240

bugarela:
The quantification fix I'm working on will fix this issue 🎉

PR #970, opened 2023-06-20
Clear effect substitutions when exiting toplevel opdefs

bugarela:
Hello :octocat: 

Ok, I really don't know if I should be happy or sad about this. I started profiling the analysis of `ics20` with the intention to confirm the part of the type checker that had the slow down (my suspicion was #690), and I was surprised to see the problem was still the effect checker. 

With some more profiling, I found that the biggest offender was substitution composition, and then I tried cleaning up substitutions after each top-level def and, surprisingly, that improves performance by ~10x while keeping all the extensive test list green.

So yeah, I'm sorry I didn't look into this earlier. I really thought it was something way nastier to fix.

Before:
```
$ time quint typecheck ../examples/cosmos/ics20/ics20.qnt

________________________________________________________
Executed in    9.65 secs    fish           external
   usr time   12.02 secs    2.39 millis   12.02 secs
   sys time    0.23 secs    1.05 millis    0.22 secs
```

After:
```
$ time quint typecheck ../examples/cosmos/ics20/ics20.qnt

________________________________________________________
Executed in    1.68 secs    fish           external
   usr time    3.00 secs    2.50 millis    3.00 secs
   sys time    0.11 secs    0.93 millis    0.11 secs
```


konnov (approved):
Oh, wow! This is a great finding! A 10x improvement in 14 lines of code is great :)

PR #971, opened 2023-06-20
fix slice on the edge

konnov:
This PR fixes the behavior of `slice` for the edge case of `l.slice(length(l), length(l))`. The current implementation flags an error, which is inconvenient. The fix returns an empty list.


bugarela (approved):
Great fix :smile:

Issue #976, opened 2023-06-23
Type aliases should be inlined


bugarela:
We should have a procedure to inline type aliases, so external programs don't have to worry about it. This is a problem now for the integration with Apalache.

PR #979, opened 2023-06-23
Inline type aliases in flattening modules

### LINKED ISSUE #176: Constructing structure-preserving values from IR Visitor walks is hard
Discussed in https://github.com/informalsystems/tnt/pull/154/files/b3e16be92e5b3dabcb12fd621f035de4c4fba735#r957645771

Our IR Visitor methods return void, so all information constructed in a walk has to be stored in class attributes. A common use case for the IR is to transform it into other objects with the same structure (i.e. converting components to strings; adding types to components), but doing that with a visitor requires maintaining component stacks and composing them on each node.

Return values for different visitor implementations have different types, and a generalization can be too complicated. I think a good idea would be to allow two types of return values: strings, and the component itself.

bugarela:
Hello :octocat: 

This is step one of two towards #976.

I've decided to finally close #176 by implementing an `IRTransformer`. Is not as generic as it could be, as it transforms components into the same type of component, so i.e. can't be used for pretty printing (which transforms components into strings). I think this was a good decision since it simplifies the important part, `aliasInliner`, by a lot, delegating the complexity of recursing over the IR to a separate simple module.

This PR only inlines aliases on the actual module and the lookup table. We still have to inline aliases in the type map. That should be pretty straight forward now that we already have the transformer, but I'll do it in a follow up PR since this is already quite big.

I chose to introduce `IRTransformer` and `aliasInliner` in the same PR since the latter is a great usage example of the former, so I thought it would help your understanding during review.


bugarela:
At first I was removing the typedefs from the module while inlining, but that caused problems with the incremental REPL. If the user wants to add a new def like `val a: MY_ALIAS = 1` and `MY_ALIAS` no longer exists in the module, name resolution fails. 

We can just keep ignoring this defs in Apalache. If you think there's a problem with not removing them, I can think about keeping an alternative data structure for the incremental REPL - that just feels like too much overhead right now.

quint/src/flattening.ts
```diff
@@ -459,6 +468,22 @@ class Flatenner {
...
+function inlineTypeAliases(modules: QuintModule[], table: LookupTable): { modules: QuintModule[]; table: LookupTable } {
+  const modulesWithInlinedAliases = modules.map(m => inlineAliases(table, m))
+  const tableWithInlinedAliases = new Map(
+    [...table.entries()].map(([id, def]) => {
+      if (!def.typeAnnotation) {
+        return [id, def]
+      }
+
+      const inlinedType = inlineAliasesInType(table, def.typeAnnotation)
+      return [id, { ...def, typeAnnotation: inlinedType }]
```
thpani, line 480: This constructs a fresh `AliasInliner` on each table entry.

Would it make more sense to refactor this code into a function in `aliasInliner.ts`?
That would give better symmetry with the call for modules.

quint/src/types/aliasInliner.ts
```diff
@@ -0,0 +1,87 @@
...
+ * the provided LookupTable. Uninterpreted types are left unchanged.
+ *
+ * @param lookupTable - The LookupTable containing the type aliases to be
+ * resolved.
+ * @param quintModule - The QuintModule to transform.
+ *
+ * @returns The transformed QuintModule with all type aliases replaced with
+ * their resolved types.
+ */
+export function inlineAliases(lookupTable: LookupTable, quintModule: QuintModule): QuintModule {
```
thpani, line 32: Purely stylistic, but I would flip the order of arguments here, given that the "main" argument is the module and the lookup table is mostly a helper.

```suggestion
export function inlineAliases(quintModule: QuintModule, lookupTable: LookupTable): QuintModule {
```

quint/src/types/aliasInliner.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Inlines type aliases in a QuintType using the provided LookupTable.
+ *
+ * @param lookupTable - The LookupTable containing the type aliases to be
+ * resolved.
+ * @param type - The QuintType to transform.
+ *
+ * @returns The transformed QuintType with all type aliases replaced with
+ * their resolved types.
+ */
+export function inlineAliasesInType(lookupTable: LookupTable, type: QuintType): QuintType {
```
thpani, line 47: sim

```suggestion
export function inlineAliasesInType(type: QuintType, lookupTable: LookupTable): QuintType {
```

quint/src/types/aliasInliner.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Inlines type aliases in a QuintDef using the provided LookupTable.
+ *
+ * @param lookupTable - The LookupTable containing the type aliases to be
+ * resolved.
+ * @param def - The QuintDef to transform.
+ *
+ * @returns The transformed QuintDef with all type aliases replaced with
+ * their resolved types.
+ */
+export function inlineAliasesInDef(lookupTable: LookupTable, def: QuintDef): QuintDef {
```
thpani, line 62: sim
```suggestion
export function inlineAliasesInDef(def: QuintDef, lookupTable: LookupTable): QuintDef {
```

thpani (approved):
Looks great!

I only have a few stylistic remarks.

quint/src/types/aliasInliner.ts
bugarela: I think it makes sense. I was sort of trying to use this pattern that comes from languages with currying, where the most "fixed" arguments appear first and the most varying, later. But to be honest, it's kind of a mess right now where some parts use that, and some parts use most-important-first.

I guess most-important-first makes more sense in a language without currying, so let's go with that :)

PR #993, opened 2023-06-28
Fix trace outputs in REPL

### LINKED ISSUE #982: Display the evaluation stack for nullary operators
This unexpected behavior appeared during the first Quint workshop. In the session below, execution stack is not shown when a nullary operator `step` is evaluated, in contrast to unary operators `inc` and `dec`:

```bluespec
$ quint   
Quint REPL 0.11.3
>>> var x: int

>>> action inc(i) = x' = x + i

>>> action step = inc(2)

>>> .verbosity=4
.verbosity=4
>>> x' = 0
true

[Frame 0]
q::input() => true

>>> step
true

[Frame 0]
q::input() => true

>>> inc(5)
true

[Frame 0]
q::input() => true
└─ inc(5) => true

>>> x
7

[Frame 0]
q::input() => 7

>>> action dec(i) = all { x - i > 0, x' = x - i }

>>> dec(3)
true

[Frame 0]
q::input() => true
└─ dec(3) => true

>>> dec(10)
false

[Frame 0]
q::input() => false
└─ dec(10) => false

>>>
```

### LINKED ISSUE #983: REPL breaks after evaluating the name of a `def`
In the following example, after we invoke `f`, we can no longer evaluate `f` with arguments.

```
Quint REPL 0.11.3
>>> pure def f(x) = x + 1

>>> f()
static analysis error: error: [QNT000] Expected 1 arguments, got 0
Trying to unify (int) => int and () => t0

f()
^^^

static analysis error: error: [QNT000] Expected 1 arguments, got 0
Trying to unify (Read[v0] & Temporal[v1]) => Read[v0] & Temporal[v1] and () => e0
Trying to infer effect for operator application in f()

f()
^^^


{(_a0) => ...}
>>> f(1)
2
>>> f
{(_a0) => ...}
>>> f(1)
{(_a0) => ...}
```

konnov:
Closes #982. Closes #983. This PR refactors REPL, in order to fix a series of bugs that surfaced in the detailed trace output.

Core changes:
 - refactor `ReplState` to be a class, as we have non-trivial class invariants there
 - introduce `TraceRecorderImpl` to make `rng` and `verbosityLevel` mutable
 - introduce `listener` in `EvaluationState`, as the compiled definitions must use the same listener/recorder
 - fix the mutation of `eval` in `CompilerVisitor.exitOpDef`, see the comment in the diff


quint/src/runtime/trace.ts
```diff
@@ -140,153 +140,200 @@ export const noExecutionListener: ExecutionListener = {
...
+  private bestTraceSeed: bigint
   // whenever a run is entered, we store its seed here
-  let runSeed = bestTraceSeed
+  private runSeed: bigint
   // during simulation, a trace is built here
-  let frameStack: ExecutionFrame[] = [bestTrace]
-
-  return {
-    getBestTrace: (): ExecutionFrame => {
-      return bestTrace
```
konnov, line 175: Most of the changes below are just about moving the code in the class definition

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -237,14 +241,22 @@ export class CompilerVisitor implements IRVisitor {
...
-        return r
+      boundValue = {
+        eval: () => {
+          this.execListener.onUserOperatorCall(app)
+          const r = unwrappedValue.eval()
+          this.execListener.onUserOperatorReturn(app, [], r)
+          return r
+        },
       }
     }
```
konnov, line 261: This is the final fix for #982. We should not have mutated `boundValue.eval`.

quint/src/repl.ts
```diff
@@ -516,6 +558,14 @@ function tryEvalHistory(out: writer, state: ReplState): boolean {
   return true
 }
 
+// Try to evaluate the expression in a string and print it, if successful.
+// After that, clear the recorded stack.
+function tryEvalAndClear(out: writer, state: ReplState, newInput: string): boolean {
```
bugarela, line 563: From the name, it is unclear that only the recorder is cleared after evaluation. Can we try to come up with a better name here? I thought of:

```suggestion
function tryEvalWithoutRecording(out: writer, state: ReplState, newInput: string): boolean {
```

quint/src/runtime/compile.ts
```diff
@@ -82,15 +82,17 @@ export function newCompilationState(): CompilationState {
...
+  return (errs: ErrorMessage[]) => {
+    return {
+      lookupTable: new Map(),
+      syntaxErrors: errs,
+      analysisErrors: [],
+      compileErrors: [],
+      getRuntimeErrors: () => [],
+      compilationState: newCompilationState(),
+      evaluationState: newEvaluationState(listener),
+    }
```
bugarela, line 95: Idk if we want to instantiate a new empty compilation and evaluation state for every error message. I believe we could do something like this:
```suggestion
export function errorContextFromMessage(listener: ExecutionListener): (errors: ErrorMessage[]) => CompilationContext {
  const compilationState = newCompilationState()
  const evaluationState = newEvaluationState(listener)
  return (errs: ErrorMessage[]) => {
    return {
      lookupTable: new Map(),
      syntaxErrors: errs,
      analysisErrors: [],
      compileErrors: [],
      getRuntimeErrors: () => [],
      compilationState,
      evaluationState,
    }
```

bugarela (approved):
LGTM! Awesome work on these fixes :rocket: Just a couple of comments

PR #999, opened 2023-06-30
Remove unused state var in `clockSync3.qnt`

thpani:
Remove the unused state variable `adj` from `clockSync3.qnt`.
This corresponds to unimplemented functionality in the [TLA+ template](https://github.com/informalsystems/tla-apalache-workshop/blob/main/examples/clock-sync/ClockSync3.tla) where this was translated from.

The variable was declared and initialized, but overlooked in the other actions.

@bugarela Shouldn't the effects checker notice a missing assignment?
It was only flagged in Apalache's transition finder.

### DIFF
--- a/examples/classic/distributed/ClockSync/clockSync3.qnt
+++ b/examples/classic/distributed/ClockSync/clockSync3.qnt
@@ -46,9 +46,6 @@ module clockSync3Spec {
     /// Hardware clock of a process
     var hc: PROC -> int
 
-    /// Clock adjustment of a process
-    var adj: PROC -> int
-
     /// Messages sent by the processes
     var msgs: MSGS
 
@@ -76,7 +73,6 @@ module clockSync3Spec {
         { nondet hc0 = oneOf(setOfMaps(Procs, Nat))
           hc' = hc0 },
         msgs'  = Set(),
-        adj'   = Procs.mapBy(_ => 0),
         state' = Procs.mapBy(_ => "init"),
         rcvd'  = Procs.mapBy(_ => Set()),
     }


konnov:
There must be something wrong in the spec. If we are not using adjustments, the clock sync algorithm does not make a lot of sense. We should compare it with the TLA+ version

thpani:
> There must be something wrong in the spec. If we are not using adjustments, the clock sync algorithm does not make a lot of sense. We should compare it with the TLA+ version

Ah, this is clockSync3, I assume the full spec is clockSync7.
Shall we file an issue to write the entire spec in Quint?

konnov:
> Ah, this is clockSync3, I assume the full spec is clockSync7.
> Shall we file an issue to write the entire spec in Quint?

Oh yes, that would be great. I thought we would use it for a demo today, but apparently not :D

konnov (approved):
Good catch

bugarela:
> @bugarela Shouldn't the effects checker notice a missing assignment?
It was only flagged in Apalache's transition finder.

The effect checker doesn't know that `init` and `step` should be used together. They are evaluated as any other action, so `init` updates `adj`, `step` doesn't, and since they are not ever combined in the spec, there is no issue. You'd probably see an issue if you declare something like `run a = init.then(step)`.

I've raised this before but I can't find any issue on this. My proposal is that `init` and `step` are required to have their effect updating all state variables. Currently there is no check for unused state variables as well.

thpani:
> The effect checker doesn't know that `init` and `step` should be used together. They are evaluated as any other action, so `init` updates `adj`, `step` doesn't, and since they are not ever combined in the spec, there is no issue. You'd probably see an issue if you declare something like `run a = init.then(step)`.

Aah good point, that makes sense!

shonfeder:
> I've raised this before but I can't find any issue on this. My proposal is that init and step are required to have their effect updating all state variables. Currently there is no check for unused state variables as well.

This came up in a workshop too. Rather than having special treatment of operator named `init` or `step`, it might be more flexible and consistent with our design to have a keyword modifier like:

```
complete action foo = ...
```

Or

```
transition foo ...
```

or something, that acts as an effect annotation stating all regions are meant to be updated by the action.

Issue #994, opened 2023-06-28
Improve performance of name resolution


bugarela:
After profiling a large spec by Manu, I found that our biggest offender is now name resolution (taking about 75% of the total time), specially conflict scanning. We call the name resolver multiple times while flattening, with bigger and bigger modules, and that is not scaling well. 

Using incremental name resolution (#934) in flattening might be one approach, but there might be structural improvements in name resolution itself.

PR #1001, opened 2023-06-30
Cache results of `scopesForId`

bugarela:
Hello :octocat: 

Towards #994 

This is a small addition that improves performance significantly. The problem here is a bit more structural in the name resolution step, which also doesn't have a good interface to enable incremental resolution at the moment.

I will address the structural problems (probably next week), but I decided to propose this small change right now as I had already implemented it and the performance improvement can be useful for our users (specially Manu, whose spec was taking ~60s to load and now takes ~25s).


shonfeder (approved):
Awesome precision improvement!

PR #1003, opened 2023-07-03
Small improvements/fixes to the examples tooling

thpani:
Two small improvements/fixes to the examples tooling:
- Make target for `examples/README.md` phony, so that it can be run even when the example sources don't change (e.g., to test Apalache modifications)
- Use `APALACHE_DIST` in `run-examples.sh` if already set, to allow reusing a local Apalache repo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 # @file
 # @version 0.1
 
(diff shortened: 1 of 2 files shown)


konnov (approved):
LGTM

PR #1007, opened 2023-07-04
Add instantiating module to `LamportMutex`

thpani:
Instantiate parameterized module `LamportMutex` to let it go through `quint test` and `quint verify`.

I chose the same parameters as in Apalache tests: https://github.com/informalsystems/apalache/blob/191bca02b14572e5b13f3dbdc8b490d4d89be50b/test/tla/MC_LamportMutexTyped.tla

This also updates the example runner to consider (hardcoded) `--main`/`--init`/`--step` that differ from the file name. This is a quick fix for now, I will revisit this with a better solution when we have a majority of the examples working.

### DIFF
--- a/examples/classic/distributed/LamportMutex/LamportMutex.qnt
+++ b/examples/classic/distributed/LamportMutex/LamportMutex.qnt
@@ -175,3 +175,10 @@ module LamportMutex {
   // The main safety property of mutual exclusion.
   val Mutex = crit.forall(p => crit.forall(q => p == q))
 }
+
+module LamportMutex_3_10 {
+  import LamportMutex(
+    N = 3,
+    maxClock = 10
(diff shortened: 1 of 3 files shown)


Kukovec (approved):

examples/classic/distributed/LamportMutex/LamportMutex.qnt
```diff
@@ -175,3 +175,10 @@ module LamportMutex {
   // The main safety property of mutual exclusion.
   val Mutex = crit.forall(p => crit.forall(q => p == q))
 }
+
+module LamportMutex_3_10 {
```
konnov, line 179: I like that!

konnov (approved):
LGTM!

PR #1008, opened 2023-07-04
Implement topological sort of definitions

### LINKED ISSUE #129: Topological sort of definitions in phase 2
The language manual lets the user to write definitions out of order. For instance:

```scala
module OutOfOrder {
  val n1 = e1
  val n2 = n3 + 1
  val n3 = n1 * 2
}
```

Currently, the parser (phase 1) and the name resolver (phase 2) keep the order of definitions. Moreover, the name resolver does not check for cyclic definitions.

As a result, when writing a later phase (e.g., the simulator) with `IRVisitor`, I have to think about the out-of-order definitions. It looks like we have enough information to make a topological sort of definitions in phase 2. If we do so, using `IRVisitor` in later phases would be simpler, as we can assume that all names are defined before they are used (in the traversal order of `IRVisitor`).

Apalache is doing a topological sort of definitions already, perhaps not optimally, so we can get some inspiration from [DeclarationSorter.scala](https://github.com/informalsystems/apalache/blob/unstable/tlair/src/main/scala/at/forsyte/apalache/tla/lir/transformations/standard/DeclarationSorter.scala).

Once we have topological sort, we can detect cyclic references in phase 2.

konnov:
Closes #129. This PR implements topological sort of definitions inside modules, to large extent. This feature happened to be much harder to implement than I anticipated (1 week instead of 1-2 days). The big difference between sorting Quint definitions and Apalache definitions is that Apalache assumes all modules to be flattened at the import stage, which is not the case with Quint.

The main contributions are:

- [x] `callgraph.ts` implements the general computation of a call graph/use graph. This class may be used to easily find forward and backward references in the future, e.g., in the VSCode plugin.
- [x] `toposort.ts` implements topological sort of an array of definitions, given the call graph
- [x] phase 4 in `quintParserFrontend.ts`

Given all that, this feature seems to work. For example, the following source code is parsed without errors:

```bluespec
module A {                                                                
  pure def sqr(x) = x * x
}

module B {
  const M: int
  pure val doubleM = M + M
}

module C {
  val n2 = n3 + 1
  val n3 = n1 * 2
  val n1 = 50
}

module main {
  pure val myM = sqr(3)
  pure val quadM = 2 * B1::doubleM
  import A.*
  import B(M = myM) as B1
  export B1.*
}
```

However, sometimes name resolution (phase 3) fails on out of order definitions. Hence, topological sort does not have a chance to fix things. Still, it is an improvement, as name resolution is failing with readable errors.

Also, sorting modules is postponed for #1009.

One thing to keep in mind during the review is whether it would correctly interact with REPL. Tagging @bugarela here.


quint/src/flattening.ts
```diff
@@ -241,6 +241,10 @@ class Flatenner {
     const qualifiedName = def.defName ? undefined : def.qualifiedName ?? def.protoName
 
     const protoModule = this.importedModules.get(def.protoName)!
+    if (!protoModule) {
```
thpani, line 244: When could you enter the `if` here?
You already have a non-null assertion on `get`?

```suggestion
    const protoModule = this.importedModules.get(def.protoName)!
    if (!protoModule) {
```

quint/src/flattening.ts
```diff
@@ -241,6 +241,10 @@ class Flatenner {
...
+      // Something got went really wrong
+      console.error(`Imported module ${def.protoName} not found. Definitions out of order?`)
```
thpani, line 246: I thought this PR adds support for out-of-order defs?

quint/src/static/callgraph.ts
```diff
@@ -0,0 +1,247 @@
...
+import type { RecordOf } from 'immutable'
+
+import { IRVisitor } from '../IRVisitor'
+import { LookupTable } from '../names/lookupTable'
+import { QuintApp, QuintDef, QuintExport, QuintImport, QuintInstance, QuintModule, QuintName } from '../quintIr'
+import { QuintConstType } from '../quintTypes'
+
+/**
+ * The call graph is simply a mapping from the caller's id
+ * to the set of callee's ids.
```
thpani, line 22:
```suggestion
 * to the set of callees' ids.
```

quint/src/static/callgraph.ts
```diff
@@ -0,0 +1,247 @@
...
+  /**
+   * The map from a module name to the corresponding module.
+   */
+  modulesByName: Map<string, QuintModule>
+
+  /**
+   * For every module id A and the imported module name B, keep the set of
+   * import statements that refer to B in A.
+   */
+  importsByName: Map<ImportKeyRecordType, Set<bigint>>
```
thpani, line 54: iiuc, the `bigint` is the id of the import statement?
Maybe we can add this in the comment.

quint/src/static/callgraph.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  get graph() {
+    return this._graph
+  }
+
+  /**
+   * Print the graph in the graphviz dot format. Use it for debugging purposes,
+   * e.g., print(console.log)
+   */
+  print(out: (s: string) => void) {
```
thpani, line 138: wow, that's great!

quint/test/static/callgraph.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    return d
+  }
+
+  function findExport(module: QuintModule, pred: (imp: QuintExport) => boolean): QuintDef {
+    const d = module.defs.find(d => d.kind === 'export' && pred(d))
+    assert(d, `Export not found in ${module.name}`)
+    return d
+  }
+
+  it('const, var, and operator definitions', () => {
```
thpani, line 66: I think the BDD idea is to follow `it` with a verb 😉 

```suggestion
  it('computes a call graph of const, var, and operator definitions', () => {
```

quint/test/static/callgraph.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    const w = findDef(main, 'w')
+    const N = findDef(main, 'N')
+    const getW = findDef(main, 'getW')
+    expect(graph.get(plus.id)?.size).to.equal(2)
+    expect(graph.get(double.id)?.toArray()).to.include.members([plus.id])
+    expect(graph.get(triple.id)?.toArray()).to.include.members([plus.id, double.id])
+    expect(graph.get(getW.id)?.toArray()).to.include.members([w.id, N.id])
+  })
+
+  it('typedefs', () => {
```
thpani, line 95:
```suggestion
  it('computes a "uses" graph of typedefs', () => {
```

quint/test/static/callgraph.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    const graph = visitor.graph
+    const bagOfApples = findDef(main, 'BagOfApples')
+    const boxOfApples = findDef(main, 'BoxOfApples')
+    const x = findDef(main, 'x')
+    expect(graph.get(bagOfApples.id)).to.be.undefined
+    expect(graph.get(boxOfApples.id)?.toArray()).to.eql([bagOfApples.id])
+    expect(graph.get(x.id)?.toArray()).to.eql([boxOfApples.id])
+  })
+
+  it('imports and defs', () => {
```
thpani, line 117:
```suggestion
  it('computes a "uses" graph of imports and defs', () => {
```

thpani (approved):
Looks great!
A large changeset, but it was a pleasure to read! 🙏 🎉

quint/src/flattening.ts
konnov: oh, that's a good catch! I actually wanted to remove the `!` operator, as it makes the code crash with an unreadable stack trace.

quint/src/flattening.ts
konnov: It's a guard rail. I have added it before toposort was working, as it was very hard to understand what was going on.

konnov: It totally does. I have added an explanatory exception there, just in case the parse phase 4 is missing somewhere.

Issue #539, opened 2023-01-19
Add support for discriminated unions/sum types


konnov:
We have been envisioning redesign of discriminated unions, which we have not really implemented yet.

Related issues: #233, #244, #440

In all of our discussions, we seem to have converged on the idea of having (non-inductive, non-recursive) ADTs instead of discriminated unions. We have to think about the syntax, but following other languages, it should look similar to the ones below, modulo renaming of tokens:

```
// similar to Result in Rust
type Result[T, E] = Ok(T) | Err(E)
// similar to Either in Scala
type Either[L, R] = Left(L) | Right(R)
```

What we have learned from Apalache is that discriminated unions almost immediately require constructors for the options such as `Ok` and `Err` in the above code, see [an example]( https://github.com/informalsystems/apalache/blob/8baab28cf4d46b02a9d2f5b8ebceebfa2b972884/src/tla/Option.tla#L13-L32). Otherwise, they stay open and are not very useful. The good thing about the ADT syntax is that it also introduces constructors for the options.

As a first step, we would definitely avoid recursive ADTs. It requires a kind of reasoning that is different from what is normally done in TLA+. Moreover, there is no practical need for recursive ADTs yet. We may revisit this decision in the future.

shonfeder:
Afaiu, the record-based discriminated unions are a totally workable basis for a sum type that can be with use with other ADTs (i.e. product types), but I agree that the tagged constructor/deconstruct notation is nicer than having to go through records with a special field :)

shonfeder:
Seems to be a duplicate of #244

PR #1010, opened 2023-07-05
Update Paxos example

thpani:
Update `Paxos` example:
* Replace the inline mock `Voting` module with a proper file import
* Add an instance for model checking (same parameters as in Apalache tests)

(This still doesn't pass through Apalache, because the spec uses discriminated unions, which [aren't implemented as of now](#539))


bugarela:
@thpani seems like you forgot to commit `Voting.qnt`

bugarela:
I think we also need to add a `from "./Voting.qnt"` to the existing import.

thpani:
> I think we also need to add a `from "./Voting.qnt"` to the existing import.

Aaah, forgot to add a line to the commit. PTAL

bugarela (approved):

Issue #452, opened 2022-12-20
Use type signatures from `builtin.qnt`


bugarela:
We currently have a duplication of information in `builtinSignatures.ts` and `builtin.qnt`. We could get rid of the signature lists in `builtinSignatures.ts` and instead parse them from `builtin.qnt`

Issue #451, opened 2022-12-20
We can't write proper type signatures for some operators


bugarela:
Our current type syntax doesn't allow proper representation of:

1. Variadic operators
2. Record operators
3. Tuple operators

As @Kukovec pointed out, for some of them, there's not really a way to write them:
- https://github.com/informalsystems/tnt/pull/386#discussion_r1051057817
- https://github.com/informalsystems/tnt/pull/386#discussion_r1051052713

Perhaps we would need to include constraints and even new levels of quantification to proper express that.

I believe we should compromise here and use overly generic type signatures for records and tuples (as @Kukovec suggested)

- [ ] Decide on a syntax for Variadic operators and update the parser
- [ ] Decide on a approach for record/tuple operators
- [ ] Update `builtin.tnt`

shonfeder:
This has been discussed at some length on #1062 and in passing in #1081. I believe there are two distinct cases:

1. Variadic operators that can't be typed just because we don't have a variadic syntax.
2. Exotic operator that can't be typed in our type system because their type depends on the values they are given, and we don't want types that depend on values in our system.

`List` is an example of (1). `field` is an example of (2). `Rec` combines both.

PR #1011, opened 2023-07-05
Name resolution restructure for performance boost

### LINKED ISSUE #994: Improve performance of name resolution

bugarela:
Hello :octocat: 

This PR collapses 5 stages of name resolution into 2, getting rid of A LOT of complexity (code-wise and performance-wise) now that I have a much better understanding of what we need to resolve names. This should also enable incremental name resolution (#934) with much more ease.

Error reporting is much better managed as well, as we now have a single interface that always returns an already processed `QuintError`. See the difference in `quintParserFrontend.ts`. I also changed some error messages to be more straight to the point.

Definitions collection and import resolution are now done in a single step in `collector.ts`, while name resolution and scope handling are both done in `resolver.ts`. Detecting conflicts is done on the fly instead of in an additional (very expensive) pass. Data structures were also simplified, and there is no such thing as a `ScopeTree` anymore :pray:.

:warning: The diff is huge, but I wouldn't know how to break this down in a feasible way. I'd recommend just reading the added code in `collector.ts` then `resolver.ts`, as everything else  is adaptation to new data structures. The tests can be hard to review, but they are simply an adaptation of the existing tests + a few others that I added to reach 100% coverage (from `npm run coverage`).

Close #994 


konnov (commented):
I have left a few comments. So far looks amazing! I have to continue tomorrow.

quint/src/names/base.ts
```diff
@@ -0,0 +1,184 @@
...
+  'field',
+  'fieldNames',
+  'item',
+  'unionMatch',
+  'assign',
+  'of',
+  'eq',
+  'neq',
+  'ite',
+  'cross',
```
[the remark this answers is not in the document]
bugarela: The plan is to have that all under `builtin.qnt`, see https://github.com/informalsystems/quint/issues/452. However, that is blocked because some operators don't have parseable signatures (see https://github.com/informalsystems/quint/issues/451).

quint/src/names/base.ts
```diff
@@ -0,0 +1,184 @@
...
+export type DefinitionsByModule = Map<string, DefinitionsByName>
+
+/**
+ * A lookup table from IR component ids to their definitions.
+ * Should have an entry for every IR component with a name
+ * That is:
+ * - Name expressions
+ * - App expressions (opcode is a name)
+ * - Override parameters in instance definitions
+ * - Constant types (which are references to type aliases or uninterpreted types)
```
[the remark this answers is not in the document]
bugarela: `DefinitionKind` is the codomain, while this list talks about the domain. Makes sense?

quint/src/names/base.ts
```diff
@@ -0,0 +1,188 @@
...
+ */
+export type DefinitionKind = 'module' | 'def' | 'val' | 'assumption' | 'param' | 'var' | 'const' | 'type'
+
+/**
+ * A minimal representation of a QuintDef, to be stored in `LookupTable` and
+ * `DefinitionsByName`.
+ */
+export interface Definition {
+  kind: DefinitionKind
+  reference: bigint
```
konnov, line 28:
```suggestion
  originId: bigint
```

I was always puzzled by this name. The problem with references is that they may be transitive. I think we are always pointing to the original definition, right?

quint/src/names/collector.ts
```diff
@@ -0,0 +1,287 @@
...
+    if (def.protoName === this.currentModuleName) {
+      // Importing current module
+      this.errors.push(selfReferenceError(def))
+      return
+    }
+
+    const moduleTable = this.definitionsByModule.get(def.protoName)
+
+    if (!moduleTable) {
+      // Importing unexisting module
```
konnov, line 147:
```suggestion
      // Importing non-existing module
```

quint/src/names/collector.ts
```diff
@@ -0,0 +1,287 @@
...
+  enterExport(def: QuintExport) {
+    if (def.protoName === this.currentModuleName) {
+      // Exporting current module
+      this.errors.push(selfReferenceError(def))
+      return
+    }
+
+    const moduleTable = this.definitionsByModule.get(def.protoName)
+    if (!moduleTable) {
+      // Exporting unexisting module
```
konnov, line 184:
```suggestion
      // Exporting non-existing module
```

quint/src/parsing/quintParserFrontend.ts
```diff
@@ -287,75 +292,9 @@ export function parsePhase2sourceResolution(
...
-      }
-    )
-
-    return mergeInMany([importResult, conflictResult, resolutionResult])
-      .chain(([_, __, table]) => result.map(t => new Map<bigint, Definition>([...t.entries(), ...table.entries()])))
-      .mapLeft(errors => errors.flat())
-  }, right(new Map()))
-
-  return definitions.map(table => ({ ...phase2Data, table }))
+  return resolveNames(phase2Data.modules)
```
konnov, line 295: I like this new interface a lot :)

quint/testFixture/_1006unexpectedExpr.json
```diff
@@ -1 +1 @@
-{"stage":"parsing","warnings":[],"errors":[{"explanation":"Failed to resolve name name in definition for access, in module unexpectedExpr","locs":[{"source":"mocked_path/testFixture/_1006unexpectedExpr.qnt","start":{"line":2,"col":21,"index":84},"end":{"line":2,"col":24,"index":87}}]}]}
\ No newline at end of file
+{"stage":"parsing","warnings":[],"errors":[{"explanation":"[QNT404] Name 'name' not found","locs":[{"source":"mocked_path/testFixture/_1006unexpectedExpr.qnt","start":{"line":2,"col":21,"index":84},"end":{"line":2,"col":24,"index":87}}]}]}
```
konnov, line 1: Shall we automatically exclude these json files from reviews, similar to how we have done it in the tutorials, see `tutorials/.gitattributes`

konnov (approved):
This is a great refactoring! The code is much cleaner now.

quint/testFixture/_1006unexpectedExpr.json
bugarela: I like having them in the diff so the behavior changes are somewhat explicit in the review, i.e. here you can see a concrete error message diff. Of course, when it's just ids moving around, it's noise. But most importantly, this is a sanity check that we didn't just automatically update the fixtures to error outputs and our unit tests are passing only because the expectation is for things to error out.

quint/src/names/base.ts
bugarela: This got me thinking (because `originId` seems weird to me here), and actually I think this should be just `id`. I'm sort of using `id` as primary keys an `reference` as foreign keys across the code base, but I think this `Definition` is just a projection of some fields from a `QuintDef`, so it's not like the `QuintDef` is the `origin` here, they are the same thing, only with different representations.

I'll change this to `id`.

quint/testFixture/_1006unexpectedExpr.json
bugarela: Oh damn, that actually breaks integration with Apalache because this is part of our JSON output. I'll roll back the change to get this merged and we can change the name in a new PR   in sync with an Apalache release.

quint/src/names/base.ts
konnov: `id` may be confusing, as we are typically using `id` to denote the unique identifier of the object, that is, the id of `Definition` in this case.

bugarela: whops, this comment was meant to be on this thread: https://github.com/informalsystems/quint/pull/1011#discussion_r1254368644

quint/testFixture/_1006unexpectedExpr.json
bugarela: (wrong thread, this comment was about the id vs reference)

quint/src/names/base.ts
thpani: I think this would be a single-line change in Apalache:

https://github.com/informalsystems/apalache/blob/f6bfe31271b38f07b77600525356f3ba4455219d/tla-io/src/main/scala/at/forsyte/apalache/io/quint/QuintIR.scala#L59

Issue #693, opened 2023-03-09
Nested definition quantification allows arbitrary types


bugarela:
Igor reported a bug where he expected I type error but didn't get any. By trying to minimize the example, I found that the bug is pretty big: we can use nested definitions to completely ignore the type annotation.

```
module foo {
  def f(x: int): bool = {
    val bar = x
    bar
  }
}
```

I'm almost sure this is due to me not being careful with non free variables, and the solution should be quite similar to the one I just implemented on the effect system. 

Note: It's impressive how consistent I am in making mistakes on both the type and the effect system. I actually thought this would be a problem in the type system as well and was surprised that it wasn't showing any issues so far. I should probably be more pro-active in fixing something on the type system when I find it on the effect system and vice-versa.

shonfeder:
Do you think using an approach like ABTs or HOAS to handle variable binding and scope in a systematic and rigorous way could help us reduce the error surface for this sort of thing?

shonfeder:
Well, the gist of my suggestion is that, e.g., the point of ABTs is to ensure a systematic treatment of name binding, that is easy to reason about and rules out the possibility of confusing bound and free variables. So my understanding is that this should exactly help deal with the problem of

> I'm just quantifying all type variables instead of only non-free variables

since it looks like you currently are having to compute this by two different sets of heuristics, and since by my recollection we seem to be hitting some variety of subtle problems around naming (I’m considering also the issues with substation, that also seems to involve scope iiuc?), it makes me wonder if we might solve this class problems across both the effect and type checking, and perhaps prevent other similar problems, by borrowing from these techniques for reasoning about freedom and bondage (of variables? :)


That said, I’ve not had a chance to study the issue here close enough to be certain it would do the trick. I can spend some cycles trying to evaluate tho if that might be helpful.

shonfeder:
Tho, even if that would help, I guess it’s good to keep in mind the costs of an AST/IR change at this point :)

bugarela:
Another case of this reported by @konnov:
```
  pure def listSorted(__list: List[a], __lt: (a, a) => bool): List[a] = {
    pure def __countSmaller(__j: int): int = {
      pure val __jth = __list[__j]
      __list.indices().filter(__i =>
        __lt(__list[__i], __jth) or (__list[__i] == __jth and __i < __j)
      )
      .size()
    }

    pure val __permutation = __list.indices().mapBy(__i => __countSmaller(__i))
[...]

PR #1017, opened 2023-07-06
Update ponzi examples

thpani:
* Explicitly annotate 2 variables to work around #693, to make it pass through Apalache/Snowcat.
* Update example runner script and dashboard.

`gradualPonzi` works after https://github.com/informalsystems/apalache/pull/2635


bugarela (approved):
The integration with Apalache is helping to uncover some type checker errors, that's great!

Kukovec (approved):

PR #1020, opened 2023-07-06
Report error for conflicting module names

### LINKED ISSUE #623: Name resolution should report an error on duplicate module names
It took me a while to debug something like this:

```scala
module A {                                                                      
  val x = 1
}

module A {
  val y = 2
}
```

bugarela:
Hello :octocat: 

Closes #623. Reports a simple error when multiple modules are defined with the same name.

I also had to change source resolution, since it was reversing the order of modules when there was no dependency between them, causing the "already defined" error to point to the first module, not the second one. The trick there is that `reverse()` mutates the array, so I'm now making copies to pass to that method.


konnov (approved):
Nice!

Issue #832, opened 2023-04-18
Update the language manual on modules


konnov:
Now that we have new `import`, `export` and `import... from`, we have to update the manual.

bugarela:
@konnov I think this is a duplicate of https://github.com/informalsystems/quint/issues/530, which is the next item in my queue. I can assign it to you if you want to do it :smile:

shonfeder:
We should include docs on how to use modules to structure specs, and the limits of that usage.

As a special case, we should makes sure to document how module parameters (i.e. "constants") are "inherited" via importing modules.  E.g., 


```ocaml
module m {
  const C:int
}

module n {
  import m.*

  def c = C
}

module o {
  import n(C=42) as n_inst
  def c = n_inst::c
}
```

Gives us

```
$  echo 'c' | quint -r t.qnt::o
Quint REPL 0.18.1
Type ".exit" to exit, or ".help" for more information
>>> c
42
>>> 
```

This was discussed with @p-offtermatt today.

Also relevant is whether this behavior and syntax is really ideal, or if we should find a way to make this much more explicit. See https://github.com/informalsystems/quint/discussions/949 for previous related discussion.

I take fully documenting our current module behavior to be a step towards evaluating whether it is suitable and to thinking carefully about what alternatives we might prefer.

bugarela:
Just a note: that only works if `C` is referenced somehow in `n`. If it is not, we get a flattening error:

```qnt
module m {
  const C:int
}

module n {
  import m.*

}

module o {
  import n(C=42) as n_inst
  def c = n_inst::C
}
```

```
Error: Internal error while flattening [QNT406] Instantiation error: 'C' not found in 'n',[QNT404] Name 'n_inst::C' not found,[QNT404] Name 'n_inst::C' not found
```

PR #1027, opened 2023-07-07
clean the manual

### LINKED ISSUE #530: Update language manual on `Namespaces and Instances` and `Imports`

### LINKED ISSUE #706: Explain C-like type signature in the manual
We've added a new type signature syntax but it is not explained in the language manual, so people might be surprised to see it.

konnov:
Closes #706. Closes #530. I have made a pass over the manual and removed/fixed the outdated material. Not introducing `import ... from` yet, as it requires more careful rewriting. Postponed for #832.

### DIFF
--- a/doc/lang.md
+++ b/doc/lang.md
@@ -2,7 +2,7 @@
 
 | Revision | Date       | Author                                                  |
 |:---------|:-----------|:--------------------------------------------------------|
-| 32       | 19.01.2023 | Igor Konnov, Shon Feder, Jure Kukovec, Gabriela Moreira, Thomas Pani |
+| 33       | 07.07.2023 | Igor Konnov, Shon Feder, Jure Kukovec, Gabriela Moreira, Thomas Pani |
 
 This document presents language constructs in the same order as the [summary of
 TLA+](https://lamport.azurewebsites.net/tla/summary.pdf).
@@ -127,7 +127,7 @@ Multi-line comments:
  */
 ```
 
-<a name="types"/>
+<a name="types"></a>
 
 ## Types
 
@@ -168,7 +168,7 @@ A type is one of the following:
  - Operator: `(T_1, ..., T_n) => R` for `n >= 0` argument types `T_1, ..., T_n`
    and result type `R`.
 
- - Discriminated union:
+ - **Work in progress:** Discriminated union:
     ```
        | { tag: string_1, <ident>: T_1_1, ..., <ident>: T_1_n_1}
        ...
@@ -182,15 +182,13 @@ A type is one of the following:
 It is often convenient to declare a type alias. You can use `type` to define
 an alias inside a module definition. For instance:
 
-```
-type STR_OPTION =
-    | { tag: "none" }
-    | { tag: "some", value: str }
+```bluespec
+type Temperature = int
 ```
 
 A type identifier can also introduce an uninterpreted type by defining a type without any constructors for values of that type:
 
-```
+```bluespec
 type MY_TYPE
 ```
 
@@ -257,7 +255,7 @@ module Foo {
 A single file *should* contain one top-level module, though there is no limit on
 the number of top-level modules that can be included in a file. Modules cannot
 be nested.  To ease module lookup, the name of the top-level module *should*
-match the file name. For example, a file `Orange.tla` should contain `module
+match the file name. For example, a file `Orange.qnt` should contain `module
 Orange { ... }`.  This is not a strict requirement. You have to follow it only,
 if you are using multiple modules in a file.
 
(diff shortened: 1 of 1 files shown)


doc/lang.md
```diff
@@ -495,59 +487,49 @@ module Foo {
 In our example, module `Foo` has four names in its namespace: `N`, `x`,
 `Init`, and `Next`.
 
-A nested module defines its own namespace, which extends the namespace of the
-outer module:
+Quint does not allow for nested modules. However, you can mimick nested
+namespaces by using `::`. For example:
```
thpani, line 491: So if I import a module `Imported`, is it merely mangling the imported names to have an `Imported::` prefix?

doc/lang.md
```diff
@@ -495,59 +487,49 @@ module Foo {
...
-In the above example, the module `Outer` has the following names in its name
-space: `x`, `xPlus`, `inv`, and `Inner`, whereas the module `Inner` contains
-the names `x2` and the names defined in its parent modules: `x`, `xPlus`, and
-`inv`. The module `Inner` is *stateless*, so the module `Outer` may access the
-definitions of `Inner` via the dot-notation, e.g., `Inner.x2`.
-
-**WARNING:** We are redesigning nested modules,
-see [#496](https://github.com/informalsystems/quint/discussions/496). If you are
-starting with the language, wait a bit before using nested modules.
+Note that in the above example, `Inner::x2` is treated as a qualified identifier, that is, it consists of the prefix `Inner` and the name `x2`.
```
thpani, line 506: If that's the case, why not have a `namespace {}` primitive?

doc/lang.md
```diff
@@ -557,16 +539,15 @@ Quint is not Fashion Police.
...
-  }
+```bluespec
+module Math {
+  def sqr(x) = x * x
+  def pow(x, y) = x^y
+}
 
+module Imports {
   // import 'pow' from Math
   import Math.pow
```
thpani, line 553: Bit inconsistent that this uses `.` instead of `::` to qualify the name, no?

doc/lang.md
```diff
@@ -596,27 +576,55 @@ module ImportAll {
 }
 ```
 
-*No export.* It is important to know that `import` does not introduce any
-definitions in the module that uses `import`. Hence, the following example
-produces a lookup error:
+*No re-export by default.* It is important to know that `import` does not automatically introduce *any definitions in the module that uses `import`. Hence, the following example produces a lookup error:
```
thpani, line 580: Is there an extra asterisk at `*any`?

doc/lang.md
```diff
@@ -629,28 +637,36 @@ G(x) == { F(x) }
...
+not hide definitions. However, if you really want to do that, `import ` allows
+you to model this behavior via a proxy module:
 
-```scala
-module Local {
-  var y: int
-  def G(x) = Set(_local.F(x))
+```bluespec
+module A::Impl {
+  pure val secretNum = 123
```
thpani, line 646: Wow, so module names can contain `::`?

thpani (approved):
Looks good overall.

Left a few comments where the current design looks surprising/inconsistent, from a user's perspective.

doc/lang.md
thpani: I meant: why `Math.pow` instead of `Mat::pow`?

doc/lang.md
konnov: Conceptually, it should be like that. I don't think it is implemented like this in the lookup table.

doc/lang.md
konnov: The idea is obviously stolen from C++ `namespace`. We just thought that we maybe don't need the keyword yet, as one can simply add `Inner::` as a prefix. But we may introduce this keyword one day.

doc/lang.md
konnov: This is actually a very good point. @bugarela wdyt? I am actually not even sure about why we would need `::` anymore. It could be properly resolved at the name resolution phase, as in many programming languages. I would assume that `import Math::pow` is a bit unconventional.

konnov:
I will merge the PR for now, as the language manual really needs an update. Let's keep the questions about `.` vs `::` and imports in mind though.

doc/lang.md
shonfeder: I also find this inconsistency jarring (and noted it somewhere). It also shows up in the unqualified import syntax: ‘import A.*`.

I don’t think `import Math::pow` unconventional? I know C++ doesn’t support importing single functions, but Rust, which does, uses `::` consistently as the module accessor: https://doc.rust-lang.org/reference/items/use-declarations.html

And this follows all other languages I know of that use the same `.` syntax for qualified reference and for specific import (e.g., Scala, Python, Ocaml, Haskell, JavaScript).

konnov: Thanks @shonfeder! I keep forgetting about rust. Since there is a precendent in the form of rust, we could of migrate all of our import syntax to `::`. It should be much easier to implement than disambiguation of `.` with a name resolver.

Issue #52, opened 2022-03-24
Uninterpreted type definition and resolution


bugarela:
We are supporting constant uninterpreted types but don't provide a way to declare them. A uninterpreted type declaration could just be a normal type alias declaration but with no assignent/equality:

```
type NAME
```

We would then have to collect this definitions (since constants are already being resolved)

PR #1029, opened 2023-07-10
Add cosmwasm example

thpani:
This adds an example spec of a CosmWasm smart contract.
The example follows the tutorial at: https://github.com/Callum-A/cosmwasm-zero-to-hero

The spec is still missing [the last part of the tutorial](https://github.com/Callum-A/cosmwasm-zero-to-hero/tree/main/14%20-%20Query%20Tests), and more Quint-specific parts (state machine, invariants, ...), but I wanted to merge this as intermediate state.

I'm putting it into a new `cosmwasm` directory for visibility – the `cosmos` directory is rather low level.

Towards #52

### DIFF
--- a/examples/cosmwasm/zero-to-hero/vote.qnt
+++ b/examples/cosmwasm/zero-to-hero/vote.qnt
@@ -1,0 +1,284 @@
+/// A Quint specification following Callum-A's cosmwasm-zero-to-hero tutorial:
+/// https://github.com/Callum-A/cosmwasm-zero-to-hero
+///
+/// Numbers in square brackets [nn] refer to the relevant chapter.
+
+module vote {
+
+  /*** TYPES ***/
+
+  // 05 - Contract state
+  // https://github.com/Callum-A/cosmwasm-zero-to-hero/tree/main/05%20-%20State
+
+  type Addr = str
+  val None : Addr = ""
+
+  type UUID = str
+
+  type Config = {
+    admin: Addr,
+  }
+
+  type Poll = {
+    creator: Addr,
+    question: str,
+    options: List[(str, int)]
+  }
+
+  type Ballot = {
+    option: str
+  }
+
+  type ContractState = {
+    config: Config,
+    polls: UUID -> Poll,
+    ballots: (Addr, UUID) -> Ballot
+  }
+
+  // 06 - Instantiate
+  // https://github.com/Callum-A/cosmwasm-zero-to-hero/tree/main/06%20-%20Instantiate
+
+  type MessageInfo = {
+    sender: str
+  }
+
+  type InstantiateMsg = {
+    admin: str
+  }
+
+  // 08 - ExecuteMsg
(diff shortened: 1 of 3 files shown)


examples/cosmwasm/zero-to-hero/vote.qnt
```diff
@@ -0,0 +1,284 @@
...
+  type UUID = str
+
+  type Config = {
+    admin: Addr,
+  }
+
+  type Poll = {
+    creator: Addr,
+    question: str,
+    options: List[(str, int)]
```
thpani, line 25: Ideally, this would be a `str -> int` map, but apparently CosmWasm storage has no first-class support for maps.

I don't trust the logic around this, so using the original encoding here may be a good showcase for Quint :)

examples/cosmwasm/zero-to-hero/vote.qnt
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  type Ballot = {
+    option: str
+  }
```
thpani, line 30: Chose this over `type Ballot = str` to match the Rust type in the tutorial

examples/cosmwasm/zero-to-hero/vote.qnt
konnov: I guess, they are avoiding maps, since maps normally do not have deterministic iteration

konnov: introduce `type u64 = int` and say in a comment that we do not account for overflows/underflows yet?

examples/cosmwasm/zero-to-hero/vote.qnt
```diff
@@ -0,0 +1,284 @@
...
+
+      val updated = if (not(cs.ballots.keys().contains(_key))) {
+        // sender hasn't voted on this poll; simply add the ballot
+        (poll.options, cs.ballots.put(_key, { option: vote }))
+      } else {
+        // sender already voted on the poll; adjust vote count for removing their old vote
+        val ballot = cs.ballots.get(_key)
+        val optionsDecr = poll.options.foldl([], (res, opt) => {
+          // FIXME: in the original, this only adjusts the _first_ option in the list
+          if (opt._1 == ballot.option) res.append((opt._1, opt._2 - 1)) else res.append(opt)
```
konnov, line 132: How does it work in the original contract? It looks to me like an inplace modification, but should not the borrow checker handle that somehow?

konnov (approved):
Looks great! Multiple nice patterns are emerging here

examples/cosmwasm/zero-to-hero/vote.qnt
thpani: > I guess, they are avoiding maps, since maps normally do not have deterministic iteration

Tutorial says: "Some users of other programming languages may suggest using a Map structure however these cannot be stored in cosmwasm storage plus."

Logically, it should be a map, but I'm not sure their logic really implements a map.

PR #1030, opened 2023-07-10
Fix exporting modules that were imported with a qualifier

bugarela:
Hello :octocat: 

This is a simple fix for the scenario:
```
import X as MyX
export MyX.*
```

That is not currently working.

### DIFF
--- a/quint/src/names/collector.ts
+++ b/quint/src/names/collector.ts
@@ -154,6 +154,11 @@ export class NameCollector implements IRVisitor {
       return
     }
 
+    if (def.qualifiedName) {
+      const newTable = new Map([...moduleTable.entries()])
+      this.definitionsByModule.set(def.qualifiedName, newTable)
+    }
+
     const qualifier = def.defName ? undefined : def.qualifiedName ?? def.protoName
     const importableDefinitions = copyNames(moduleTable, qualifier, true)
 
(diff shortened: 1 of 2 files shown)


konnov:
I am confused by a pull request with the label "bug" :))

quint/test/names/collector.test.ts
```diff
@@ -147,6 +147,20 @@ describe('NameCollector', () => {
       assert.deepEqual(definitions.get('T'), { kind: 'type', reference: 0n, typeAnnotation: { kind: 'int', id: 0n } })
     })
 
+    it('exports previously imported qualified definitions', () => {
```
konnov, line 150:
```suggestion
    it('exports imported definitions that were previously qualified', () => {
```

quint/src/names/collector.ts
```diff
@@ -154,6 +154,11 @@ export class NameCollector implements IRVisitor {
       return
     }
 
+    if (def.qualifiedName) {
+      const newTable = new Map([...moduleTable.entries()])
```
konnov, line 158: I believe you that it fixes the issue, but I don't understand why it fixes the issue. Could you please add a comment here on how `export Foo.*` and `import ...` are cooperating here.

konnov (commented):
This looks like a fix, but I would like to have a comment about the fix, so we do not have to think hard about it in the future.

bugarela:
> I am confused by a pull request with the label "bug" :))

If there was an issue for this, the issue would be labeled as bug. Since I'm using the PR to track the work in the board instead of an issue, I thought the label bug would be useful. Is a PR that fixes a bug :)

konnov (approved):
Looks great! Let's merge it!

PR #1031, opened 2023-07-10
Fix missing name resolution error in incremental compilation

bugarela:
Hello :octocat: 

This fixes the problem described here: https://github.com/informalsystems/quint/discussions/949#discussioncomment-6377955


quint/src/runtime/compile.ts
```diff
@@ -175,8 +177,17 @@ export function compileDef(
 ): CompilationContext {
   // Define a new module list with the new definition in the last module,
   // ensuring the original object is not modified
-  const modules: QuintModule[] = [...state.modules] // Those are flat, but introducing `def` might make them non-flat
-  const lastModule: FlatModule = state.modules[state.modules.length - 1] // This is not modules.pop() to ensure flatness
+  const originalModules = [...state.originalModules]
+  const originalLastModule = originalModules.pop()
+  if (!originalLastModule) {
+    throw new Error('No original modules in state')
+  }
```
shonfeder, line 184: iiuc, this violates a contract for this method which we should maybe document on the method?

Also, given that this check and the analogous one on line 191 are checking requirements of this function, my personal preference would be to group these both at the top of function as part of its "preamble". But I'm not stubborn on this point if you disagree.

quint/test/runtime/compile.test.ts
```diff
@@ -1053,5 +1054,33 @@ describe('incremental compilation', () => {
       const computable = context.evaluationState?.context.get(kindName('callable', def!.id))!
       assertComputableAsString(computable, '3')
     })
+
+    it('uses the oringial modules to resolve names', () => {
```
shonfeder, line 1058:
```suggestion
    it('uses the original modules to resolve names', () => {
```

shonfeder: Unless I misunderstand, this test name seems to reflect an implementation detail rather than the behavior that is actually being tested, which is more like: "non-exported imports are not visible in subsequent importing modules".

quint/src/runtime/compile.ts
```diff
@@ -60,6 +60,7 @@ export interface CompilationContext {
 export interface CompilationState {
   // The ID generator used during compilation.
   idGen: IdGenerator
+  originalModules: QuintModule[]
```
shonfeder, line 63: Please add a comment explaining what this is used for, since it's a bit subtle.

shonfeder (commented):
Nice fix! My only substantive request is for a bit of commentary on the solution :)

quint/src/repl.ts
```diff
@@ -426,7 +427,7 @@ function loadFromFile(out: writer, state: ReplState, filename: string): boolean
       const defsAndExprs = modulesAndRepl[1].split(replEnd)
 
       // save the definition history
-      newState.defsHist = defsAndExprs[0]
+      newState.defsHist += defsAndExprs[0]
```
konnov, line 430: I am surprised by `=` changed to `+=` here. My expectation is that `.load` forget all previous definitions.

quint/src/runtime/compile.ts
```diff
@@ -175,8 +177,17 @@ export function compileDef(
 ): CompilationContext {
   // Define a new module list with the new definition in the last module,
   // ensuring the original object is not modified
-  const modules: QuintModule[] = [...state.modules] // Those are flat, but introducing `def` might make them non-flat
-  const lastModule: FlatModule = state.modules[state.modules.length - 1] // This is not modules.pop() to ensure flatness
+  const originalModules = [...state.originalModules]
```
konnov, line 180: It looks like you could use `splice` here, instead of copying the array and modifying it

quint/src/repl.ts
```diff
@@ -123,7 +123,8 @@ class ReplState {
     // Introduce the __repl__ module to the compilation state so new expressions
     // are pushed to it.
     const state = newCompilationState()
-    return { ...state, modules: [{ name: '__repl__', defs: simulatorBuiltins(state), id: 0n }] }
+    const modules = [{ name: '__repl__', defs: simulatorBuiltins(state), id: 0n }]
+    return { ...state, modules: modules, originalModules: modules }
```
konnov, line 127: Afaik, the only purpose of `__repl__` is to carry auxiliary definitions. Since we have `modules` and `originalModules` now, would not it be easier just to save the auxiliary REPL definitions as an array of definitions and inject it in the main module?

konnov (commented):
Left a few comments

quint/src/repl.ts
bugarela: Yes, I think we should do that! I mentioned it in the last quint meeting. Will do in a followup.

quint/src/repl.ts
bugarela: Your expectation is absolutely correct, but this is `loadFromFile`, not `load`. `load` calls `state.clear()` before calling `loadFromFile`.

Another important thing to consider is that `state.clear()` doesn't empty `defsHist`, but set it to the initial state, which is a state containing the simulator builtins (from `newCompilationState()`), so `=` will overwrite that while `+=` will keep the builtins. 

This is something I found while debugging, not necessarily related to the actual bug. But seems like I got something wrong because we have integration tests breaking because of this. I'll look into that.

bugarela: Ah, the `__repl__` module is also written to the saved file, so those are loaded back twice now. Well, this will all change when I get rid of `__repl__` in the followup as you suggested, so I just reverted this change for now.

quint/src/runtime/compile.ts
bugarela: We cannot change the original array, so I'm not seeing how `splice` would help here.

quint/src/runtime/compile.ts
```diff
@@ -60,6 +60,8 @@ export interface CompilationContext {
 export interface CompilationState {
   // The ID generator used during compilation.
   idGen: IdGenerator
+  // A list of modules as they are constructed, without flattening.
```
shonfeder, line 63: Sorry to be a stickler, but I think it would help for this comment to note why this is needed. E.g., something like

```suggestion
  // A list of modules as they are constructed, without flattening. This is needed to derive correct name resolution during incremental compilation in a flattened context. 
```

My thought is that this info will then make sense of all the later uses. We of course risk the comment growing stale, but we always do. :)

the principle motivating my suggestion is that we want the comments to tell help us understand the why, and we mostly leave it to the code to self-document the what.

bugarela: Makes a lot of sense! Thanks

PR #1039, opened 2023-07-11
Only create a `__repl__` module when a main module is not available

bugarela:
Hello :octocat: 

We've been unnecessarily creating a `__repl__` module for every REPL instance, even when a main module is specified. This PR makes it so we only create a `__repl__` module when no main module is specified.

For `.save` and `.load` commands consistency, I also introduced a `// @mainModule <name>` annotation to REPL-generated files (from `.save`), so the same main module is loaded with `.load`.


bugarela:
One thing that's bothering me is that now `quint -r filename.qnt::module` doesn't have the same behavior as `quint -r filename.qnt` + `import module.*`, since the latter will add a `__repl__` module. I'm not sure if we should make some magic so that the first `import` automatically replaces the `__repl__` module somehow.

PR #1044, opened 2023-07-13
Extend CosmWasm example with state machine + invariants

thpani:
Update the CosmWasm `from-zero-to-hero` example with a state machine + invariants.

Closes https://github.com/informalsystems/apalache-internal/issues/52


konnov (approved):
Nice! Looking at how Cosmwasm contracts are usually structured, would it make sense to break down the spec in exactly the same files?

thpani:
> Nice! Looking at how CosmWasm contracts are usually structured, would it make sense to break down the spec in exactly the same files?

Good point, I will do a follow-up PR!

Also, we need sum types! ☝️😄

PR #1047, opened 2023-07-17
Fix `SimpleAuction` example for dashboard

thpani:
Fix `SimpleAuction` example:
* Add a workaround for #693
* Rename .qnt file to match module name

(This fixes `quint test`. `verify` is N/A because there's no state machine, only runs.)


konnov (approved):
LGTM

bugarela (approved):

bugarela:
Seems like GitHub might have lost a webhook or something. All checks have passed but it was not merged

PR #1049, opened 2023-07-17
Fix prisoners for examples dashboard

thpani:
Fix `prisoners` example in the dashboard.

This one only needed a `--main` flag, and fixing the `import` statement (which is [to be revised](https://github.com/informalsystems/quint/discussions/949)).


bugarela:
I see there are failing integration tests. That's because the `run` command still requires the export. I have a fix for that locally, I'll see if I can easily extract it from my changes into a PR. If you can hold this until then, that'd be great.

PR #1050, opened 2023-07-17
produce a friendlier error on const

### LINKED ISSUE #1048: Improve error message when `run`ning a parameterized module
Launching `quint run` on a parameterized module gives weird errors that the `const` names are not found:

```
$ quint run prisoners.qnt
An example execution:

[failure] Found an issue (44ms).
Use --seed=0x0 to reproduce.
<module_input>:19:23 - error: Name allPrisoners not found
19:     val nonCounters = allPrisoners.exclude(Set(counter))
                          ^^^^^^^^^^^^

<module_input>:19:48 - error: Name counter not found
19:     val nonCounters = allPrisoners.exclude(Set(counter))
                                                   ^^^^^^^
```

We should give a better error message if there are uninitialized constants.
Maybe we could even suggest a list of other modules in the file, to set with `--main`.

konnov:
Closes #1048. The compiler produces a more meaningful message on a constant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 
 ### Added
+### Changed
 
 - `quint run` produces a friendlier message when it meets a `const` (#1050)
 
-### Changed
 ### Deprecated
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 
 ### Added
+
+- `quint run` produces a friendlier message when it meets a `const` (#1050)
+
 ### Changed
 ### Deprecated
 ### Removed
--- a/quint/src/quintError.ts
+++ b/quint/src/quintError.ts
@@ -45,7 +45,7 @@ export type ErrorCode =
   | 'QNT007'
   /* QNT099: Found cyclic definitions */
   | 'QNT099'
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 
 ### Added
+
+- `quint run` produces a friendlier message when it meets a `const` (#1050)
```
thpani, line 12: Should this be under `Changed`?

quint/src/quintError.ts
```diff
@@ -61,8 +63,8 @@ export type ErrorCode =
   | 'QNT406'
   /* QNT407: Cannot import self */
   | 'QNT407'
-  /* QNT099: Found cyclic definitions */
-  | 'QNT099'
+  /* QNT500: Unitialized constant */
+  | 'QNT500'
```
thpani, line 67: Just curious, do we have a numbering scheme for these?

thpani (approved):
LGTM

quint/src/quintError.ts
konnov: Not really, I used `5**`, as REPL and simulator are at phase 5.

CHANGELOG.md
konnov: Yeah, I also thought about this section.

shonfeder: I’d say fixed :)

changed is meant to signal breaking changes I think, and add is for new features, but not-great error messages seem more like bugs to me. See https://keepachangelog.com/en/1.0.0/ for reference.

thpani: Oh, right, I didn't see the `Fixed` section below.
Should go there.

PR #1051, opened 2023-07-17
Remove wrapper `__run__` module from simulation

### LINKED ISSUE #1041: Use the main module instead of creating a `__run__` module in `run`.
Similar to what we did in https://github.com/informalsystems/quint/pull/1039

bugarela:
Hello :octocat: 

Closes #1041. Unblocks https://github.com/informalsystems/quint/pull/1049. This changes the simulator so it doesn't wrap the code inside a `__run__` module, which was done with the goal of adding auxiliary definitions, and was a problem because required the simulated spec to export definitions.

For that, I'm actually re-printing the parsed defs, and adding the auxiliary definitions to the main module as the printed modules are reconstructed. This is not ideal, as the pretty printer can print stuff differently than the user input (because it prints all in normal form), which can lead to slightly different code in the error messages. However, there shouldn't be many errors at the `run` stage.

I feel like we want to get rid of `compileFromCode` and have a more appropriate interface that doesn't require re-constructing source code and re-runs parsing/typechecking stuff. But, since this is a pressing issue blocking work (both for me and for Thomas), I propose we fix it like this and later improve the interfaces.

Also, I had to fix/update `IRPrintinting` in order for this approach to work.


bugarela:
Well, injecting an `export` (actually, one export per import) would also require pretty printing, but only for the `__run__` module which is more stable. However, until I get the flattening right, `export` is our most unstable feature :grimacing:.

I believe there is an ideal solution here with a combination of this (injecting simulator defs) and #1052. This is actually what is done for the REPL when a main module is supplied (load the modules and inject auxiliary defs). If y'all think it's better not to merge this and do it properly, I can hold this until I finish flattening and then do it. The reason why I'm proposing this intermediate change is because I already had it locally as a quick hack for my local tests regarding flattening, and I thought it would be useful to unblock things possibly hitting problems with export.

bugarela:
Another possible temporary solution is to regex match the main module in the input and add the definitions as text there.

bugarela:
It's actually going throught `IRprinting.ts`, which I wrote and it's quite simple, but I also don't trust it a lot. It is definitely a problem with error reporting. I can try the regex solution if you think that's a better temporary fix.

bugarela:
As I explained in the standup: regex was more complicated than I thought, but I figured I can use the index in the `Loc` returned by the parser, which points exactly to where the module begins and ends. I believe that's good enough for now :)

thpani (approved):
LGTM! Much better than re-printing everything 🎉😁

thpani:
`prisoners` is also working now: #1049 🤩 Amazing!

PR #1056, opened 2023-07-18
Output the execution stack on undefined values in REPL

konnov:
This is a quality of life improvement in REPL. I've noticed that I would usually like to see an execution trace when a test is failing in REPL, when setting `.verbosity` to 3-4. This is what this PR does.

### DIFF
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -279,6 +279,7 @@ exit $exit_code
 
 
   Use --verbosity=3 to show executions.
+      HOME/failingTestCounters.qnt
 ```
 
 ### test counters produces no execution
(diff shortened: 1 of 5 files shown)


quint/io-cli-tests.md
```diff
@@ -279,6 +279,7 @@ exit $exit_code
 
 
   Use --verbosity=3 to show executions.
+      HOME/failingTestCounters.qnt
```
thpani, line 282: What is `HOME` here?

thpani (approved):
LGTM

quint/io-cli-tests.md
konnov: It's just a placeholder for a path, so we do not have to use a real path in the test output

Issue #673, opened 2023-03-02
Draft ADR on sum types syntax, statics, and dynamics


shonfeder:
Supports #233 / #539 

Will include:

- Plans for concrete syntax
- Mapping of concrete syntax into row types for statics
- Desugaring of concrete syntax into conditionals and eliminator for dynamics

A draft is in progress.

PR #1057, opened 2023-07-18
Remove redundant type constraints on record label strings

shonfeder:
The way we construct and deconstruct records is interesting: in the
normal form we interpret a special set of operators such that
string-valued expressions given as arguments in our object language are
"lifted" into product type indexes in our meta language. E.g., in `label(r,
"foo")` we extract the value of the internal representation of the quint
string `"foo"` and use this as the label for projecting a value out of
the record `r`.

This lets us do some interesting (and perhaps undesirable?) things, like
construct labels which cannot be used to project out values using the
`r._` syntax sugar:

```
>>> Rec(":", 1)
{ :: 1 }
```

In the process of lifting the value of these strings into labels, we
also verify that each label supplied is in fact a string litteral:

E.g.,

https://github.com/informalsystems/quint/blob/be8bb62dbc046da134d07f44367c4a60970adb45/quint/src/types/specialConstraints.ts#L93-L99

We were previously *also* adding constraints to the our type checking
that reasserted this invariant, but those constraints aren't necessary.
For the core reason that the operator that construct, update, and
eliminate records are not "normal operators" but special constructs, and
the strings (or, for tuples) supplied to describe the labels do not end
up as expressions that require type constraints, but are "puns" that
represent the index in question.

This change is a side effect of WIP on #673, during which I've noted that
our rule was

As an example, prior to this change, our rule for record projection with `field` was

$$
\frac
{ \Gamma \vdash e \colon (r, c) \quad \Gamma \vdash \langle l \rangle \colon str \quad fresh(t) }
{ \Gamma \vdash field(e, \langle l \rangle): (t, c \land r \sim \\{ \ l : t |
tail\_t \ \\} \land str \sim str )}
$$

Where $\langle l \rangle$ stands for the quint string literal with the internal
string representation `l`. Formalizing this in a more abstract representation 
makes it obvious that the last constraint is vacuous.

The rule after this change is:

$$
\frac
{ \Gamma \vdash e \colon (r, c) \quad \Gamma \vdash \langle l \rangle \colon str \quad fresh(t) }
{ \Gamma \vdash field(e, \langle l \rangle): (t, c \land r \sim \\{ \ l : t | tail\_t \ \\}) }
$$

Thanks to @bugarela for talking the current impl over with me prior to me opening this PR.

-------

### DIFF
--- a/quint/src/types/specialConstraints.ts
+++ b/quint/src/types/specialConstraints.ts
@@ -64,7 +64,10 @@ function recordConstructorConstraints(
   const constraints: Constraint[] = []
   // A record constructor has the normal form Rec('field1', value1, 'field2', value2, ...)
   // So we iterate over the arguments in pairs (chunks of size 2)
-  const fields = chunk(args, 2).map(([[key, keyType], [_value, valueType]]) => {
+  //
+  // - We can ignore the _keyType because we are verifying here that every key is a string litteral
+  // - We can ignore the _value because we are only doing type checking
+  const fields = chunk(args, 2).map(([[key, _keyType], [_value, valueType]]) => {
     if (key.kind !== 'str') {
       return left(
         buildErrorLeaf(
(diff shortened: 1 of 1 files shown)


konnov (approved):
This makes sense to me. Thanks for the detailed analysis

PR #1060, opened 2023-07-19
Refactor value generators in the simulator

### LINKED ISSUE #954: REPL returns `<undefined value>` instead of failing, when picking from maps with unbounded (`Int`) codomains
```
val m = setOfMaps(1.to(5), Int)
```
fails with `Error: Infinite set Int is non-enumerable`, as expected, however

```
var x: int
  
action foo = {
    nondet y = oneOf(setOfMaps(1.to(5), Int))
    x' = y.get(1)
}
```

returns `<undefined value>` when attempting to execute `foo`.

konnov:
Closes #954. To support generation of maps over infinite co-domains, I had to refactor the interface of `pick` in `RuntimeValue`. I like the new interface better, as it clearly exposes an iterator instead of doing tricky arithmetic over indices.

**This is a breaking change.** Since it changes how `oneOf()` picks values, the seeds in the tests are not valid anymore.

~I have an impression that the simulator is slower now. Have to do some measurements.~


examples/classic/sequential/BinSearch/BinSearch10.qnt
```diff
@@ -4,5 +4,6 @@ module BinSearch10 {
     INPUT_SEQ=[ 1, 2, 3, 4, 5, 6, 8, 9, 10, 33 ],
     INPUT_KEY=7,
     INT_WIDTH=16
-  ).* from "./BinSearch"
-}
\ No newline at end of file
+  ) as BS from "./BinSearch"
+  export BS.*
```
konnov, line 8: @thpani did I just break the apalache integration?

thpani: Yeah, although I don't understand why it breaks with the export 😂

konnov: Let's maybe disable this test for now, as it waiting for the fix similar to #1051

CHANGELOG.md
```diff
@@ -10,10 +10,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - `quint repl` produces an evaluation trace on errors too (#1056)
+- `S.setOfMaps(Int).oneOf()` is now supported (#1060)
 
 ### Changed
 
 - `quint run` produces a friendlier message when it meets a `const` (#1050)
+- The behavior of `oneOf` has changed, update the seeds in `quint test` (#1060)
```
thpani, line 18: Perhaps be more explicit about this:

```suggestion
- Breaking: the behavior of `oneOf` has changed, existing seed values for `quint test` can exhibit different behavior than before (#1060)
```

thpani: And maybe move it up, to be the first item.

quint/cli-tests.md
```diff
@@ -308,31 +308,37 @@ fi
...
       --invariant=noNegativeInv --main=gradualPonziTest \
       ../examples/solidity/GradualPonzi/gradualPonzi.qnt
 
-### FAIL on run gradualPonzi::progressInv
+### OK (FAIL?) on run gradualPonzi::progressInv
 
-<!-- !test exit 1 -->
+This test should fail. We know that from `quint verify`.
+However, `quint run` has hard time finding a counterexample.
+If you see this test failing, you are a lucky winner!
```
thpani, line 326: Forget elliptic curves, we can do PoW in Quint 🤣

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1391,19 +1391,28 @@ export class CompilerVisitor implements IRVisitor {
...
+            .map(sz => {
+              if (sz === 0n) {
+                this.errorTracker.addRuntimeError(sourceId, `Applied oneOf to an empty set`)
+              }
+              return this.rand(sz)
+            })
+            // An infinite set, pick an integer from the range [-2^255, 2^255).
+            // TODO: make it a configurable parameter:
+            // https://github.com/informalsystems/quint/issues/279
+            .or(just(-(2n ** 255n) + this.rand(2n ** 256n)))
```
thpani, line 1410: Can't the codomain be `Nat` and shouldn't the lower bound then be non-negative?

thpani (approved):
One question about `Nat` as co-domain, otw LGTM

quint/src/runtime/impl/compilerImpl.ts
konnov: Yep, the implementation of `pick` for Int/Nat checks that and uses the absolute value in case of `Nat`. I just do not want to differentiate them here.

thpani: Okay. Maybe leave a comment then, so it's not confusing in the future?

PR #1062, opened 2023-07-19
RFC 001: Extend Quint type system to support Sum Types

### LINKED ISSUE #673: Draft ADR on sum types syntax, statics, and dynamics

shonfeder:
This RFC proposes the design for sum types to be introduced to quint.

The overview gives the gist of the proposal and detailed discussion follows.

Closes #673.


doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+
+```quint
+// Constructoin of a previously declared sum type
+val a : T = A(1)
+// Construction of a polymorphic variant of an anonymous type
+val b : AdHocVariant(int) | v = AdHocVariant(1)
+```
+
+Identifiers starting with a capital letter would them be available only for
+module names, type aliases, and variant constructors.
```
konnov, line 63: This is probably too restrictive. When people are specifying protocols that come from the Golang or Rust implementation, they might want to keep they uppercase notation. I have seen it several times already.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+
+This is the only way I can think of to preserve a Rust-like syntax for sum type
+injectors and also support the anonymous sum types that available via
```
konnov, line 66: Just curious, why are they called "injectors"? I thought these things were called "constructors", thought I am sure there are over 9k names for the same thing in the literature.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+row-ploymorphic sum types. If we instead require all variant types to be defined
+in advance.
+
+In any case, I think there is a gain in readability and parseability in ensuring
+we reason about code purely locally to determine whether a normal operator is
+being applied or
```
konnov, line 72: something seems to be missing here

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+Is normalized as
+
+```quint
+type S = C({})
+```
+
+
+#### Injection
+
+I propose introducing a lexical rule to syntactically differentiate data
```
konnov, line 50: What is the motivation for this lexical rule? As far as I understand, a type definition is introducing constructors for each option, similar to `pure def A(i: int) = ...`. Why would we need this lexical rule? We could say that is a **good practice** to lexically distinguish type names from operator names.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+
+#### Injection
+
+The typing rule for constructing a variant of a sum type:
+
+$$
+\frac
+{ \Gamma \vdash e \colon (t, c) \quad \Gamma \vdash \`l\` \colon str \quad fresh(s) }
+{ \Gamma \vdash \ inj(\`l\`, e) \ \colon (s, c \land s \sim \\{ \ l \colon t | tail\_s \ \\}) }
+$$
```
konnov, line 134: VSCode preview produces a syntax error, most likely, on `\`l\``, but I could get the meaning.

This rule poses one interesting question. If we assume that the user specifies the type beforehand, e.g., via:

```
type T =
  | A(int)
  | B(str)
```

then the type of `T` should be closed, that is, it does not contain a row variable. There are exactly two variables. I do not see how it is reflected in the above rule. If the above rule assumes open types, then we would be able to mix constructors of different sum types. I don't think it is the expected behavior, e.g., in Scala.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+
+#### Declaration
+
+This group of alternatives is closer to Rust, but with significant variation for
+reasons discussed in [Syntax](#syntax).
+
+```quint
+type T =
+  | A(int)
+  | B(str)
```
konnov, line 31: Just looking at [Rust enum types](https://doc.rust-lang.org/book/ch06-01-defining-an-enum.html), I can actually see how they understand sum types to be generalization of C enums. As far as I remember, Haskell also has special keyword for sum types, like `data` or something similar.

If we replace `|` in `match` (see my comment below) with `,`, then the Rust syntax makes sense to me:

```rust
enum T {
  A(int),
  B(str),
}
```

I have noticed it before, but this syntax is actually closer to the variant types in Apalache than I thought.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+#### Projection
+
+Here is a concrete example of projecting a value out of a record using our
+current syntax:
+
+```quint
+val r : {a:int} = {a:1}
+val ex : int = r.a
+// Or, using our exotic field operator, which is currently the normal form
+val ex_ : int = r.field("a")
```
konnov, line 345: By reading this, I realized that our normal form is just the abstract syntax tree (AST) that is exposed in the syntax. For many of the Quint operators, the normal form (AST) coincides with the surface syntax. For some cases, like `r.field("a")`, the surface syntax is different. I guess, the same would happen if we tried to express Quint syntax in Lisp, though we could use atoms there, e.g., `(field r 'a)`.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+$$
+
+This tells us that for any expressions $e_1 : \tau_1, \ldots, e_n : \tau_n$
+derivable from our context we can form a product that indexes those
+$n$ expressions by $i_1, \ldots, i_n$ distinct labels, and packages all data
+together in an expression of type $\\{ i_1 \hookrightarrow \tau_1, \ldots,
+i_n \hookrightarrow \tau_n \\}$. If we're given all the things of the needed
+types, we can conjoint them all together into one compound package.
+
+The following rule describes our current implementation:
```
konnov, line 495: This does not render in VSCode, and it renders strangely on Github. I must admit that reading complex LaTeX in markdown is quite hard. What I did not notice in this rule is a row variable. Maybe I just missed it.

doc/rfc001-sum-types.md
shonfeder: I do mention this option. It’s just that I don’t think this is compatible with support for anonymous (not previously defined) sum types. If we require all sum types to be defined in advance, then we can um generate the constructors and don’t need the lexical rule. 

But as I note, I also think there is an inherent advantage to enabling readers to see what is a what is a sum type constructors just by looking at he code locally.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+together so that the recipient _can_ chose any thing; thus, when a record is
+constructed we must supply all the things and there is no room for variability
+in the row. For sum types, in contrast the constructor _can_ supply any one thing (of
+a valid alternate type), and requires the recipient _must_ be prepared to
+handle every possible alternative. 
+
+In the presence of row-polymorphis, however, the responsibility of the recipient
+is relaxed: the recipient can just handle a subset of the possible alternatives,
+and if the expression falls under a label they are not prepared to handle, they
+can pass the remaining responsibility on to another handler.
```
konnov, line 584: This explanation makes sense to me. However, I am not sure how we would handle that in Apalache. Maybe it is not a problem at all, since Apalache variants would require us to write series of `IF-THEN-ELSE` with unsafe access, but we should check that.

doc/rfc001-sum-types.md
shonfeder: Yeah :) they are “injecting” an expressing into the sum type. It’s dual to the projection of an expression out of a product type.

I can probably just use constructor tho :)

doc/rfc001-sum-types.md
shonfeder: Oops, that should end with “whether a value of a sum type is being constructed”.

doc/rfc001-sum-types.md
shonfeder: Yeah, they’re tradeoffs here either way.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+- ReScript: https://rescript-lang.org/docs/manual/latest/polymorphic-variant
+- OCaml: https://v2.ocaml.org/manual/polyvariant.html
+
+Considerations
+
+- Assuming we support anonymous variant types, we need a way of constructing
+  variants without pre-defined constructors. Potential approaches include:
+  - A special syntax that (ideally) mirrors the syntax of the type 
+  - A special lexical marker on the labels (what ReScrips and OCaml do), e.g., 
+  - Reserve uppercase letters for variant injectors
```
konnov, line 640: Now thinking about this, I would prefer not to have anonymous/open variant types. We have them in Apalache, as we had to introduce ad hoc type annotations. However, the very first thing I always have to do in Apalache is to write a few constructors by hand that eliminate polymorphism. For example, see [erc20_mempool_typedefs.tla](https://github.com/informalsystems/tla-apalache-workshop/blob/f327c96e58858e585a7430e820eead97b668c344/examples/erc20/erc20_mempool_typedefs.tla#L18-L36).

This behavior in Apalache is quite annoying. It causes duplication: (1) Declare the variant type; (2) Introduce constructors by hand. I would prefer to have Rust-like `enum` type declarations that implicitely produce constructors.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+#### Copy rust exactly
+
+```rust
+enum T {
+  A(int),
+  B(int)
+}
+```
+
+- Breaks with our current convention around type declarations and keywords.
```
konnov, line 703: alternatively, we could have the form:

```rust
type T = enum {
  A(int),
  B(int),
}
```

konnov (commented):
This is very detailed RFC. Thank you for thinking about the sum types. I reviewed this document with the VSCode plugin, so this may have lead to too many comments. If it's too hard to read on Github, try the plugin.

doc/rfc001-sum-types.md
shonfeder: Sorry about that! I also had trouble with the rendering of these.

I can convert them to ascii if you think that would better?

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+match e {
+  | A(a) => ...
+  | B(b) => ...
+}
+```
+
+`|` can be swapped for `,`, but IMO it is valuable and clarifying to keep the
+syntactic connection with disjunction and with the type syntax. 
+
+The normal form can be `match(e, {...})` to match our current conventions.
```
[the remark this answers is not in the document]
shonfeder: On further thought, of we want a normal form different from the match statement, we should just use the `decomposition` approach from the paper. Then we don't have to do weird things like sort on the lexigraphic order of labels etc.

doc/rfc001-sum-types.md
shonfeder: `data` in Haskell differentiates from `type`, for type aliases, and `newtype` a wrapper thing needed cause of their typeclass system. Every user defined type in Haskell is wrapped in (at least) a unary sum type label, so in that case it is kind of related to sum types. But it is clear from ML that there is no need to have different keywords for records and sum types.

I think the syntax in rust just follows that of C++ https://en.cppreference.com/w/cpp/language/enum

But, critically, this is consistent syntax in rust. They have:

```
struct Foo { ... }
mod foo { ... }
enum Foo { ... }
trait Foo { ... }
impl Foo { ... }
```

If we opt for `enum T { ... }` and don't also change our syntax for declaring record types, we lose the consistency with type defs for our records. So I'd just advocate that we be consistent internally.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+```
+
+`|` can be swapped for `,`, but IMO it is valuable and clarifying to keep the
+syntactic connection with disjunction and with the type syntax. 
+
+The normal form can be `match(e, {...})` to match our current conventions. 
+
+##### As an anonymous operator
+
+We could also consider a case analysis `{ A : a => e1 | B : b => e2 }` -- where
```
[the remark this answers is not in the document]
shonfeder: yeah this was from an earlier version I forgot to remove. fixed now!

doc/rfc001-sum-types.md
shonfeder: this was indeed following the rules for row-polymorphic variants, and constructing a variant in that system leaves the row open. e.g., in utop:

```ocaml
utop # let x = `A(1);;
val x : [> `A of int ] = `A 1
utop # let y : [`B | `A of int] = x;;
val y : [ `A of int | `B ] = `A 1
```

You can freely mix variants so long as they fit structurally with one of the alternatives. In the context where you are working with a closed type, as in the annotated `y` above, you can only use a constructor that is occurs in the alternatives.

Since we are not aiming to take advantage of the row-types any more, as per our discussion, I'll need to tweak the rule.

doc/rfc001-sum-types.md
shonfeder: It's not the AST really, because the AST doesn't have quint strings for labels. These operators are quite weird because the type of the return expression depends on the quint *value* supplied as an argument. I'm not sure we want that kind of dependently typed operator in the long term, but this is a broader question that is properly out of scope for this PR. :)

doc/rfc001-sum-types.md
shonfeder: I've ported to org-mode and will include a generated pdf.

doc/rfc001-sum-types.md
shonfeder: In conversation we agreed that this would create inconsistency in our syntax unless we also changed the syntax for records to be something like

```type T = struct {
  ...
}
```

so we'll stick without the keyword for know.

doc/rfc001-sum-types.md
```diff
@@ -0,0 +1,846 @@
...
+
+```quint
+match e {
+  | A(a) => ...
+  | B(b) => ...
+}
+```
+
+`|` can be swapped for `,`, but IMO it is valuable and clarifying to keep the
+syntactic connection with disjunction and with the type syntax.
```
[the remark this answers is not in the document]
shonfeder: people don't seem to have a problem with `|` in typescrips unions ;)

shonfeder:
I still need to do a pretty extensive editing, clean up, and revision pass, but I've made the following improvements following our conversation:

- Committed a generated pdf, so the latex is easily and reliably readable. Feedback as annotated PDF is welcome!
- Made the syntax more rust-like in the ways we discussed (swapped `|` for `,`)
- Added a fragment from the ERC20 spec that has been adapted to use sum types, so we can get a feel for what this syntax will look like in a spec.
- Changed the injection rule so that we ensure a (necessarily closed) sum type has been defined.
- Some cleanup to the first part of the doc.

Sorry for the rough state of the rest, but 6

konnov (commented):
I left a few comments to the new version. Thank you for producing a pdf. It's way easier to read!

doc/rfc001-erc-sum-types-example.qnt
```diff
@@ -0,0 +1,207 @@
...
+
+    pure def mapState(r:Erc20Result, f:Erc20State => Erc20State): Erc20Result = r match {
+         Ok(s)    => Ok(f(s)),
+         Error(_) => r
+    }
+
+    pure def flatMap(r:Erc20Result, f:Erc20State => Erc20Result): Erc20Result = r match {
+         Ok(s)    => f(s),
+         Error(_) => r
+    }
```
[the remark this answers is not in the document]
shonfeder: Maybe “niceMap”? Lol 😂

doc/rfc001-erc-sum-types-example.qnt
```diff
@@ -0,0 +1,207 @@
...
+    type Erc20State = {
+        balanceOf: Address -> Uint,
+        totalSupply: Uint,
+        allowance: (Address, Address) -> Uint,
+        owner: Address,
+    }
+
+    // Example of a sum type representing a
+    // NOTE: Once we support parametric sum-types we can generalize this alias and the map operators
+    type Erc20Result = {
```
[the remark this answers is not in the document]
shonfeder: Yeah, that is really the worry I have too. I think the `|` is really a nice option. Again, people in type script don’t seem to have any problem with it. 

Having optional keywords is not a good idea imo: it means inconsistent code for readers and unnecessary variance on trivial things. If we must add keywords, let’s please make them required?

doc/rfc001-erc-sum-types-example.qnt
```diff
@@ -0,0 +1,207 @@
...
+
+    pure def flatMap(r:Erc20Result, f:Erc20State => Erc20Result): Erc20Result = r match {
+         Ok(s)    => f(s),
+         Error(_) => r
+    }
+
+    // An auxilliary definition similar to Solidity's require
+    pure def require(r: Erc20Result, cond: bool, msg: str): Erc20State = match r {
+        Error(_) => s,
+        Ok(s)    => if (cond) Ok(s) else Error(msg)
```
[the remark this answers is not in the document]
shonfeder: Ah yeah. S should be r. True that this could be rewritten with flatmap.

doc/rfc001-sum-types.org
```diff
@@ -0,0 +1,858 @@
...
+need a more restrictive typing. This is discussed with the typing rules.
+
+When a sum type declaration is parsed, constructor operations for each
+variant will be generated internally. E.g., for the type =T= defined
+above, we will add to the context the following operators:
+
+#+begin_src quint
+pure def A(x:int): T = variant("A", x)
+pure def B(x:str): T = variant("B", x)
+pure def C(): T      = variant("C", {})
```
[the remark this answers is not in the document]
shonfeder: The latter would conflict with our module system notion: it would not be clear from syntax whether you were accessing a module operator.

If we supported nested modules, we could generate a module with operators for the constructors, but we don’t support that. So I don’t know how we could resolve that ambiguity. I also don’t see the benefit for practically to that syntax Do you have anything in mind?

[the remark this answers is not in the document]
shonfeder: Afaict, Rust doesn’t have that ambiguity because their modules names must be lower case. function names must also be lower case, upper case is reserved for data constructors.

doc/rfc001-erc-sum-types-example.qnt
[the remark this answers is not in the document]
shonfeder: To help us get a clear and holistic vision of how all the bits of syntax fit together in our (and other) languages, I'll put together a table comparing the different constructs in different languages.

I am worried that if we try to over-index on superficial similarity to rust one corner of the language, we will lose the bigger picture, and its import all our syntax works together. Rust's syntax for sum types makes particular decisions in light of other aspects of the syntax, many of which we don't share. So, imo, we should make sure if adopt rust-isms that they fit with the rest of our syntax, or that we make needed adjustments to rustify other aspects of our language too.

doc/rfc001-sum-types.md
shonfeder: See https://github.com/informalsystems/quint/issues/1081 for discussion on this point. I've removed this part from the proposal.

doc/rfc001-sum-types.md
shonfeder: We can translate this into Apalache IR using the `VariantGetUnsafe` and `VariantTag` operators, like the pattern here:

https://github.com/informalsystems/apalache/blob/4e0094ee5d03a068188a85dcdc8ead021b17033e/src/tla/Option.tla#L34-L60

doc/rfc001-erc-sum-types-example.qnt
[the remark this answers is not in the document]
shonfeder: In discussion with the team last week, all three of us agreed that the `|` based syntax seemed most sensible given the rest of our syntax. I've expanded the RFC to discuss the trade offs in some depth, and I've outlined some considerations on how we could change our syntax if we want to pursue a more rust-like syntax.

For the short term, I've moved forward with `|` for the concrete syntax, but we are not considering it stable at all at this point.

konnov (approved):
I read the syntax part in detail and skimmed through the detailed part, as I have read it before. I agree with the syntax. Let's start with this one. We can always change the grammar later. I actually did not know that Quint admits empty records: `{}` does not parse as a record, but `Rec()` does. If we want to rely on `Rec()` being the unit value/type, we have to emphasise this decision in the language document.

PR #1063, opened 2023-07-20
wrap formatError

### LINKED ISSUE #1061: Error reporter fails during test
When running `quint test` on [this spec](https://github.com/informalsystems/partnership-heliax/tree/manuel/quint-report-issue/2023/Q2/artifacts/PoS-quint), a test fails and the error reporter crashes.

Quint version: 0.12.0

The quint test command:
`quint test --main=namada --max-samples=10 --match=slashValidatorTest --verbosity=3 namada-redelegation.qnt`

The error:
```
RangeError: Invalid count value
    at String.repeat (<anonymous>)
    at formatLine (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/errorReporter.js:64:19)
    at /opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/errorReporter.js:50:23
    at Array.reduce (<anonymous>)
    at formatError (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/errorReporter.js:34:25)
    at /opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/cliCommands.js:272:65
    at Array.forEach (<anonymous>)
    at runTests (/opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/cliCommands.js:271:25)
    at EitherConstructor.asyncChain (/opt/homebrew/lib/node_modules/@informalsystems/quint/node_modules/@sweet-monads/either/cjs/index.js:125:16)
    at /opt/homebrew/lib/node_modules/@informalsystems/quint/dist/src/cli.js:33:39
```

Line 64 of errorReporter.js: `output += '^'.repeat(1 + endCol - startCol);`

konnov:
Closes #1061. It happened several times that the error formatter was throwing a new exception, instead of explaining the error. In the long run, we should carefully proof-read `formatError`. This PR wraps `formatError` code in try/catch, so `quint test` does not simply crash in error formatting, but print the stack trace and continues.

I believe that this should be the tool behavior for the non-critical logic.

### DIFF
--- a/quint/src/errorReporter.ts
+++ b/quint/src/errorReporter.ts
@@ -67,8 +67,8 @@ export function formatError(
   } catch (error: any) {
     // If it happens that an error occurs during the error reconstruction,
     // print that we could not construct a good explanation, instead of throwing.
-    console.error('\nSomething unexpected happened during error reconstruction. Report a bug.\n')
-    console.error('Please include this in the bug report: {')
+    console.error('\nSomething unexpected happened during error reconstruction. Please report a bug.\n')
+    console.error('Include this in the bug report: {')
     console.trace(error.message)
     console.error('} // end of the bug report\n')
 
--- a/quint/src/errorReporter.ts
+++ b/quint/src/errorReporter.ts
@@ -39,29 +39,42 @@ export function formatError(
     return `error: ${message.explanation}`
   }
 
-  return message.locs.reduce((output, loc) => {
-    // If lineOffset is a number, print the source location.
-    // If lineOfsset is undefined, omit the source location (e.g., in REPL).
-    const locString = lineOffset.isJust()
-      ? `${loc.source}:${loc.start.line + lineOffset.value}:${loc.start.col + 1} - `
-      : ''
-    output += `${locString}error: ${message.explanation}\n`
+  try {
+    // try to extract the location of the error
+    return message.locs.reduce((output, loc) => {
+      // If lineOffset is a number, print the source location.
+      // If lineOfsset is undefined, omit the source location (e.g., in REPL).
+      const locString = lineOffset.isJust()
+        ? `${loc.source}:${loc.start.line + lineOffset.value}:${loc.start.col + 1} - `
+        : ''
+      output += `${locString}error: ${message.explanation}\n`
 
-    const endLine = loc.end ? loc.end.line : loc.start.line
-    const endCol = loc.end ? loc.end.col : loc.start.col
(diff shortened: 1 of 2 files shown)


quint/src/errorReporter.ts
```diff
@@ -39,29 +39,42 @@ export function formatError(
...
-    for (let i = loc.start.line; i <= endLine; i++) {
-      // finder's indexes start at 1
-      const lineStartIndex = finder.toIndex(i + 1, 1)
-      const line = text.slice(lineStartIndex).split('\n')[0]
+      const endLine = loc.end ? loc.end.line : loc.start.line
+      const endCol = loc.end ? loc.end.col : loc.start.col
+      for (let i = loc.start.line; i <= endLine; i++) {
+        // finder's indexes start at 1
+        const lineStartIndex = finder.toIndex(i + 1, 1)
+        const line = text.slice(lineStartIndex).split('\n')[0]
```
konnov, line 58: I did not change the code inside `try/catch`, simply moved it inside

quint/src/errorReporter.ts
```diff
@@ -39,29 +39,42 @@ export function formatError(
...
+        const lineIndex = lineOffset.map(offs => offs + i)
+        output += formatLine(lineIndex, lineStartCol, lineEndCol, line)
+      }
+      return output
+    }, '')
+  } catch (error: any) {
+    // If it happens that an error occurs during the error reconstruction,
+    // print that we could not construct a good explanation, instead of throwing.
+    console.error('\nSomething unexpected happened during error reconstruction. Report a bug.\n')
+    console.error('Please include this in the bug report: {')
```
thpani, line 71: Let's be polite first, then use the imperative 😄 

```suggestion
    console.error('\nSomething unexpected happened during error reconstruction. Please report a bug.\n')
    console.error('Include this in the bug report: {')
```

quint/src/errorReporter.ts
```diff
@@ -39,29 +39,42 @@ export function formatError(
...
+    console.trace(error.message)
```
thpani, line 72: Can/shall we also print a stacktrace?
In this case it would be helpful. We could make `console.trace` go into a file.

thpani (approved):
LGTM

quint/src/errorReporter.ts
konnov: it does print a trace actually. The message is just the header. I would rather avoid writing to a file, as it can cause another exception %)

PR #1070, opened 2023-07-24
Update ICS20 spec for examples dashboard

thpani:
* Set proper `--main` argument for ics20 specs in example runner
* Update `denomTrace.qnt` imports to work with `verify`
* Add a step action + an invariant to `ics20.qnt`

This gives `bank.qnt` and `denomTrace.qnt` a ✅ in the dashboard.
For `ics20.qnt`, the workarounds for #693 would've been excessive, so it's still marked as :x:


bugarela (approved):

examples/.scripts/run-example.sh
```diff
@@ -25,6 +25,12 @@ get_main () {
     main="--main=LamportMutex_3_10"
   elif [[ "$file" == "classic/distributed/ReadersWriters/ReadersWriters.qnt" ]] ; then
     main="--main=ReadersWriters_5"
+  elif [[ "$file" == "cosmos/ics20/bank.qnt" ]] ; then
+    main="--main=bankTests"
+  elif [[ "$file" == "cosmos/ics20/denomTrace.qnt" ]] ; then
+    main="--main=properChannelsTests"
+  elif [[ "$file" == "cosmos/ics20/ics20.qnt" ]] ; then
+    main="--main=ics20Test"
```
shonfeder, line 33: A general remark rather than something to address in this PR: This is my first time seeing these configs in the runner script. I can understand the motivation, but I think this is not a very sustainable or discoverable way to address the need.

My naive assumption was that we could set up all the examples so that they can run with a plain invocation. I'm wondering why the "main" module for all these examples is a `*Test* module? His would make sense to me if it were only for `test`, but since we are using it for verify as well, perhaps these modules should be renamed to match the file name, so that we can run without additional configs?

WDYT?

shonfeder (approved):

examples/.scripts/run-example.sh
thpani: > perhaps these modules should be renamed to match the file name, so that we can run without additional configs?

Yeah, I agree this is a hack, not too happy about these either. But also, I think we have a bit too much variation in the config options to pass all parameters by convention:

In the context of `const`, we have been using `ParaemeterizedModule_X_Y` for a module instance setting constants to values `X` and `Y`. In the long run, maybe we need a CSV file or put commands in the actual spec to store these flag values.

> I'm wondering why the "main" module for all these examples is a `*Test* module? His would make sense to me if it were only for `test`

This has been roughly our convention so far. Basic (functional) protocol logic goes into module `X`, state machine into module `XTest`. The namespace is becoming crowded with the Apalache integration, but we haven't really arrived at a good separation. For example, `test` needs the actions, but not `init` and `step`, while `run` and `verify` need `init` and `step`. However, it makes sense to put `init` and `step` in the same module as the other actions. A lot of this can be resolved after we fix import/export (#949) to make proper use of multiple modules between `test`, `run`, and `verify`.

TLDR: I figured we can hack these in for now, then do a cleanup when all the requirements are clear, and a convention emerges from usage patterns.

shonfeder: > or put commands in the actual spec to store these flag values

I suspect this is the way! Things that are useful for testing and CI often end up being useful per se: having a way of setting default configs and params seems useful imo.

> TLDR: I figured we can hack these in for now, then do a cleanup when all the requirements are clear, and a convention emerges from usage patterns.

Makes sense!

PR #1075, opened 2023-07-24
Use gRPC reflection to retrieve CmdExecutor file

rnbguy:
This PR removes dependency to `APALACHE_DIST` while retrieving `CmdExecutor.proto` file.

If `APALACHE_DIST` is not set, it retrieves the file directly from the default gRPC endpoint `localhost:8822`.

- [x] Put [`reflection.proto`](https://github.com/grpc/grpc/blob/master/src/proto/grpc/reflection/v1alpha/reflection.proto) in local package and remove axios dependency to make http requests.
- [x] Improved coding style with Quint team (I am not sure about my either-monad usage)

_Related to #701, #823_


quint/src/quintVerifier.ts
```diff
@@ -182,7 +184,76 @@ const grpcStubOptions = {
...
+  const reflectionClient = new serverReflectionService(APALACHE_SERVER_URI, grpc.credentials.createInsecure())
+
+  // Query reflection endpoint
+  const call = reflectionClient.ServerReflectionInfo()
+  call.write({ file_containing_symbol: 'shai.cmdExecutor.CmdExecutor' })
+
+  const protoDefResponse: ServerReflectionResponseSuccess = await new Promise((resolve, reject) => {
+    call.on('data', (reflectionResponse: ServerReflectionResponse) => {
+      if ('file_descriptor_response' in reflectionResponse) {
+        call.cancel()
```
thpani, line 225: On my machine, this `cancel` call triggers the `error` handler below.

Not sure why this uses a streaming API in the first place, but should we just accumulate the responses in a `Buffer[]` and instead resolve the promise in `call`'s `'end'` event?

```
Error: 1 CANCELLED: Cancelled on client
    at callErrorFromStatus (/Users/thomas/src/quint/quint/node_modules/@grpc/grpc-js/build/src/call.js:31:19)
    at Object.onReceiveStatus (/Users/thomas/src/quint/quint/node_modules/@grpc/grpc-js/build/src/client.js:410:73)
    at Object.onReceiveStatus (/Users/thomas/src/quint/quint/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:323:181)
    at /Users/thomas/src/quint/quint/node_modules/@grpc/grpc-js/build/src/resolving-call.js:94:78
    at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
for call at
    at ServiceClientImpl.makeBidiStreamRequest (/Users/thomas/src/quint/quint/node_modules/@grpc/grpc-js/build/src/client.js:394:34)
    at ServiceClientImpl.<anonymous> (/Users/thomas/src/quint/quint/node_modules/@grpc/grpc-js/build/src/make-client.js:105:19)
    at loadProtoDefViaReflection (/Users/thomas/src/quint/quint/dist/src/quintVerifier.js:138:35)
    at verify (/Users/thomas/src/quint/quint/dist/src/quintVerifier.js:236:19)
    at verifySpec (/Users/thomas/src/quint/quint/dist/src/cliCommands.js:454:39)
    at EitherConstructor.asyncChain (/Users/thomas/src/quint/quint/node_modules/@sweet-monads/either/cjs/index.js:114:16)
    at /Users/thomas/src/quint/quint/dist/src/cli.js:33:39 {
  code: 1,
  details: 'Cancelled on client',
  metadata: Metadata { internalRepr: Map(0) {}, options: {} }
}
```

thpani (commented):
Really cool! 🎉 Thanks a lot for the contribution, @rnbguy! 🙏

I've cleaned up the code a bit, and added the reflection proto locally.
Would also like to get @shonfeder's opinion before approving.

Is there a reason you used [`v1alpha/reflection.proto`](https://github.com/grpc/grpc/blob/master/src/proto/grpc/reflection/v1alpha/reflection.proto) instead of [`v1/reflection.proto`](https://github.com/grpc/grpc/blob/master/src/proto/grpc/reflection/v1/reflection.proto)?

quint/package.json
```diff
@@ -67,7 +67,7 @@
     "yargs": "^17.2.1"
   },
   "scripts": {
-    "compile": "genversion -e src/version.ts && tsc && copyfiles -u 1 ./src/builtin.qnt ./dist/src/",
+    "compile": "genversion -e src/version.ts && tsc && copyfiles -u 1 ./src/reflection.proto ./src/builtin.qnt ./dist/src/",
```
shonfeder, line 70: My understanding was that the big benefit of dynamically generating the proto interface rather than generating it at compile time is that we don’t have to add a proto file dep and don’t need a build step. This leads to a ( probably very naive) question: if we are going to be adding a build step to copy over a proto file (which we’ll then use to generate glue code for the reflection I guess) why not instead add a build step the grabs the Apalache proto file? Then we don’t have to worry about the reflection request, and can just read off the interface directly from our proto file. This would let us drop the jar stuff, and and end up with a net reduction in code, rather an increase.

we could even just generate the glue code in that case, saving runtime costs a bit an making it easier to debug.

Am i missing a basic value prop from the reflection approach that justifies the added complexity?

quint/src/reflection.proto
```diff
@@ -0,0 +1,136 @@
+// Copyright 2016 gRPC authors.
```
shonfeder, line 1: Towards my other point re: needing a proto file: this proto file is more complex than the Apalache proto file. So we’d win on reduced code here too.

quint/src/quintVerifier.ts
```diff
@@ -182,7 +184,76 @@ const grpcStubOptions = {
...
+    }
+  }
+  type ServerReflectionResponse = ServerReflectionResponseSuccess | ServerReflectionResponseFailure
+  type ServerReflectionService = {
+    new (url: string, creds: any): ServerReflectionService
+    ServerReflectionInfo: () => any
+  }
+  type ServerReflectionPkg = {
+    grpc: { reflection: { v1alpha: { ServerReflection: ServerReflectionService } } }
+  }
```
shonfeder, line 206: Perhaps this coming from too much of a static mindset, but it makes me feel it a bit nervous to have types generated dynamically in a function body! XD

Is this a normal typescript idiom? Any reason we can't define these outside of the function?

quint/src/quintVerifier.ts
```diff
@@ -182,7 +184,76 @@ const grpcStubOptions = {
...
+
+  // Obtain a reflection service client
+  // To load from HTTP:
+  // const protoJson = protobuf.parse(response.data, grpcStubOptions).root.toJSON()
+  // const packageDefinition = proto.fromJSON(protoJson, grpcStubOptions)
+  const protoPath = require.resolve('./reflection.proto');
+  const packageDefinition = proto.loadSync(protoPath, grpcStubOptions);
+  const reflectionProtoDescriptor = grpc.loadPackageDefinition(packageDefinition) as unknown as ServerReflectionPkg
+  const serverReflectionService = reflectionProtoDescriptor.grpc.reflection.v1alpha.ServerReflection
+  const reflectionClient = new serverReflectionService(APALACHE_SERVER_URI, grpc.credentials.createInsecure())
```
shonfeder, line 216: So with this change, we're going to have to maintain two grpc clients, one of which is just to fetch the file from our own code to generate the second? This seems a bit round-a-bout to me...

shonfeder (commented):
Thanks for the contribution, Rano!

rnbguy:
@thpani hey good catch ! I think I randomly searched online and picked a link from the official repo.

You should use `v1` over `v1alpha`.

quint/src/reflection.proto
rnbguy: My typescript is rusty. The idea is to use a gRPC client that supports the reflection API. I can't find it on npm. So this is my solution. You can definitely statically compile reflection.proto to typescript and import it in the Quint package.

quint/src/quintVerifier.ts
rnbguy: We can definitely put it outside. I kept them in the function as I wanted to keep them in limited scope. But, it uses the same style used for CmdExecutor.

quint/src/quintVerifier.ts
rnbguy: How is this effort any different from extracting the proto file from the jar file? 😅

The ideal solution is to compile and public Apalache proto files on npm 😄

quint/package.json
shonfeder: Thomas kindly talked this thru with me sync. I'm still not convinced that this is worth the added complexity, given how tightly coupled quint and apalache are. But I am convinced that accepting this change won't cause problems for us (assuming Thomas question is worked out).

quint/src/reflection.proto
shonfeder: but if we are gonna statically compile, we can just statically compile the apalache proto file :D

quint/src/quintVerifier.ts
shonfeder: Thanks for explaining! Thomas and Gabriela also explained that this is just a way of controlling the scope of the type aliases. I'd not but it inside a function myself, but won't block on it, and iiuc, Thomas seems to like this approach :)

quint/src/quintVerifier.ts
rnbguy: I also have no clue. For now, I pushed a commit without `call.cancel()` and the error is gone. I will look little bit more before I mark this PR as ready-for-review.

quint/src/quintVerifier.ts
rnbguy: I would not argue against adding Apalache proto file. I think that is a nice solution.

minor correction, my solution doesn't write to disk. it parses the files in memory. nonetheless, your argument stays valid.

quint/src/quintVerifier.ts
rnbguy: @thpani apparently it was not a thrown error. rather they are just `console.log` from `error` and `status` callback. I added new commits. these should be printed anymore.

quint/src/quintVerifier.ts
rnbguy: just wanted to reiterate the differences between different solution to this.

### Extract from Jar

- download and put jar in a location
- pass the location as `APALACHE_DIST`
- create a temporary directory
- extract the proto file by a spawning a separate `jar` process in the temporary directory
- load the client
- delete the temporary directory

Pro: The easiest solution
Con: User has to download and extract the apalache distribution and then pass it as `APALACHE_DIST`

### Extract from gRPC reflection

- include `reflection.proto` - which is a standard file. shouldn't change in near future
- parse it and create gRPC reflection client
- retrieve `cmdExecutor.proto`
- parse and create `CmdExecutor` client

Pro: No need to include apalache source
Con: Dynamic client creation means no statically type checking Apalache gRPC types

### Include `CmdExecutor.proto` in quint

- include `cmdExecutor.proto` in quint
- parse and create `CmdExecutor` client

Pro: Bypass gRPC reflection
Con: Need to sync `CmdExecutor.proto` from apalache source

### Use Apalache gRPC client

- statically compile and publish gRPC client on npm
- directly import the client as dependency

Pro: No gRPC reflection. No parsing. Everything statically compiled.
Con: Need to maintain Apalache gRPC client on npm

shonfeder: Thanks for the analysis. 

I think you are missing the context that the current arrangement is only provisional and part of WIP. There has never been any expectation that end users will have to manually manage the apalache distribution. This is only a temporary state of things to unblock work on the integration in the language level while we design out the UX properly.

In practice, in the current state of things, the user still needs the apalache distribution, because they will be running apalache locally. The reflection is only saving the need to set an environment variable which, as I have noted, will go away with our plans for quint to manage the apalache distribution.

Our current plan is that Quint will manage a compatible apalache distribution on its own, that will be separate from any apalache version users may have installed. It will therefore know where the distribution is located, and none of this should be exposed to users. We can still chose whether we want to extract the proto file from the jar, build it in to the dist, use reflection, or generate a grpc client.

Re

> Con: Need to sync CmdExecutor.proto from apalache source

This con isn't relevant given our actual architecture, afaik: I can't think of any change to the apalache proto files that would break the proto interface but not require changes to our typescrips in quint. The coupling is tight enough that we *have* to keep the code in sync for this kind of thing whether or not it goes thru the proto file. On the other hand, the proto file is extremely minimal, and I also can't think of any changes that are planned and it has not been touched in 9 months.

Here's an option you have not enumerated here: we generate grpc client code and include it in our source, without worrying about an NPM dependency.

We have issues discussing an planning out this whole design track, and any input on this would be welcome.

quint/src/quintVerifier.ts
rnbguy: btw, it is using streaming API because [the service is defined as streaming](https://github.com/grpc/grpc/blob/498fc99479a01177aa2ad257f1390521495b0788/src/proto/grpc/reflection/v1/reflection.proto#L34-L35).

quint/src/quintVerifier.ts
rnbguy: I agree with you Shon. I just improved the current provisional arrangement - that is all :raised_hands: I don't see it to be the final solution under the hood. But it already offers (almost) the same UX with when Quint manages a compatible apalache distribution :slightly_smiling_face:

> Here's an option you have not enumerated here: we generate grpc client code and include it in our source, without worrying about an NPM dependency.

And yes, I missed it. But compiling the grpc client will add extra build steps and dependencies for Quint.

rnbguy:
I finalized the PR. Let me know if I should add any test or update changelog or readme for this.

rnbguy:
@thpani FYI, I just noticed `apalache server` users `v1alpha`.

```bash
$ apalache server &
$ grpcurl -plaintext localhost:8822 list
grpc.reflection.v1alpha.ServerReflection
shai.cmdExecutor.CmdExecutor
shai.transExplorer.TransExplorer
```

thpani:
> @thpani FYI, I just noticed `apalache server` users `v1alpha`.

Yeah, I will replace it today :)

thpani:
> @thpani FYI, I just noticed `apalache server` users `v1alpha`.

Aaah yes! It looks like `v1alpha` is the version baked into `grpc-java` (and all the other languages): https://github.com/grpc/grpc-java/blob/944de93f7987d050a58c8414ec79d677c4aa7709/documentation/server-reflection-tutorial.md#enable-server-reflection

thpani (dismissed):
LGTM
Thanks for helping out here, @rnbguy! 👍

thpani:
I will adapt the Apalache integration tests that currently fail CI, then this is good to merge from my side.

thpani:
It took a some iterations, but I'm finally happy with this 😄
Squashed the main changes into a single commit, to help with reviewing.

@shonfeder PTAL?

shonfeder (approved):
Looks great! Awesome work you both and thanks for improving the interface and kicking this off, @rnbguy !

PR #1078, opened 2023-07-27
Add arity checks for exotic operators

### LINKED ISSUE #1054: Records operators don't type check argument arity correctly
Looks like we don't have the right arity checks in place for the record operators. Here are some examples:

```
>>> r
{ a: 1 }
>>> r.field("a", 1)
static analysis error: error: [QNT000] Expected 3 arguments, got 2

>>> {a: 1}.fieldNames()
Set("a")
>>> {a: 1}.fieldNames(1, 2, 3)
Set()
>>> {a: 1}.with("a", 2, 4, 5, 6)
static analysis error: error: [QNT000] Expected 6 arguments, got 3

>>> Tup(1, 2).item(1, 2, 3)
static analysis error: error: [QNT000] Expected 4 arguments, got 2
```

This is because the constraints that guard correct operator signatures are not applied if we use a "specialConstraint": https://github.com/informalsystems/quint/blob/123a237513bbc92d7b629566a5573f3b076f66ab/quint/src/types/constraintGenerator.ts#L156-L171

I have a fix for this in mind and would like to pick up the issue as a way to ease into work on the type system for #539, assuming no objection from @bugarela.

shonfeder:
Closes #1054

Our "normal form" operator for tuple and records construction and elimination
can't be typed in the quint type system, and the constraint generation for those
operator was not attending to their arity.

This PR introduces a little helper function to validate the arity of arguments
against a predicate on their length, which we then use to check arity is correct
before preceding with constraint generation.

I also add a small refactor that simplifies the logic around dispatching the
constraint generation functions for these operator. It may be helpful to review
by commit.


quint/src/types/constraintGenerator.ts
```diff
@@ -38,14 +38,39 @@ import { getSignatures } from './builtinSignatures'
...
+function validateArity(
+  operator: string,
+  args: [QuintEx, QuintType][],
+  pred: (arity: number) => Boolean,
+  msg: String
+): Either<Error, null> {
+  if (!pred(args.length)) {
+    return left(
+      buildErrorLeaf(
+        `Checking arity for application of ${operator}`,
```
thpani, line 65: The first argument to `buildErrorLeaf` is expected to be a location.
Should we move this inside `ConstraintGeneratorVisitor` so we can use its `location` field?

quint/src/types/constraintGenerator.ts
```diff
@@ -38,14 +38,39 @@ import { getSignatures } from './builtinSignatures'
...
+} from './specialConstraints'
 import { FreshVarGenerator } from '../FreshVarGenerator'
 
 export type SolvingFunctionType = (
   _table: LookupTable,
   _constraint: Constraint
 ) => Either<Map<bigint, ErrorTree>, Substitutions>
 
+function validateArity(
+  operator: string,
```
thpani, line 57: I think this would benefit from a brief explanation.
Also, maybe rename the parameter, to make clear it's a string.

```suggestion
// Check if an application of operator `opcode` to arguments `args`
// has correct arity. In particular, this returns `right(null)` if 
// `pred(args.length)` holds, and a `left(ErrorTree)` otherwise.
//
// Use this for operators that cannot be typed in the Quint type.
//
// `msg` contains a textual description of the expected argument
// length ("1", "2", "even number of", ...).
function validateArity(
  opcode: string,
```

quint/test/types/constraintGenerator.test.ts
```diff
@@ -115,4 +115,58 @@ describe('ConstraintGeneratorVisitor', () => {
...
+    )
+  })
+
+  it('catches invalid arity on Tup operator', () => {
+    testArityError(
+      'val x = Tup()',
+      'Checking arity for application of Tup',
+      'Operator expects at least one arguments but given 0'
+    )
+  })
```
thpani, line 171: Should we also test `'item'`?

thpani (commented):
Looks great, and nice cleanups!

I only have a few small suggestions:

quint/src/types/constraintGenerator.ts
bugarela: That's not really necessary. This location is the `leaf` of the error tree. The returned error here will be combined with the `ConstraintGeneratorVisitor`'s location in its `addToResults` method, and the resulting error will look somewhat like:
```
Error: Operator expects 2 arguments but given 1
Checking arity for operator application of foo
Generating constraints for foo("bar")
```

quint/src/types/constraintGenerator.ts
```diff
@@ -38,14 +38,39 @@ import { getSignatures } from './builtinSignatures'
...
+  args: [QuintEx, QuintType][],
+  pred: (arity: number) => Boolean,
+  msg: String
+): Either<Error, null> {
+  if (!pred(args.length)) {
+    return left(
+      buildErrorLeaf(
+        `Checking arity for application of ${operator}`,
+        `Operator expects ${msg} arguments but given ${args.length}`
```
bugarela, line 66: Should this be
```suggestion
        `Operator expects ${msg} arguments but was given ${args.length}`
```

quint/src/types/constraintGenerator.ts
```diff
@@ -152,31 +177,47 @@ export class ConstraintGeneratorVisitor implements IRVisitor {
...
-        } else {
-          // Otherwise, define a constraint over the signature
-          return signature.map(t1 => {
-            const t2: QuintType = { kind: 'oper', args: results.map(r => r[1]), res: a }
-            const c: Constraint = { kind: 'eq', types: [t1, t2], sourceId: e.id }
-            return [c]
-          })
+    // We want `definedSignature` computed before the fresh variable `a` so that the
+    // numbering of ther fresh variables stays in order, with `a`, used for return types,
+    // bearing the highest number.
```
bugarela, line 182: Thanks for adding that! I realize I should have done that when I faced this issue :sweat_smile:

quint/test/types/constraintGenerator.test.ts
```diff
@@ -115,4 +115,58 @@ describe('ConstraintGeneratorVisitor', () => {
 
     assert.sameDeepMembers(Array.from(errors.values()), [error])
   })
+
+  function testArityError(expr: string, location: string, message: string) {
+    const defs = [expr]
```
bugarela, line 120: This is called `expr` but is expected to contain the text for a `def` 
```suggestion
  function testArityError(def: string, location: string, message: string) {
    const defs = [def]
```

bugarela (approved):
Looks great, thanks for taking care of this!

quint/src/types/constraintGenerator.ts
thpani: Okay, but should we then update the docs on `buildErrorLeaf`?
It looks like `location` is more of an explanation.

quint/test/types/constraintGenerator.test.ts
shonfeder: Yes we should. I thought I had one but I must have hallucinated it ;) thanks.

quint/src/types/constraintGenerator.ts
shonfeder: So I understand the location here to be the context in which the thing is happening. That is inferred from the use I’ve seen (e.g. in the special constraints). But if it’s causing confusion for Thomas maybe a rename of the param or added docs is a good idea. I think it’s out of scope for this PR tho, iiuc, and can be addressed in a followup.

quint/src/types/constraintGenerator.ts
shonfeder: Nice comment. Thanks :)

quint/test/types/constraintGenerator.test.ts
shonfeder: It somehow escaped being committed but was in my buffer. Fixed now. Thanks!

shonfeder:
All comments here have been addressed. Thanks for the reviews!

Issue #1082, opened 2023-07-28
Add syntax for sum-types


shonfeder:
Add the syntax proposed in #1062, which requires:

- Two small extensions to the grammar for sum type defs and case analysis
- Two corresponding additions to the IR
- Tests confirming the correspondence of these two constructs

PR #1085, opened 2023-08-01
Refactor ToIrListener id generation

shonfeder:
This is a small refactor supporting #1082

Add a helper function to generate location IDs and register them into
the source map. These must always go together, so adding the two into
the same function saves a number of lines and makes the code safer since
it ensures we won't miss a needed addition to the map.

-------

### DIFF
--- a/quint/src/parsing/ToIrListener.ts
+++ b/quint/src/parsing/ToIrListener.ts
@@ -73,8 +73,7 @@ export class ToIrListener implements QuintListener {
     assert(this.exprStack.length === 0, 'expression stack must be empty')
     assert(this.identOrHoleStack.length === 0, 'parameter stack must be empty')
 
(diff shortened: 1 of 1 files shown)


thpani (approved):
Looks good!

Is the side-effect obvious enough to not reflect it in the name (e.g. `getIdAndUpdateSourceMap()` vs `getId()`)?

shonfeder:
Thanks for the review!

> Is the side-effect obvious enough to not reflect it in the name

Given that the method is private and commented, I feel it is nice and convenient to phrase it is I did here. The map updates is really an implementation detail happening behind the scenes, and there's no reason for the adder for a visitor method to think about it.

I could encapsulate the id gen and map update in a class, but seems a bit overkill here.

Issue #1081, opened 2023-07-28
Introduce stronger lexical rules for identifiers


shonfeder:
This follows from discussion in #1062 and discussion with @bugarela and @thpani in person.

Currently, type variables must be lower case and type constants must be upper case, but otherwise we allow any identifier to start with either lower or upper case letters.  In #1062 I discuss some of the benefits we could gain in our sum types syntax from introducing this difference across other constructs.

This issue addresses the more general issue that we allow an (imo) unhealthy level of leeway in all of our identifiers. 

### Cryptic identifiers

This spec is completely valid -- it parses and type checks:

```quint
module ___::Arst::t {
    type __a::FPRT::__ = { a::B::__::c : int }
    const X::B::C: __a::FPRT::__
    def X::b::d(__::__::__:__a::FPRT::__ ): __a::FPRT::__ = __::__::__
    def f(x: List[x::B]): x::B = head(x)
}

module ex {
    import ___::Arst::t(X::B::C = { a::B::__::c : 2 }) as A::B
    val x = A::B::X::B::C
}
```

### Escape hatch thru exotic operator

There are a number of weird edge cases we allow that would be ruled on at the very outer edge of our language if we enforced a more restrictive grammar. E.g.,

These are legal records

```quint
>>> Rec("{a: 1, b: 'foo'}", 3)
{ {a: 1, b: 'foo'}: 3 }
>>> Rec("                 ", 3)
{                  : 3 }
>>> Rec("()()()()()", 3)
{ ()()()()(): 3 }
```

These operations, that we should be able to exclude based on syntax alone, make it all the way thru our type system to be evaluated by the REPL:

```quint
>>> val x = "a"; {a: 1}.field(x)
static analysis error: error: [QNT000] Record field name must be a string expression but is name: x
Generating record constraints for Rec("a", 1),x
Generating constraints for x

val x = "a"; {a: 1}.field(x)
             ^^^^^^^^^^^^^^^


1
>>> Tup().item(100)
static analysis error: error: [QNT000] Operator expects at least one arguments but was given 0
Checking arity for application of Tup
Generating constraints for Tup()

Tup().item(100)
^^^^^


runtime error: error: Out of bounds, nth(99)
Tup().item(100)
^^^^^^^^^^^^^^^

<undefined value>
```

### Types defined in modules starting with a lower-case letter cannot be referenced with qualifiers. 

E.g., given

```quint
module foo {
  type T = int
}

module bar {
  import foo
[...]

Issue #17, opened 2021-11-14
Improve error messages by antlr


konnov:
The default messages by antlr are hardly readable:

```
examples/ReadersWriters/ReadersWriters.tnt:22:65 - error no viable alternative at input 'NumActorsdefToSet(s)=s.keys().map(i->s[i])defread(s)=s[1]=="read"defwrite(s)=s[1]=="write"valWaitingToRead=ToSet(waitingselectread)map(p->p._2)'
examples/ReadersWriters/ReadersWriters.tnt:22:57 - error extraneous input '->' expecting {'module', '{', '}', 'const', 'var', 'assume', 'type', 'import', 'val', 'def', 'pred', 'action', 'temporal', '(', '|', '[', '^', 'if', '&', STRING, BOOL, INT, 'and', 'or', 'iff', 'implies', 'subseteq', 'in', 'notin', '+', '-', '*', '/', '%', '>', '<', '>=', '<=', '!=', '==', '<-', IDENTIFIER}

```

For reference: syntax error, parsing, parser error

shonfeder:
Here are some useful references:

- Catching exceptions: https://github.com/antlr/antlr4/blob/master/doc/parser-rules.md#catching-exceptions
- Adding a custom error listener: https://stackoverflow.com/questions/18132078/handling-errors-in-antlr4
- Using "error alternatives": https://stackoverflow.com/a/58115099

The last seems very clean!

konnov:
I have tried [Peggy](https://peggyjs.org/). It is producing nice error messages, but it does not have a real lexer. This makes error messages not so nice. There is also [Ohmjs](https://ohmjs.org/). Let's see how fast I hit its limitations.

PR #1086, opened 2023-08-01
Require record fields to be simple identifiers

shonfeder:
When a `::` appears in a quint identifiers, I am referring to that identifier as
"qualified". E.g., `Foo::bar` is qualified, but `_ding` is not.

This takes care of one problem identified in #1081 by introducing a syntax error
a record field is a qualified identifier.

Along the way, the PR also gives further illustration of how we can address #17
by using error grammar rules. (There was already precedent for this in the
`errorEq` rule checking for the misuse of  `=` for equality). This shows how we
can move stronger syntactic checks into the grammar without sacrificing the
quality of our error messages.

E.g., errors for this new rule look like so:

```quint
>>> type r = {a: int, b::c::d: str}
syntax error: error: [QNT008] Identifiers in a record cannot be qualified with '::'. Found b::c::d.
type r = {a: int, b::c::d: str}
                         ^
```

To enable this improvement in the grammar, we had to refactor the lexical rule
for `IDENTIFIER`: without splitting this into a lexical rule for the
unqualified case and a grammar rule for the qualified, we end up with an
ambiguous lexical rule that makes parsing fail.

One incidental change is added to make the `reference` field in our error type
optional. This is implied by the fact that parser errors are errors that should
be represented with this structure, but we may not have a reference for them yet.


shonfeder:
The last two commits are generated/mechanical!

quint/src/generated/QuintListener.ts
```diff
@@ -1,6 +1,10 @@
 // Generated from ./src/generated/Quint.g4 by ANTLR 4.9.0-SNAPSHOT
 
 
+    // Used for forming errors
+    import { quintErrorToString } from '../quintError'
```
thpani, line 5: Unusual indent here. Should we dedent it in `Quint.g4`? I'm okay either way.

quint/src/generated/Quint.g4
```diff
@@ -214,6 +221,21 @@ operator: (AND | OR | IFF | IMPLIES |
...
+// An unqualified identifier that raises an error if a qualId is supplied
+simpleId[context: string]
+    : IDENTIFIER
+    | qualId {
+        const err = quintErrorToString(
+          { code: 'QNT008',
+            message: "Identifiers in a " + $context + " cannot be qualified with '::'. Found " + $qualId.text + "."
+          },
+        )
+        this.notifyErrorListeners(err)
```
thpani, line 235: I'm torn whether we should use `quintErrorToString` here at all (with all it's implications, making error fields optional), or simply pass a string as for `QNT006` above. WDYT?

thpani (approved):
Looks great!

I have an alternative suggestion to making the error type fields optional; not sure myself if I prefer one over the other.

quint/src/generated/Quint.g4
shonfeder: So my justification for bringing it in is that we want to put white a few more errors here probably, and using the type and function ensures we are using valid error ids, and if/when we overhaul the error system or add hints to tic etc, we will know to add them here as well. I do agree it’s more overhead, but already in this case it caught a typo on in my error if 😅.

If you feel strongly that we should stick to strings tho I’m ok with that. We could get @bugarela’s opinion as a tiebreaker! :)

regarding making the reference field optional, note that we already effectively treat it as optional (afaict), in that we allow references which aren’t ok our location map and then just print and unknown location.

thpani: > that we want to put white a few more errors here probably, and using the type and function ensures we are using valid error ids, and if/when we overhaul the error system or add hints to tic etc, we will know to add them here as well

sgtm 👍

quint/src/generated/QuintListener.ts
shonfeder: Oh yeah, I didn't look at the generated code! I think removing the indent won't hurt readability of the antlr, so I'll do that.

PR #1087, opened 2023-08-01
Add ADR 7 on flattening

### LINKED ISSUE #1071: Flattening - Meta issue
Current flattening has way too many problems, and I can't see a way to solve them in the current algorithm. Therefore, I'm working on a different approach with a two phase flattening. This should fix:
- https://github.com/informalsystems/quint/issues/1032
- https://github.com/informalsystems/quint/issues/836
- https://github.com/informalsystems/quint/issues/1040
- Broken scenarios described in https://github.com/informalsystems/quint/discussions/949

I think the outcome of this issue is an ADR describing the solution, and then I'll create other issues for blocks of implementation.

bugarela:
Hello :octocat: 

I wasn't sure if this should be an ADR or RFC, but I now think it should be recorded as an ADR.

Closes #1071 


doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+  4["(4) val c = C::A1::a"] --> 8
+  5["(5) pure val B::A1::N = 1"]
+  6["(6) pure val B::A1::a = B::A1::N"] --> 5
+  7["(7) pure val C::A1::N = 0"]
+  8["(8) pure val C::A1::a = C::A1::N"] --> 7
+```
+
+##### Override expressions
+
+In an instance statement like `import A(N=e).*`, `e` might depend on expressions defined in the current module. In this case, we need to ensure that the dependencies of `e` are copied to the new module created for the instance, otherwise `pure val myModule::A::N = e` won't resolve. To accomplish this, we use the actual `Flattener` (described below) to inspect only the expressions in the overrides and find any definitions they might depend on.
```
shonfeder, line 199: Having to copy values "both ways" -- both into the importing module and into the instantiated module -- sounds a bit tricky and error prone, esp. considering that the instantiated module will get copied back into the importing module, meaning well be duplicating the definition of `e`.  Do I understand correctly, that given

```quint
module A {
  const a : int
}

module B {
   val e = 42
   import A(a = e).*
}
``` 

the proposed approach for "instance flattening" will end up giving us the intermediate result

```quint
module A {
  const a : int
}

module B::A {
  pure val B::A::e 
  pure val B::A::a : int = B::A::e
}

module B {
   pure val e = 42
   import B::A.*
}
``` 

and then this will be finally turned on the flattening pass into

```quint
module B {
   pure val e = 42
   pure val B::A::e 
   pure val B::A::a : int = B::A::e
}
```

Is that right?

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+	 - We might be able to manipulate the table inside the flattener so this is not necessary, but it is not obvious how to do that. The main problem seems to be that, since the Instance Flattener replaced instance statements with import statements, the `importFrom` and `namespaces` fields of certain definitions have to be updated.
+7. Sort names topologically, since the added definitions might be out of order.
+	 - We might also avoid doing this if we add the new definitions in a proper order.
+
+### [Discussion] `FlatModule` and `FlatDef` types
+
+I have not tried this in the proof of concept, but I believe we need to improve our nomenclature and types. This is what we currently use:
+- An IR definition (`QuintDef`) is any top level definition, including vars, consts, imports, etc.
+- A flat definition (`FlatDef`) is a definition that is not a import, instance or export. All flat definitions have a `.name` field.
+- A flat module (`FlatModule`) is a module with only flat definitions
```
shonfeder, line 231: Just a thought: This looks like a round about way of breaking the language into two fragments:

- Core language, corresponding to `FlatDef`
- Module language, which just consists of the `import` and `export` constructs.

`const` is a bit tricky, because if a module includes a `const` it means it cannot be used without supplying core language value as parameters in an instantiation.

`FlatModule` is module that only uses the core language.

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+- A definition in the lookup table (`Definition`) is any `QuintDef` or a `QuintLambdaParameter`
+- Sometimes, we call `QuintDef` by 'unit' (in the parser and parsing functions such as `parseExprOrUnit`)
+
+I think we should do something like:
+- Call this high level definitions that encapsulate everything as 'statements' or 'declarations'.
+- `QuintModule` has a list of 'statements' or 'declarations'.
```
shonfeder, line 237: I suggest just

```suggestion
- `QuintModule` has a list of 'declarations'.
```

just cause of the baggage with `statement`.

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+
+## 2. Context
+
+Our current strategy for flattening has intrinsic problems, and some issues that popped up cannot really be solved with the current approach. The first of those issues, and perhaps the one that better flags the problem in the current implementation, is [Flatenning breaks when importing by name a definition that depends on others #836](https://github.com/informalsystems/quint/issues/836).
+
+The existing flattener takes import/instance/export statements and replaces them with a copy of the definitions from that module, assigning new ids to those definitions and its inner IR components and adding namespaces when necessary. With this, there are two possible scenarios:
+
+1. All imports, instances and exports include all definitions (i.e. there is no `import A.foo`). In this case, we will end up with a module with all of the definitions from all used modules.
+2. There is some statement refering to a specific definition like `import A.foo`. In this case, the other definitions from `A` won't be copied, and if `foo` depends on something else, the flattened module won't be valid.
```
shonfeder, line 18: Could we solve this by doing inlining on expressions, rather than manipulating modules namespaces? E.g.,

```
module Foo {
   pure val e = 1
   def foo() = e + 1
}

module Bar {
   import Foo.foo
   def bar() = foo()
}
```

Then, while flattening, when we encounter the need to import `Foo.foo` could we just inline foo first to `foo() = 1 + 1`, then add this inlined `foo` into the importing bar?

```
modules Bar {
  def foo() = 1 + H
  def bar() = foo()
}
```

shonfeder (commented):
Thanks for the write up to document the development of your design process so far!

I have some questions and a few small suggestions. Overall I find this quint clarifying!

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+When collecting names from import/export/instance statements, we should be able to keep track of the namespaces to add to the definition in order to refer back to it uniquely in a flattened module. The following method describes which namespaces to add depending on `def`, returning a list of namespaces to be accumulated starting from the namespace closer to the actual name (i.e. namespaces `['a', 'b', 'c']` applied to name `foo` will result in `c::b::a::foo`).
+
+```typescript
+private namespaces(def: QuintImport | QuintInstance | QuintExport): string[] {
+  if (def.kind === 'instance') {
+    return compact([def.qualifiedName ?? def.protoName, this.currentModuleName])
+  }
+
+  return def.defName ? [] : [def.qualifiedName ?? def.protoName]
+}
```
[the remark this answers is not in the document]
thpani: same here – I think we could do without the code, the description above is quite clear

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+Imports are the simplest scenario, since they only copy definitions, possibly adding a namespace to them:
+1. `import A.*`, `import A.foo` don't add namespaces
+2. `import A` adds `A::` namespace
+3. `import A as MyA` adds `MyA` namespace
+
+##### Namespaces for instances
+
+For instances, we need to ensure uniqueness, since the same names in the instanced module can assume different values if the constants are different. Therefore, we add `this.currentModuleName` to the namespace, differentiating possible instances with the same name from different modules. For the following cases, suppose the statements are inside the `myModule` module.
+1. `import A(N=1).*`  adds a `myModule::A` namespace
+2. `import A(N=1) as A1` adds a `myModule::A1` namespace
```
thpani, line 77: Not clear to me how the examples here play out. If I `import A(N=1).*`, are the names from `A` available inside `myModule` in an unqualified version?
* If yes, is this because there is always an implicit `myModule::` qualification?
* If no, I'm afraid that this asymmetry compared to "regular" imports may be confusing to users.

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+
+Each instance statement is replaced with an import statement and generates a new module. The new module has the constants replaced with `pure val` definitions (from the overrides in the instance statement), and all its names are fully-qualified, constructed with the previously defined namespace algorithm. The new import statement, replacing the instance statement, is a unqualified import of the new module.
+
+The names in the module containing the instance statement also have to be updated with a namespace, since they might refer to definitions that now include a namespace. We look at each name in the module and, if it refers to a definition that comes from an instance, (that is, its `importFrom` field has an instance), then we replace the name with a namespaced version of it (using the `namespaces` field). For this to work properly, any existing `import` and `export` statements should have been flattened before the instance flattener is run, ensuring that the definitions referred to by these names already have their proper namespaces.
```
thpani, line 123: Probably connected to my question above – I don't follow why this is necessary.
Could you give an example to illustrate?

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+The instance flattener will create two new modules: `B::A1` and `C::A1`, with new ids for the definitions inside it.
+```bluespec
+module A {
+  const N: int // id 1
+  val a = N // id 2
+}
+
+module B::A1 {
+  pure val B::A1::N = 1 // id 5
+  pure val B::A1::a = B::A1::N // id 6
```
thpani, line 156: Maybe I don't have full context on how we resolve qualified names –
is the FQN of `N` `B::A1::N` or `B::A1::B::A1::N`?

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+}
+
+module C::A1 {
+  pure val C::A1::N = 0 // id 7
+  pure val C::A1::a = C::A1::N // id 8
+}
+
+module B {
+  import B::A1.*
```
thpani, line 165: This is `import A(N=1) as A1` above, why does it become a `.*`?

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+  val b = B::A1::a // id 3
+}
+
+module C {
+  import C::A1.**
```
thpani, line 170: double asterisk?

```suggestion
  import C::A1.*
```

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+
+This deep namespace-adding procedure is already done in the current implementation of flattening (see [`addNamespaceToDef`](https://github.com/informalsystems/quint/blob/main/quint/src/flattening.ts#L276C4-L276C4)), with the only difference being that we don't want to generate new ids in this case (we only create new ids in the Instance Flattener).
+
+Although the Instance Flattener requires imports and instances to be flattened, we can't flatten definitions that come from instances before running the instance flattener, since that would copy definitions that would potentially be replaced by new definitions in new modules created by the instance flattener. Therefore, we need to run the flattener twice: first not copying definitions that come from instances (checking the `importedFrom` field once again), and then, after the instance flattener, flattening anything that remains to be flattened.
```
thpani, line 209: I don't follow here: you say we cannot flatten before instance-flattening, but why can't we instance-flatten and then flatten?

What happens if we have a instance->import->instance->import->... chain? Does this degenerate into a recursive procedure?

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Overall flattening process
+
+The flattening process is run completely for each module before proceeding to flatten the next module. The modules are assumed to be topologically sorted. The need for running this by module is that we do remove the import statements after we resolve all dependencies, but we can only ensure that all dependencies were flattened if the depending modules were completely flattened before.
+
+The whole flattening process is composed by the following steps:
+1. Run the Flattener, ignoring definitions that come from instances
+2. Remove `import` and `export` statements
+3. Run the Instance Flattener
+4. Resolve names for the module and its dependencies, obtaining a new lookup table
```
thpani, line 219: What keeps us from producing the updated lookup table as a by-product of flattening?

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+5. With the new table and modules, run the Flattener again, this time for all definitions
+6. Resolve names again for all modules (using the flattened version for those that are already flattened, and the original version for those that are not). The resulting table is used as the starting point of flattening for the next module, or returned as the final table if this is the last one.
+	 - We might be able to manipulate the table inside the flattener so this is not necessary, but it is not obvious how to do that. The main problem seems to be that, since the Instance Flattener replaced instance statements with import statements, the `importFrom` and `namespaces` fields of certain definitions have to be updated.
```
thpani, line 222:
```suggestion
	 - We might be able to manipulate the table inside the flattener so this is not necessary, but it is not obvious how to do that. The main problem is that, since the Instance Flattener replaced instance statements with import statements, the `importFrom` and `namespaces` fields of certain definitions have to be updated.
```

thpani, line 222: > [...] the `importFrom` and `namespaces` fields of certain definitions have to be updated.

This does not sound like the root cause 😄 What keeps us from updating these inside the flattener?

thpani (commented):
Thanks for the clear write-up!

I have a few questions – bear with me while I learn these things 😅

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+
+Although we don't need or want to change anything in the name resolution phase for flattening, we do collect some additional metadata in this phase to be used later in flattening
+
+#### `LookupTable` with full quint definitions
+
+Our current lookup table only stores a projection of the quint definition: id (called reference), name and type annotation (if present). There is no strong reason for that, only a premature optimization and the fact that the lookup table values might also be lambda parameters, which are not quint definitions.
+
+In flattening, we need to manipulate and reorganize definitions, and is much better for performance and readability if the definitions are in the lookup table, as oppose to having to scan the modules for a definition.
+
+Therefore, we change the lookup table to have either a `QuintDef` or a `QuintLambdaParam` as its value, as long as two additional fields (`namespaces` and `importedFrom`) described below.
```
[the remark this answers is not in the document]
bugarela: I think I merged "as well as" and "along with" in my brain and typed this out :laughing:

doc/adr007-flattening.md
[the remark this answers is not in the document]
bugarela: It describes the same thing as the next 3 headers, that is, which namespaces should be added to defs coming from a given import/instance/export statement. I guess we could do without it, but I also think it can be a good summary of this algorithm, and summaries are great.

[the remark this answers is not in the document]
bugarela: Perhaps I can move it to after the entire description and clarify that this represents the thing that was already described.

doc/adr007-flattening.md
```diff
@@ -0,0 +1,255 @@
...
+  8["(8) pure val C::A1::a = C::A1::N"] --> 7
+```
+
+##### Override expressions
+
+In an instance statement like `import A(N=e).*`, `e` might depend on expressions defined in the current module. In this case, we need to ensure that the dependencies of `e` are copied to the new module created for the instance, otherwise `pure val myModule::A::N = e` won't resolve. To accomplish this, we use the actual `Flattener` (described below) to inspect only the expressions in the overrides and find any definitions they might depend on.
+
+### Flattener
+
+The flattener adds missing definitions to a module by recursively fetching the lookup table and copying the referenced definition, then looking into that definition for any other names and doing the same.
```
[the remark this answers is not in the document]
bugarela: Missing as in it is not on the module where it is needed. For example:
```bluespec
module B {
  import A.*
  val b = foo + 1
}
```

`foo` is missing a corresponding definition inside of B. Since we know this module was already name resolved, `foo` must be defined in `A`, and we can actually fetch the definition for `foo` from the lookup table and copy it over to `B`.

I'll clarify this in the text as well.

doc/adr007-flattening.md
bugarela: Yes, I believe that this is the case.

(I'm assuming you meant `pure val B::A::e = 42` instead of only `pure val B::A::e` since that's not even a valid definition)

doc/adr007-flattening.md
bugarela: Well, even in that case, we would need a general type like `QuintDef` to encapsulate both core and module language, at least in order to topologically sort them, as module language statements might depend on core language defs (i.e. `import A(N = foo)` where `foo` is a core language def) and core language statements might depend on module language statements (as in referring to imported definitions).

So, if we need a general type, we'd end up with three types. Something like
```typescript

File: doc/adr007-flattening.md (hunk 4 above)
Comment (bugarela) [reply doc/adr007-flattening.md:18]: Well, for the specific case of importing one definition by name, I think we could. But for the general case, I don't think we want to do inlining on expressions, and the general solution for manipulating modules namespaces also works for this.

I don't know why exactly, but I've been trying to stay away from inlining on expressions when working on this. I think @konnov said something about not doing inlining on expressions, but I don't remember if it was something like "we shouldn't do it" or "we don't need it".

File: doc/adr007-flattening.md (hunk 6 above)
Comment (bugarela) [reply doc/adr007-flattening.md:77]: I think you are mixing things up. Name "availability" is part of name resolution, and yes, `A` will be available inside `myModule` unqualified. This is what the user sees. Then, after the user got all their names correctly, the modules will go through flattening, which might add namespaces to both the definition and its references. The user should never see this additional namespaces, and they are only meant to disambiguate names in the flattened module, used by the compiler and apalache.

File: doc/adr007-flattening.md (hunk 7 above)
Comment (bugarela) [reply doc/adr007-flattening.md:123]: Yes! So, this happens "after"[^1] we generate a new module for each instance and replace the instances with imports. 

Consider the original modules (is the same from the example section but without `C` to simplify)

```bluespec
module A {
  const N: int 
  val a = N
}

module B {
  import A(N=1) as A1
  val b = A1::a
}
```

Imagine we've created the new module (`B::A1`) for the instance statement `import A(N=1) as A1` in `B`, and replaced that statement with `import B::A1.*`:

```bluespec
module A {
  const N: int
  val a = N
}

module B::A1 {
  pure val B::A1::N = 1
  pure val B::A1::a = B::A1::N
}

module B {
  import B::A1.*
  val b = A1::a
}
```

Now we have a problem in the last definition here: `val b = A1::a`, because `A1::a` doesn't exist anymore, and the name for that is `B::A1::a`, so we have to fix it. This is what the paragraph is describing.

One approach that might seem like a good idea looking at this is what I thought about in the quint meeting last week (and also when I was first implementing this). Don't add the namespaces to the definitions in the new module, and import them with a qualifier: 

```bluespec
module A {
  const N: int
  val a = N
}

module B::A1 {
  pure val N = 1
  pure val a = N
}

module B {
  import B::A1 as A1
  val b = A1::a
}
```

That looks simpler here, but it results in name collisions in some of our examples. 

[^1]: It is actually at the same time, but imagine it's after since this is a consequence of the other thing

doc/adr007-flattening.md
bugarela: In theory it is the latter, but we don't actually use FQN for anything right now or have anything like that in the implementation.

doc/adr007-flattening.md
bugarela: instances always become `.*` because we already take care of the namespaces inside the new module. My example in https://github.com/informalsystems/quint/pull/1087#discussion_r1284418538 might make this clearer.

doc/adr007-flattening.md
bugarela: > but why can't we instance-flatten and then flatten?

Because the Instance Flattener requires imports and instances to be flattened.

> Does this degenerate into a recursive procedure?

It doesn't. One round of flattener -> instance flattener -> flattener is enough. That's because the first round of flattening will get rid of all imports and exports, leaving us with only instances. The instance flattener will then replace the instances with imports, leaving us with only imports. The last round of flattener gets rid of the remaining imports.

We could maybe avoid the last flattening round if, in the instance flattener, instead of generating an import statement, we did some inner flattening as well. But I think that would only complicate the code and result in no benefit. With the current approach, there is a clearer (or at least less unclear) separation of concerns between the flattener and the instance flattener.

doc/adr007-flattening.md
bugarela: My cognitive abilities, mainly :laughing:. I think the same note I made for item (6) is valid here, with some added difficulty: step (3) adds new ids, so we would have to update all lookup table pointers involving those ids.

My reasoning is: we need to have proper names anyway since apalache has it's own name resolution thing (afaiu from your's and Igor's response when I asked about that need). And, if we have the names correct, than all we have to do to get our lookup table correct is run the name resolver.

In another direction, we could just get our lookup table to be right, and then ensure we have proper (unique) names by ensuring each id in the lookup table has a different name. I tried doing that, but is was much easier for me to reason about modules and names (which you can print and visualize) than with lookup table pointers.

doc/adr007-flattening.md
bugarela: Ah, I didn't mean they were the root cause. What I wanted to say is:
"I think we shouldn't need to run the resolver again here, we just have to update the `importFrom` and `namespaces` inside the flattener as we manipulate the definitions. However, I didn't try to do that because this is working and therefore is good enough for the proof of concept" :sweat_smile: 

That being said, I'm not sure how much I should pursue these improvements while working on the real implementation. Perhaps I should timebox some time for each of them.

bugarela: I haven't validated this hypothesis, so I'd rather leave it as "seems to be"

doc/adr007-flattening.md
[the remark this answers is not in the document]
bugarela: I added a sentence defining what "missing" means in this context

doc/adr007-flattening.md
bugarela: I added a note to this point

doc/adr007-flattening.md
bugarela: I rephrased this bullet point

doc/adr007-flattening.md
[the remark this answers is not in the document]
thpani: Yeah, I think we need the description first, otherwise the code snippet is out of context.

doc/adr007-flattening.md
```diff
@@ -0,0 +1,260 @@
...
+
+In flattening, we need to manipulate and reorganize definitions, and it is much better for performance and readability if the definitions are in the lookup table, as oppose to having to scan the modules for a definition.
+
+Therefore, we change the lookup table to have either a `QuintDef` or a `QuintLambdaParam` as its value, as well as two additional fields (`namespaces` and `importedFrom`) described below.
+
+#### `namespaces`
+
+When collecting names from import/export/instance statements, we should be able to keep track of the namespaces to add to the definition in order to refer back to it uniquely in a flattened module. The following method describes which namespaces to add depending on `def`, returning a list of namespaces to be accumulated starting with the innermost namespace and ending with the outermorst (i.e. namespaces `['a', 'b', 'c']` applied to name `foo` will result in `c::b::a::foo`).
```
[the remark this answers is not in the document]
thpani: I'm not actually sure where this `namespaces` field goes. Inside each def? Or is it a separate top-level field in the lookup table?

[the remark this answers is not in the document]
thpani: I have a suggestion for the entire paragraph below, so maybe skip ahead.

doc/adr007-flattening.md
thpani: > That looks simpler here, but it results in name collisions in some of our examples.

Ah – I guess that's the same thing that confused me here: https://github.com/informalsystems/quint/pull/1087#discussion_r1285617259

But I trust you if you say it's not possible this way.

doc/adr007-flattening.md
thpani: fair enough

thpani: > However, I didn't try to do that because this is working and therefore is good enough for the proof of concept

Oh sure, just wanted to leave a trace in the design doc :)

doc/adr007-flattening.md
bugarela: First of all, the module names need to be qualified in order to differentiate modules generated for different instances (i.e. `import A(N=1).*` in module `Foo` vs `import A(N=2).*` in module `Bar`, etc). So that needs to be qualified for sure.

As for the definitions: In theory (as in I haven't tried this fully to know that aren't more complications than I see), we could avoid qualifying the definition names here. We do qualify them to simplify the process of guaranteeing no conflicts in the flattened module. That's because these definitions will be imported later (as the instance statement was replaced with an import one), and this approach with the qualified names makes this very straight forward because the `namespaces` field doesn't need further manipulation.

If we didn't qualify the definitions here, we would also have to change the import statement produced from the instance statement, and that would make it so the `namespaces` field generated from the instance statement (during name collection) is no longer compatible with the new import statement.

My approach: `import A(N=1).*` becomes `import myModule::A.*` (no namespaces need to be added in neither)
Other approach: `import A(N=1).*` becomes `import myModule::A` (qualified import). We would have to update the `namespaces` field for the definitions inside `myModule::A` to contain `['A', 'myModule']`. This way, when the import statement is processed by the last flattening round, the flattener will add the namespaces and the definitions in the flattened module would be qualified like `myModule::A::foo`.

The end result in the flattened module is the same. Writing this down makes me realize that this other approach makes more sense than I initially thought. I think I'll try it out and see if it makes sense to go with it (in which case I'll update the ADR). Thanks for pointing it out.

PS: There is additional complexity to these approaches when the instance is qualified i.e. (`import A(N=1) as A1`), so I picked a simpler scenario to explain. I actually think the "Other approach" might be better for reducing that complexity.

thpani: Thanks for walking me through this! 🙏

Looks like neither way is completely friction-less, so picking the one easiest to implement makes a lot of sense!
Let's update the ADR with a note once we know which way we went – seems like a decision that's good to persist for future reference :)

thpani (approved):

doc/adr007-flattening.md
bugarela: This is a consequence of the fact that we fully qualify the names inside `B::A1`. Since the names are already qualified inside the module, we can always import with `.*`.

bugarela:
I believe I've added explanations to everything that you pointed out was lacking. I'll get this merged :smile:

PR #1088, opened 2023-08-02
Add parsing of sum type declarations

shonfeder:
Addresses the type declaration part of #1082.

This adds support for the sum type declaration syntax proposed in #1062.

The most interesting part of these changes is the addition of the IR constructs
to represent the declaration of sum types. This includes a specialized 
variant of the row types allowing us to statically enforce that the rows of a
variant type cannot be open. See the discussion on #1062 for the motivation for
this design decision.

Note that we do not add a parsing rule to `type` for sum types. This reflects
the fact that sum types can only be referenced by their defined name: anonymous
sum types are not supported.

Note that this PR is into the `sf/sum-type` feature branch.

- [x] Tests added for any new code
- [ ] Documentation added for any new functionality
- [x] Entries added to the respective `CHANGELOG.md` for any new functionality
- [ ] Feature table on [`README.md`](../README.md#roadmap) updated for any listed functionality
- [x] Rebase once #1086 is merged in

PR #1089, opened 2023-08-03
Add parsing of sum type declarations (#1088)

shonfeder:
This just merges in #1088 to `main`.

The initial plan was to use this branch to stage the sum-type work until more of the team was back in office and we could get additional eyes to review it. However, given the OOO overlaps over the coming risk @thpani and I reasoned that it would be better to merge in these changes to the main branch:

- This will prevent the risk of merge conflicts over the next weeks when I go OOO
- It poses no risk to users expecting stability, as we have not announced or released these features, and at this stage users cannot make any use of the sum types yet.
- The only aspect of the implementation lingering in discussions of #1062 is details of surface syntax, and this will only take a few lines to change.


thpani (approved):

quint/src/quintTypes.ts
```diff
@@ -74,6 +74,24 @@ export interface QuintRecordType extends WithOptionalId {
...
+  }
+}
+
+export function isTheUnit(r: QuintType): Boolean {
+  return r.kind === 'rec' && r.fields.kind === 'row' && r.fields.fields.length === 0 && r.fields.other.kind === 'empty'
+}
+
+export interface QuintSumType extends WithOptionalId {
+  kind: 'sum'
+  fields: ConcreteFixedRow
```
bugarela, line 92: This is super smart, I love this new `ConcreteFixedRow` type!

bugarela (approved):
I agree that we should merge this! It's for sure much better to have it on main preventing anything to rot away (as in accumulating conflicts and other problems), and I don't see any risk here (besides a user typing a proper sum type and failing to get an error, which is far from a real problem).

shonfeder:
Awesome. Thanks for weighing in!

PR #1092, opened 2023-08-04
Fix non-exhaustive switches

### LINKED ISSUE #196: Centralize mapping of operators parsed from symbols
TNT allows some built-in operators to be represented with symbols and then translates them at parsing (i.e. `+` to `iadd`, `{ key: value }` to `rec(key, value)`). These operator names are mostly internal, and they are referenced to in many different parts of the code base (parsing, default definitions for name resolution, effect signatures, type signatures, simulator compiling) which is a problem since we can change the name in some of these places but not others. Also, I keep switching them up (i.e. `rec` vs `record`) so that's annoying.

We should have a single module defining all of those names so we can:
1. Have a single place to update when we want to change them
2. Have static guarantees that we are not referencing an unexisting operator
    - This should also be true for built-in operators in general, not just the ones with symbol representation

shonfeder:
While working on the second part of #1082 I realized that there are a lot of places
where we had created type checking blind spots due to our use of mutation in
control flow via switches that really require exhaustive analysis of data types
they are operating on. This refactoring pass fixes all the places I could find
readily. This should make it a bit easier to finish #1082 correctly and should
make our code base a bit more maintainable in general.

Please see the addition to our `CONTRIBUTING.md` in the first commit for a
detailed explanation and rationale.

Also closes #196, tho we will probably need followup to ensure the type is being
used everywhere it should be.

## Reviewing

Please review ignoring whitespace. It's a considerably smaller diff that git is reckoning, and the bloat is just from whitespace changes required by restructuring some conditionals.


quint/src/IRTransformer.ts
```diff
@@ -410,26 +427,30 @@ function transformExpression(transformer: IRTransformer, expr: ir.QuintEx): ir.Q
...
+          throw new Error('Let operator definition transformed into non-operator definition')
+        }
+
+        newExpr.opdef = opdef
+        newExpr.expr = transformExpression(transformer, newExpr.expr)
+
+        if (transformer.exitLet) {
+          newExpr = transformer.exitLet(newExpr)
+        }
       }
```
shonfeder, line 449: This was just a formatting change.

quint/src/cliCommands.ts
```diff
@@ -513,34 +514,21 @@ export async function runSimulator(prev: TypecheckedStage): Promise<CLIProcedure
...
-      case 'violation':
-        msg = 'Invariant violated'
-        break
-
-      case 'failure':
-        msg = 'Runtime error'
-        break
-
-      default:
-        msg = 'Unknown error'
```
shonfeder, line 535: This case was redundant b/c it's checked for in the `if` on line 483.

quint/src/effects/simplification.ts
```diff
@@ -57,35 +58,35 @@ export function simplify(e: Effect): Effect {
...
-        }
-      })
-
-      if (unionEntities.length > 0) {
-        const entities =
-          vars.length > 0 ? unionEntities.concat({ kind: 'concrete', stateVariables: vars }) : unionEntities
-        return entities.length > 1 ? { kind: 'union', entities: entities } : entities[0]
-      } else {
-        return { kind: 'concrete', stateVariables: vars }
+  if (entity.kind == 'union') {
```
shonfeder, line 61: I changed this from a switch that just had one branch to a conditional.

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -387,514 +392,519 @@ export class CompilerVisitor implements IRVisitor {
...
-      case 'range':
-        this.applyFun(app.id, 2, (start, end) => {
-          const [s, e] = [Number(start.toInt()), Number(end.toInt())]
-          if (s <= e) {
-            const arr: RuntimeValue[] = []
-            for (let i = s; i < e; i++) {
-              arr.push(rv.mkInt(BigInt(i)))
+    if (!ir.isQuintBuiltin(app)) {
+      this.applyUserDefined(app)
+    } else {
```
shonfeder, line 397: The only change here was adding a conditional using a type guard to check whether we have a builtin before dispatching the builtins.

quint/src/types/substitutions.ts
```diff
@@ -51,49 +52,44 @@ export function compose(table: LookupTable, s1: Substitutions, s2: Substitutions
  *          given type
  */
 export function applySubstitution(table: LookupTable, subs: Substitutions, t: QuintType): QuintType {
-  let result = t
```
shonfeder, line 54: Swapping out the mutable assignment for returning computed data.

PR #1093, opened 2023-08-05
Add syntax for constructing and eliminating sum type values

shonfeder:


quint/src/generated/Quint.g4
```diff
@@ -57,7 +57,7 @@ typeDef
     | 'type' typeName=qualId '=' '|'? typeSumVariant ('|' typeSumVariant)*  # typeSumDef
     ;
 
-// A single variant case in a sum type definition.
+// A single variant case in a sum type definition or match statement.
```
shonfeder, line 60:
```suggestion
// A single variant case in a sum type definition or match statement.
```

quint/src/flattening.ts
```diff
@@ -343,23 +344,46 @@ class Flatenner {
...
+      case 'variant':
+        return {
+          ...expr,
+          label: this.namespacedName(name, expr.label),
+          expr: expr.expr ? this.addNamespaceToExpr(name, expr.expr) : expr.expr,
+        }
+    }
+  }
+
+  // Gives us a more narrowly typed transformation on lambdsas, so we can use the same functionality for
```
shonfeder, line 376:
```suggestion
  // Gives us a more narrowly typed transformation on lambdas, so we can use the same functionality for
```

quint/src/parsing/ToIrListener.ts
```diff
@@ -859,11 +887,55 @@ export class ToIrListener implements QuintListener {
...
+  //    | ...
+  //    | Variantn(varn) => exprn
+  //    | _ => exprm              // A wildcard match, acting as a catchall
+  // }
+  exitMatchSumExpr(ctx: p.MatchSumExprContext) {
+    const matchId = this.getId(ctx)
+    // We will have one expression for each match case, plus the
+    const exprs = popMany(this.exprStack, ctx._matchCase.length + 1)
+    // The first expression is the one we are matching on
+    // the syntax rules ensure that at least this expression is given
```
shonfeder, line 907:
```suggestion
    // The first expression is the one we are matching on.
    // The syntax rules ensure that at least this expression is given.
```

quint/src/quintIr.ts
```diff
@@ -248,10 +248,43 @@ export interface QuintLet extends WithId {
...
+
+/** A variant value inhabiting a sum type */
+export interface QuintVariant extends WithId {
+  kind: 'variant'
+  /** the varian't tag, e.g., the `A` in `A(n)` */
+  label: string
+  /** the expression injected into the sum type  */
+  expr?: QuintEx
+}
+
```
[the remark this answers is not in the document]
shonfeder: In general:

> we have managed to keep the number of cases in QuintEx at minimum. This required us to introduce "special" low-level operators like Rec(...) and field(...) instead of having distinct cases in the IR. Although the normal users should not use these operators, they simplify IR manipulation at large.

But what is the gain here? I've enumerated some of the costs here:

- https://github.com/informalsystems/quint/blob/65f4877aba142b3c7f5538920bb568a69c3f8244/doc/rfcs/rfc001-sum-types/rfc001-sum-types.org#drop-the-exotic-operators
- https://github.com/informalsystems/quint/issues/1081

In the case of these kind of special constructs, any savings on the cases on `QuintEx` are paid for directly in the number of cases in the operator application logic, only we lose structure along the way and up having to perform unnecessarily validations and computations to reconstruct that lost data.

This PR is meant in part to show the benefit of dropping these exotic operator in favor a richer and more helpful representation in the IR.

But if you could help me understand where we actually see simplifications instead of complexity, that could help me move bast my aversion here.

In particular:

Wouldn't adding the `varaiant` operator require support for open rows in sum-types, which we agreed (as per your preference) we did not want to expose in #1062?

We will need special cases for most operator applications to handle `match`, but we will lose the typing info and structure provided here.

PR #1105, opened 2023-08-10
Store the whole def in `Defintion` and introduce `QuintDeclaration`

### LINKED ISSUE #1099: Store `QuintDef` and `QuintLambdaParameter` in the lookup table
Change the lookup table data structure to store the whole `QuintDef` or `QuintLambdaParameter` in its values.
- This will break integration with apalache. To avoid that, we should either:
  1. Cleanup the produced lookup table in the JSON output so it contains a field called `reference` in the values
  2. Release an apalache version that accepts a lookup table with `id` fields in the values.
- This also requires a simple update to the `typeAliasInliner`, which now has to operate over

bugarela:
Hello :octocat: 

Closes #1099 
Depends on https://github.com/informalsystems/apalache/pull/2696

This PR contains two related structural changes:
1. `LookupTable` now stores a whole definition instead of a projection of it
  - This breaks integration with apalache because the field we use to project as `reference` is now the original definition field called `id`.
  - This is needed by the new flattener and described in [ADR 7](https://github.com/informalsystems/quint/blob/main/doc/adr007-flattening.md#41-changes-to-name-collection)
  - The entries also have additional fields required by the flattener (`namespaces` and `importedFrom`), but those are not filled yet since this PR is already big. The PR includes only the type definition changes and its immediate consequences.
2. `QuintDef` no longer includes `QuintImport`, `QuintExport` and `QuintInstance`. Those are now part of `QuintDeclaration` only.
  - This breaks integration with apalache because the module field `defs` is now called `declarations` (for internal consistency).
  - This is not needed by the flattener, but the restructure came up during the design and is discussed in [ADR 7](https://github.com/informalsystems/quint/blob/main/doc/adr007-flattening.md#discussion-flatmodule-and-flatdef-types). It turned out to be much more beneficial than I anticipated.

I hope the PR is reviweable. It contains a lot of mechanical changes, specially renaming. I added comments to the most meaningful parts, and everything else is a direct consequence of those parts. I could have done this in two parts, but then I'd have to handle the apalache dependency twice.

CI checks should go through once my changes to apalache are released.


quint/src/names/base.ts
```diff
@@ -20,21 +21,37 @@ import { QuintType } from '../ir/quintTypes'
...
+   * is copied from one module to another with a qualified name. Ordered from
+   * innermost to the outtermost. */
+  namespaces?: string[]
+  /* importedFrom` is a reference to the import/instance/export statement that
+   * originated the definition, when the definition was copied from another
+   * module. */
+  importedFrom?: QuintImport | QuintInstance | QuintExport
+  /* `typeAnnotation` is the type annotation of the definition, if it has one.
+   * Some types in `QuintDef` already have a `typeAnnotation` field. This
+   * ensures that this field is always accessible */
```
bugarela, line 49: This is the main change regarding flattening. The introduction of `QuintDeclaration` makes it so that `QuintDef` is more restrictive than before, so this can never be an import/export/instance (which wouldn't make sense anyway)

quint/src/ir/quintIr.ts
bugarela, line 1: This file contains the core change. Everything else is either a consequence of this or a consequence of the change to the `Definition` type inside `names/base.ts`

quint/src/static/callgraph.ts
```diff
@@ -167,42 +176,49 @@ export class CallGraphVisitor implements IRVisitor {
...
-      this.graphAddAll(def.id, Set([importedModule.id]))
+      this.graphAddAll(decl.id, Set([importedModule.id]))
     }
   }
 
-  enterExport(def: QuintExport) {
-    const key = mkImportKey(this.currentModuleId, def.protoName)
+  exitInstance(_decl: QuintInstance) {
+    this.stack.pop()
+  }
```
bugarela, line 198: @konnov can you take a look at this part? I could make it work, but I'm not 100% sure it is sound.

Basically, the changes impacting this are:
1. This uses `exitDef`, which no longer include imports/exports/instances
2. If we change it to `exitDecl`, it is not the same as before, since a nested definition is not a declaration (that is, opdefs inside `QuintLet`).

So I kept `exitDef` and added these `enterInstance` and `exitInstance` to handle dependencies coming from expressions in overrides. IIUC, there is no need to do the same for imports and exports.

Generally speaking, I can see there is a small overhead, in some cases, arising from the fact that we no longer have an `exitSomething` that includes both imports/exports/instances and nested defs. But for the most cases, I think this is a good thing, and it is helpful to have explicitly different treatment for declarations and definitions.

quint/src/flattening.ts
```diff
@@ -249,24 +252,15 @@ class Flatenner {
...
     return defsToFlatten.map(protoDef => this.copyDef(protoDef, qualifiedName))
   }
 
-  private copyDef(def: QuintDef, qualifier: string | undefined): FlatDef {
-    if (!isFlat(def)) {
-      throw new Error(`Impossible: ${definitionToString(def)} should have been flattened already`)
-    }
-
-    const defWithQualifier = qualifier ? addNamespaceToDefinition(def, qualifier, this.currentModuleNames) : def
+  private copyDef(decl: QuintDef, qualifier: string | undefined): QuintDef {
```
thpani, line 260: Should the first parameter be called `def`?

quint/src/ir/IRTransformer.ts
```diff
@@ -271,6 +273,64 @@ export function transformType(transformer: IRTransformer, type: t.QuintType): t.
   return newType
 }
 
+/**
+ * Transforms a Quint declaration with a transformer, invoking the correspondent function for each
```
thpani, line 277:
```suggestion
 * Transforms a Quint declaration with a transformer, invoking the corresponding function for each
```

quint/src/ir/IRprinting.ts
```diff
@@ -25,10 +25,70 @@ import { typeSchemeToString } from '../types/printing'
...
 export function moduleToString(quintModule: QuintModule): string {
-  const defs = quintModule.defs.map(d => definitionToString(d)).join('\n  ')
+  const defs = quintModule.declarations.map(d => declarationToString(d)).join('\n  ')
   return `module ${quintModule.name} {\n  ${defs}\n}`
 }
 
+/**
+ * Pretty prints a declaration. Includes a type annotation if the definition is
+ * annotated, or if a type is provided. The annotation is preferred over the
+ * type.
```
thpani, line 35:
```suggestion
 * Pretty prints a declaration. Includes a type annotation if the definition is
 * annotated, or if a type is provided. A type annotation, if present, takes
 * precedence over a type provided as argument to this function.
```

quint/src/names/base.ts
```diff
@@ -20,21 +21,37 @@ import { QuintType } from '../ir/quintTypes'
 export type DefinitionKind = 'module' | 'def' | 'val' | 'assumption' | 'param' | 'var' | 'const' | 'type'
 
 /**
- * A minimal representation of a QuintDef, to be stored in `LookupTable` and
- * `DefinitionsByName`.
+ * A definition to be stored in `DefinitionsByName` and `LookupTable`. Either a `QuintDef`
+ * or a `QuintLambdaParameter`, with additional metadata fields
+ *
+ * A definition can be hidden, meaning it
```
thpani, line 27: There's something missing here :)

quint/src/names/base.ts
```diff
@@ -20,21 +21,37 @@ import { QuintType } from '../ir/quintTypes'
...
-  kind: DefinitionKind
-  reference: bigint
+export type Definition = (QuintDef | ({ kind: 'param' } & QuintLambdaParameter)) & {
+  /* Hidden definitions won't be copied over to a module when an
+   * import/intance/export statement is resolved. `hidden` can be removed with
+   * `export` statements for the hidden definitions. */
+  hidden?: boolean
+  /* `namespaces` are names to add to the definition's name, when it
+   * is copied from one module to another with a qualified name. Ordered from
+   * innermost to the outtermost. */
```
thpani, line 41:
```suggestion
  /* `namespaces` are names to add to the definition's name, when it
   * is copied from one module to another with a qualified name. Ordered from
   * innermost to the outermost namespace. */
```

quint/src/names/base.ts
```diff
@@ -20,21 +21,37 @@ import { QuintType } from '../ir/quintTypes'
...
+  namespaces?: string[]
+  /* importedFrom` is a reference to the import/instance/export statement that
+   * originated the definition, when the definition was copied from another
```
thpani, line 44:
```suggestion
   * originated the definition, if the definition was copied from another
```

quint/test/builders/ir.ts
```diff
@@ -27,12 +27,21 @@ export function buildModule(
...
+  return quintModule.declarations[0]
 }
 
 export function buildDef(def: string): QuintDef {
-  const quintModule = buildModuleWithDefs([def])
-  return quintModule.defs[0]
+  const quintModule = buildModuleWithDecls([def])
+  const decl = quintModule.declarations[0]
+  if (!isDef(decl)) {
+    throw new Error(`Error trying to build def from declaration that is not a def - ${JSONbig.stringify(decl)}`)
```
thpani, line 42:
```suggestion
    throw new Error(`Error trying to build def from declaration that is not a def: ${JSONbig.stringify(decl)}`)
```

thpani (approved):
LGTM, this wasn't too bad to review after all 😄

@konnov could you double-check Gabriela's highlight here? https://github.com/informalsystems/quint/pull/1105#discussion_r1295813492

konnov (commented):
LGTM, but we have to understand the expected behavior of the call graph

quint/src/static/callgraph.ts
bugarela: > Is it that all the definitions inside an instance depends on the instance name
I don't think we need this. 

> Given that, I don't think that push and pop are doing much, except maybe adding the dependencies on the instance parameters
I think that's what I was aiming at, so something like `import A(N = bar).*` has a dependency on `bar`.

quint/src/names/base.ts
bugarela: I meant to delete these field details from the type definition since I added individual comments for each field (a few lines below). Forgot to do that after I changed my mind from typing stuff here to typing the individual comments. I'll just delete it now.

Issue #823, opened 2023-04-17
Manage Apalache server


shonfeder:
To provide a user-friendly API, we would like to have a way for quint to manage the apalache server so users don't have to think about this.

For the immediate term, we can just inform CLI users about how to run the server when a connection cannot be made. But as soon as we expose model checking functionality via the VSCode plugin, we should give the LSP server a way to start and spot the Apalache server.

For the future, perhaps we'll want some sort of interactive console similar to `sbt`, which can be used to manage various parts of the specification lifecycle.

thpani:
Inspiration to gather options:

- bazel runs a [long-lived server process, with an idle timeout](https://bazel.build/run/client-server)
- bloop+metals
- jupyter notebooks
- proof assistants

PR #1106, opened 2023-08-11
Add ADR008: Obtaining and Launching Apalache

### LINKED ISSUE #701: Design and plan release and integration with Apalache

thpani:
This RFC discusses the problem space and proposes a solution for obtaining and launching Apalache, in support of the `quint verify` command.

I would like to receive feedback on the current proposal before expanding it further.

Towards #823, closes #701


Kukovec:
I think this aptly covers most reasonable options, and I personally agree with the "Proposed Solution" part, i.e. I prefer avoiding package managers when GH will do. Obviously, the version stuff is TBD, so I can't really comment on it. I'll just make a note here for future reference to discuss the scenario when Apalache is both present, and outdated, in the sense that it might not be a bad idea for the Quint plugin to notify the user that a newer version of Apalache is available (and either DL it, or have them update manually).

doc/rfc008-managing-apalache.md
```diff
@@ -0,0 +1,267 @@
...
+      - This shall print instructions for obtaining a JRE, if none is detected.
+2. Quint launches an on-demand instance of this local installation by spawning
+   `apalache-mc server` in a separate process.
+   - Such a process is spawned for each invocation of the `verify` command, and
+     torn down immediately after the RPC response.
+     - This seems to be a reasonable tradeoff given Apalache's memory
+       consumption and JVM startup time.
+     - We can later change this to a long-lived server, if we use a more
+       stateful approach of interacting with Shai (e.g., via the [transition
+       explorer API][]).
```
bugarela, line 248: In addition to this, we could have a special management inside the language server that keeps the apalache server running. Since not all language server users will need apalache, we could start apalache on the first request of a `verify`, and then keep it running until the language server is closed.

I don't think this is super necessary, surely not for now. But I wanted to point out that there is this language server scenario where I believe it makes sense to have the apalache server running continuously.

bugarela (approved):
Looks very good. I was feeling very pessimistic while reading the options, but at the end you found a really good solution :tada:

konnov (approved):
Looks good! The solution provides us with a good balance between simple behaviour by default and manual setup, if needed

doc/rfc008-managing-apalache.md
thpani: That's great input! I'm still a bit weary of keeping a long-running Apalache server running, because most people will only use it intermittently.
But launching it from LSP is a good pointer! Definitely the preferred option once we're able to run Quint from VSCode.

thpani:
I've added a [section with options](https://github.com/informalsystems/quint/blob/2b577b65ecc768c63c8888be264dadbd1b825327/doc/rfc008-managing-apalache.md#33-version-dependencies-between-quint-and-apalache) for managing the Apalache version from Quint.

Personally, I'm torn between 3.3.2 and 3.3.4, but maybe one of you has better ideas.
@konnov @bugarela @Kukovec

doc/rfc008-managing-apalache.md
```diff
@@ -0,0 +1,343 @@
...
+Pros:
+
+- Enables user to downgrade Quint, while maintaining compatibility with Apalache.
+
+Cons:
+
+- Users don't receive Apalache bugfix releases, unless we cut another Quint
+  release that pins the updated dependency.
+
+#### 3.3.4 Pin a minor release branch
```
bugarela, line 280: This sounds like a good enough approach to me! Also, I think that at some point in the future, when apalache, quint and their integration are all more stable, we can evolve this into "Pin a major release branch" and not have to worry about the "Cons" point more than a couple of times per year :smile:

bugarela (approved):

doc/rfc008-managing-apalache.md
```diff
@@ -0,0 +1,343 @@
+# RFC008: Obtaining and Launching Apalache from Quint
```
shonfeder, line 1: Could we name the RFCs sequentially starting with #1062?

Using the same numbering for RFCs and ADRs isn't helpful, ime, since they (should) serve pretty different purposes!

```suggestion
# RFC001: Obtaining and Launching Apalache from Quint
```

doc/rfc008-managing-apalache.md
```diff
@@ -0,0 +1,343 @@
...
+      - This shall print instructions for obtaining a JRE, if none is detected.
+2. Quint launches an on-demand instance of this local installation by spawning
+   `apalache-mc server` in a separate process[^1].
+   - Such a process is spawned for each invocation of the `verify` command, and
+     torn down immediately after the RPC response.
+     - This seems to be a reasonable tradeoff given Apalache's memory
+       consumption and JVM startup time.
+     - We can later change this to a long-lived server, if we use a more
+       stateful approach of interacting with Shai (e.g., via the [transition
+       explorer API][]).
```
shonfeder, line 319: An option not discussed here afaict is tying apalache's lifetime to the quint processes lifetime, e.g., by:

- Having an interactive terminal session for checking specs, akin to `sbt`
- Managing the server from the vs-code extension.

I think these can be postponed as possibly later feature, but the should be noted here for consideration.

shonfeder (commented):
Great writeup!

doc/rfc008-managing-apalache.md
shonfeder: Unless this is actually an ADR? Is seems like it may be.

And what is the difference? I guess maybe more thought is needed here :sweat_smile:

thpani: I think this has evolved into an ADR. I will revamp it a bit, to record the discussions here.

There is an advantage though in sharing a single counter between RFCs and ADRs: for RFCs that grow up to be an ADR, both documents will share the same index, and their correspondence is easily identifiable.

doc/rfc008-managing-apalache.md
```diff
@@ -0,0 +1,343 @@
...
+##### On-demand vs long-lived
+
+As a sub-option, we would have to decide if Shai is launched on-demand, for each
+verification invocation, or long-lived for as long as the VSCode extension is
+active. bazel and vscode-metals employ a hybrid approach, where a long-lived
+server is launched on-demand and torn down after an inactivity timeout. This
+decision should take into consideration the memory-overhead of a long-lived
+server vs. JVM startup time for an on-demand approach; at the time of writing,
+a freshly launched `apalache-mc server` takes ~125MB of memory, while JVM
+startup time is negligable (< 2sec).
```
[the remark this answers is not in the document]
thpani: Relative to usual Apalache runtime, I think it is?

doc/rfc008-managing-apalache.md
thpani: I really meant this to apply regardless of how verification is invoked. Could be the CLI, VSCode, or another frontend. I think discussing front-end options is orthogonal to this document, since all of them need to somehow spawn and connect to a Shai instance, and ideally this ADR should provide a transparent solution for all of them.

This also came up in @bugarela's review, I will revise this paragraph to be less ambiguous.

doc/rfc008-managing-apalache.md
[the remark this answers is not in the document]
bugarela: I guess if it was really negligible, then there would be no reason for us to consider managing the server in any scenario. Perhaps
```suggestion
server vs. JVM startup time for an on-demand approach; at the time of writing,
a freshly launched `apalache-mc server` takes ~125MB of memory, while JVM
startup time is negligible for most runs (< 2sec).
```

One example scenario where this is not negligible may be running the apalache simulator with a few runs to iterate over something (once we have a command for that)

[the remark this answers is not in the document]
thpani: > then there would be no reason for us to consider managing the server in any scenario

ergonomics, I think

> running the apalache simulator with a few runs to iterate over something

I think what you describe here is still a single invocation of the simulate command in Apalache? Otherwise, we'd probably want the more fine-grained interface described in [Apalache RFC-010](https://apalache.informal.systems/docs/adr/010rfc-transition-explorer.html)?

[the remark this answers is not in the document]
shonfeder: If you are running 10

doc/rfc008-managing-apalache.md
shonfeder: I don't think I agree with the premise that ADRs are an evolutionary stage of RFCs :sweat_smile: IMO, they call for quite different scope and content.

But this is part of a larger process discussion that doesn't effect this PR. I think this is an ADR :+1:

doc/rfc008-managing-apalache.md
konnov: I also like 3.3.4 the most. It would let us go ahead with the Quint and Apalache development while maintaining relatively stable versions of both.

doc/rfc008-managing-apalache.md
konnov: I know we have been discussing that multiple times. I also think that RFCs and ADRs are related. This may be as simple as a usability problem: When looking for RFCs or ADRs, one may see gaps in numbering. If we consider the (unique) document number as the most important attribute, then we could call documents like 002rfc and 003adr, meaning that "rfc" and "adr" are just tags of the envisioned document content.

doc/rfc008-managing-apalache.md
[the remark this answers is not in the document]
konnov: It looks to me that there are two main user stories here:

 1. A first time user tries to run `quint verify`. They want to try the command and definitely do not want to manage the server (they probably even do not know what it means).
 2. An experienced user runs `quint verify` periodically. They would prefer to have (one instance of) the server in the background.

At this point of development, I would prioritise the first-time user over the experienced user. However, we should give the experienced user an option to optimize their workflow by running the server in the background. As far as I understand, there is such an option.

doc/rfc008-managing-apalache.md
shonfeder: They can be related, sure. But that doesn't mean ADRs are "evolved" RFC. IMO, reading on the intent and traditions around either makes it immediately clear that they have been and should be distinct documents, with different content and purposes. See, for instance:

- https://news.ycombinator.com/item?id=26057068
- https://engineering.atspotify.com/2020/04/when-should-i-write-an-architecture-decision-record/
- https://medium.com/@klaus.haeuptle/shaping-architecture-decisions-from-request-for-comments-rfc-to-documenting-architecture-11dad9877342

> If we consider the (unique) document number to be the most important attribute

Why would we consider this? The document number is essentially arbitrary afaik. "the envisioned document content" seems much more important than the order in which documents were drafted imo.

doc/rfc008-managing-apalache.md
```diff
@@ -0,0 +1,343 @@
...
+
+We choose managed options for obtaining and launching Apalache from Quint, while
+still providing a non-managed option:
+
+1. Primarily, Quint uses a Quint-managed, on-demand instance of Shai:
+   1. If not present, Quint fetches a compatible Apalache release from Github,
+      using [Github's REST API endpoint][] and/or [`octokit/request.js`][] and
+      unpacks it into a local installation directory.
+      - This can be scripted etirely from NodeJS, foregoing the need to wrap a
+        third-party package manager binary.
```
shonfeder, line 307: If we really think it is so important to dynamically determine the latest apalche version, then I think we should not run this by default on every invocation. Rather, lets protect it behind an update flag and/or gate any automatic checks via a periodic (e.g., once per week) check.

We should also, ideally, have a way to fall back to a known good version url, i.e., the minimal minor version `m.n.0`, so that our CLI doesn't break due to api limits.

That all said, I think it is probably fine for now to just hard-code the compatible apalache version and push updates as needed via releases. If that gets to be a bottleneck, then we can optimize. This seems a bit premature to me, in light of the complexity costs we see in, e.g., #1124.

shonfeder: "this" is meant to refer to the design of dynamically determining and updating to the latest patch version of apalache. This introduces a lot of complexity and variable dependencies afaict, 'cause we are ending up trying to implement like a fragment of a package manager :)

I actually disagree that my suggestion increases complexity of the code base, or the UI. Afaiu, it would only be a slight increase in the *amount* of work for releases, but actually would simplify rather than complicate our dependency on Apalache. My perspective is tied to the view that complexity is about how difficult it is to understand or reason about a system. As a user, I'd rather know when things are going to update and control that rather than have it automatically and without my knowledge performing updates every time I run it. The former is *simpler*, in my view, even tho it requires me taking an extra action when I want to update,  because it is easier to understand what is going on and how the system works.

That said, I do think there are ways I think we can support this variable version approach without hitting a bottleneck on the github API. I just can't think of any that don't require us to invest in developing a bigger fragment of a package manager, and I just don't see how that lies on our critical path atm! :sweat_smile: 

But maybe this is best for a sync discussion next week! Nothing urgent to address I think!

thpani: > As a user, I'd rather know when things are going to update and control that rather than have it automatically and without my knowledge performing updates every time I run it.

I think your concern is taken care of under §3 below:
> To give the user more control over their system, or as a development setup, we provide a CLI flag and/or config file option [...] Quint skips the managed approach above and instead connects to the provided socket and uses this instance of Apalache dispatch verification commands.

You are free to maintain that version of Apalache in any way you like.
It would be helpful to understand if this covers your needs, and, if not, why?

> we are ending up trying to implement like a fragment of a package manager

> I'd rather know when things are going to update and control that rather than have it automatically and without my knowledge performing updates every time I run it

> As a user, I'd rather know when things are going to update and control that rather than have it automatically and without my knowledge performing updates every time I run it.

I think this is a biased framing of the situation, and I would prefer to stick to the facts in a technical discussion. You do know when things are going to update (when you run `verify`), it's not automatic but user-initiated, and it's printing messages so nothing goes on without your knowledge.

And no, I'm definitely not trying to implement a package manager.

thpani: That said, @ivan-gavran mentioned yesterday that the MBT team hit similar questions when developing Atomkraft. I will do some research next week where there discussion ended up, so we don't unnecessarily duplicate efforts.

shonfeder: Let’s talk about this sync next week! Sorry if my remarks came across as flippant.

To be clear, I don’t think that the fact this is tending towards the problems of package management is a fault in your efforts or thinking, imo that’s just the kind of problem space we are in. It’s a complex domain where there isn’t a great, widely used, cross platform solution, and figuring out a simple but flexible solution to our needs is a significant design task imo.

doc/rfc008-managing-apalache.md
```diff
@@ -0,0 +1,343 @@
...
+We want to maintain compatiblity between Quint and Apalache, therefore we need
+some mode of linking compatible versions of both tools. Both tools follow
+[semantic versioning][].
+
+The crux here is that both an upgrade and a downgrade of Apalache, relative to a
+specific Quint release, may maintain or break compatibility.
+
+Given that Quint calls Apalache, but not the other way around, it makes sense to
+maintain a list of compatible Apalache versions for a given Quint release.
+Broadly speaking, there's two options for maintaining such a list:
```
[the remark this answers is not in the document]
thpani: [606b2d7](https://github.com/informalsystems/quint/pull/1106/commits/606b2d7b84089d3b8f4a47d0486493af05094960)

thpani:
I've updated §§ 3.2.4 and 4 of this doc to match our main branch wrt to pinning a Quint version.

Please take another look!

doc/adr008-managing-apalache.md
```diff
@@ -0,0 +1,366 @@
...
+- Completely transparent to the user.
+- Allows managing the required Apalache version from within Quint.
+- No package manager dependency like above.
+
+Cons:
+
+- Assumes that the JRE is installed on the local system.
+
+*Note: We did a prototype implementation querying the GitHub REST API in [#1115](https://github.com/informalsystems/quint/issues/1115).
+However, we observed CI issues due to the Github API's rate limiting as described in [#1124](https://github.com/informalsystems/quint/issues/1124). In pratice, the same issues can affect users (e.g., behind a shared IP) and may segnificantly impact UX of the `verify` command. As a countermeasure, we reverted to a hardcoded Quint version in [`4ceb7d8`](https://github.com/informalsystems/quint/commit/4ceb7d8be824ddc0a2c2a14e105baff446f71e72).*
```
bugarela, line 229:
```suggestion
However, we observed CI issues due to the Github API's rate limiting as described in [#1124](https://github.com/informalsystems/quint/issues/1124). In pratice, the same issues can affect users (e.g., behind a shared IP) and may segnificantly impact UX of the `verify` command. As a countermeasure, we reverted to a hardcoded Apalache version in [`4ceb7d8`](https://github.com/informalsystems/quint/commit/4ceb7d8be824ddc0a2c2a14e105baff446f71e72).*
```

bugarela (approved):
Looks great! Thanks for taking such good care of this :sparkles:

doc/adr008-managing-apalache.md
```diff
@@ -0,0 +1,366 @@
...
+interface.
+
+There's three outstanding questions concerning this integration:
+
+1. How is Shai managed (launched, stopped, ...)?
+2. How is the Apalache distribution obtained?
+3. How are version dependencies between Quint and Apalache managed?
+
+Obviously, a decision on each of these questions impacts the other ones. We use
+this document to survey options, and to list advantages and disadvantages.
```
shonfeder, line 25:
```suggestion
Obviously, a decision on each of these questions impacts the other ones. We use
this document to survey options, and to list advantages and disadvantages, and to
document the proposed solution.
```

Based on the final section.

doc/adr008-managing-apalache.md
```diff
@@ -0,0 +1,366 @@
...
+    (for example, `vscode-metals` includes a `coursier` binary in its
+    distribution).
+- Spawning a package manager increases complexity by running a process whose
+  behavior is outside of our control.
+
+##### 3. Pulling the Apalache distribution from Github
+
+We either hardcode a fixed Apalache release (and its artifact URL) in Quint, or
+use [Github's REST API endpoint][] and/or [`octokit/request.js`][] to look up
+a matching release and download the JAR distribution directly.
```
shonfeder, line 216: Three additional options have occurred to me: 

1. We could maintain list of supported versions in a file somewhere (e.g., in the quint repo) and just request the file content, without having to hit the GH API. We could semi-automate updates to this page.
2. We could scrape the releases from https://github.com/informalsystems/quint/releases, again only requiring content of a served page, rather then going through the API.
3. We could require users to authenticate with github in order to use the apalache managed version. This would remove the rate-limiting issue, and we could also authenticate in this way in our CI.

Note that part of the problem in the previous implementation is that, iirc, every invocation of `verify` was hitting the GH API, meaning you could not invoke the subcommand more than 60 times an hour from the same IP address. If we only checked for updated versions periodically -- e.g., once per week, or when a user runs an `update` subcommand -- we would not hit the exact same problem (but other would persist).

shonfeder (approved):
Looks great!

Just a few notes and incidental suggestions from me

doc/adr008-managing-apalache.md
```diff
@@ -0,0 +1,366 @@
...
+##### On-demand vs long-lived
+
+As a sub-option, we would have to decide if Shai is launched on-demand, for each
+verification invocation, or long-lived for as long as the VSCode extension is
+active. bazel and vscode-metals employ a hybrid approach, where a long-lived
+server is launched on-demand and torn down after an inactivity timeout. This
+decision should take into consideration the memory-overhead of a long-lived
+server vs. JVM startup time for an on-demand approach; at the time of writing,
+a freshly launched `apalache-mc server` takes ~125MB of memory, while JVM
+startup time is negligable (< 2sec).
```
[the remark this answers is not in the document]
thpani: Then let's just record the facts:

```suggestion
a freshly launched `apalache-mc server` takes ~125MB of memory, and JVM
startup time is below 2 seconds.
```

doc/adr008-managing-apalache.md
thpani: Interesting options! Should we record these in a separate issue? I'm a bit concerned that this PR is becoming the record of decisions, rather than the ADR itself 😅

thpani: split off into #1248

shonfeder: Thank you for opening the issue!

PR #1112, opened 2023-08-22
Fix problem with loading module in the REPL

bugarela:
Hello :octocat: 

This fixes the issue encountered by Jure during the Quintshop. The problem was: we used to add incremental declarations/expressions to the last module in the list of modules, because the last module was always the `__repl__` module. When we [stopped requiring a `__repl__` module](https://github.com/informalsystems/quint/pull/1039), that became a problem, because not necessarily the last module is the actual main module that was loaded. This PR makes it so the new declarations are always added to the main module.

The reason it works when the file is imported, and not loaded via CLI, is because once no module is specified in the CLI (nor in an annotation inside the file), we introduce a `__repl__` module as the last module, and that is the main one even if you type import statements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 ### Removed
 ### Fixed
+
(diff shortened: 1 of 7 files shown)


konnov (approved):
Looks good!

PR #1115, opened 2023-08-23
Manage Apalache server

### LINKED ISSUE #823: Manage Apalache server

thpani:
Manage Apalache server from Quint, as sketched in [RFC 008](https://github.com/informalsystems/quint/pull/1106).

I ended up reworking most of the retry logic for obtaining a gRPC connection, which blows up this PR a bit.

Closes #823 


quint/src/quintVerifier.ts
```diff
@@ -311,11 +315,141 @@ async function retryWithTimeout<T>(f: () => Promise<T>): Promise<VerifyResult<T>
...
+ *    - a `right<Apalache>` equal to the path the Apalache dist was unpacked to,
+ *    - a `left<VerifyError>` indicating an error.
+ */
+async function connect(): Promise<VerifyResult<Apalache>> {
+  // Try to connect to Shai, and try to ping it
+  const connectionResult = await tryConnect()
+  // We managed to connect, simply return this connection
+  if (connectionResult.isRight()) {
+    return connectionResult
+  }
```
shonfeder, line 415: As is, this means waiting 5 seconds every startup (if the server isn't running). We can fix that by adding a way to try getting a connection without retrying and waiting.

shonfeder (commented):
Great work, because everything works! I had a noodle around with some idioms trying to streamline the Either/async interactions, and the tar extraction.

Most of those suggestions are in the comments here, but I've also open a PR into this draft in case you want to see them all together or merge them in: #1116

quint/src/quintVerifier.ts
```diff
@@ -311,11 +315,141 @@ async function retryWithTimeout<T>(f: () => Promise<T>): Promise<VerifyResult<T>
...
+  const protoDefResult: VerifyResult<proto.PackageDefinition> = process.env.APALACHE_DIST
+    ? findApalacheDistribution().chain(loadProtoDefViaDistribution)
+    : await loadProtoDefViaReflection()
+  // Load gRPC client
+  const maybeCmdExecutor = protoDefResult.map(loadGrpcClient)
+  const pingResult = await maybeCmdExecutor.asyncChain(cmdExecutor =>
+    // Try to ping the server, with a timeout
+    retryWithTimeout(() => cmdExecutor.ping())
+  )
+  return pingResult.chain(_ => maybeCmdExecutor.map(apalache))
```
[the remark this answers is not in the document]
thpani: I understand this can be chained, but I'm skeptical of chaining `Either`s to achieve logically distinct things (in this case, loading the gRPC client vs. pinging).

I don't even know where to put comments in this case.

quint/package.json
```diff
@@ -50,20 +50,24 @@
...
+    "gunzip-maybe": "^1.4.2",
     "immutable": "^4.3.0",
     "json-bigint": "^1.0.0",
     "line-column": "^1.0.2",
     "lodash": "^4.17.21",
     "lodash.isequal": "^4.5.0",
+    "node-fetch": "^2.6.13",
     "seedrandom": "^3.0.5",
     "tar": "^6.1.14",
+    "tar-fs": "^3.0.4",
```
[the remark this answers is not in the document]
thpani: Good catch. I hadn't considered that we already include a tar library.
(Also, I can't find any current use? Why is it in our package.json?)

> exposes an `async` friendly interface

yeah. why is this a plus? 😅 

to me, `tar-fs` seemed to have a nicer interface, but I like that `tar` supports gzip out of the box. Overall using `tar` looks cleaner, I will switch to it.

quint/src/quintVerifier.ts
```diff
@@ -311,11 +315,141 @@ async function retryWithTimeout<T>(f: () => Promise<T>): Promise<VerifyResult<T>
...
+  return octokitRequest('GET /repos/informalsystems/apalache/releases').then(
+    async resp => {
+      const releases = resp.data.map((element: { tag_name: string }) => element.tag_name)
+      const latestTaggedVersion = semver.maxSatisfying(releases, APALACHE_VERSION_TAG)
+      // Filter release response to get dist archive asset URL
+      const taggedReleases = resp.data.filter(
+        (element: { tag_name: string }) => element.tag_name == latestTaggedVersion
+      )
+      const tgzAssets = taggedReleases[0].assets.filter((asset: { name: string }) => asset.name == APALACHE_TGZ)
+      const downloadUrl = tgzAssets[0].browser_download_url
```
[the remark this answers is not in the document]
thpani: Yeah, but are there any guarantees that this URL is stable?

We already receive asset URLs as part of the releases query, so why not use them?

quint/src/quintVerifier.ts
thpani: Uh, you're right. I assumed that once we obtain an `AsyncCmdExecutor` and the server is already up, the ping would succeed immediately. I will change this.

quint/src/quintVerifier.ts
```diff
@@ -311,11 +315,141 @@ async function retryWithTimeout<T>(f: () => Promise<T>): Promise<VerifyResult<T>
...
+  // Try to connect to Shai, and try to ping it
+  const connectionResult = await tryConnect()
+  // We managed to connect, simply return this connection
+  if (connectionResult.isRight()) {
+    return connectionResult
+  }
+
+  // Connection or pinging failed, download Apalache
+  console.log("Couldn't connect to Apalache, downloading latest supported release")
+  const distDir = await fetchApalache()
```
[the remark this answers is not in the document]
thpani: > Does this mean we will hit the github API everytime we need to start shai?

Yes. Could you articulate what issue you see with that?
It takes an average of 360ms on my machine.

We already have a pathway for performance junkies, which is to launch Shai separately.
So what you suggest may be an incremental improvement, but I'm hesitant to include it as long as we don't know if it's necessary.

quint/src/quintVerifier.ts
```diff
@@ -311,11 +315,141 @@ async function retryWithTimeout<T>(f: () => Promise<T>): Promise<VerifyResult<T>
...
+  return fetch(url).then(
+    res =>
+      new Promise((resolve, _) => {
+        const untar = tar.extract(unpackPath)
+        res.body.pipe(gunzip()).pipe(untar)
+        untar.on('close', () => resolve(right(unpackPath)))
+        untar.on('error', error => resolve(err(`Error unpacking .tgz: ${error}`)))
+      }),
+    error => err(`Error fetching ${url}: ${error}`)
+  )
```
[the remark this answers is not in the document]
thpani: Good point. Switched to `tar` in 0ed4ad60

quint/package.json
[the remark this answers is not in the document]
thpani: Switched to `tar` in 0ed4ad60

quint/src/quintVerifier.ts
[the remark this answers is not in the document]
shonfeder: I think we can rely on it being stable, yes. At least as stable as the github API'd I'd assume. I could be wrong here tho.

> We already receive asset URLs as part of the releases query, so why not use them?

My only rationale is simpler code via reducing what we need to compute and how much of the shape of their API we depend on :)

This is not a sticking point for me if you feel that the current method is superior.

quint/src/quintVerifier.ts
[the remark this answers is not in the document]
shonfeder: Yeah, it's maybe just a question of style. IMO they are *not* logically distinct tho in the sense that we cannot ping unless we have the grpc client loaded. The chaining us justified, imo, by the logical dependency and the fact that that one way or another we are sequencing the error-prone computations.

quint/src/quintVerifier.ts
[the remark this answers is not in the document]
shonfeder: Beyond micro improvements, I have the aesthetic preference of not wanting my CLI tools to depend on hitting external APIs everytime I invoke them, esp when. it'll almost always be a wasted call.

But this also isn't a sticking point for me. We can always follow up here with incremental improvements in behavior and perf.

quint/src/quintVerifier.ts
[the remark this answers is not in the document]
thpani: The GH API is versioned, so I believe that's the more stable solution.

quint/src/quintVerifier.ts
thpani: Reworked the retry logic in [1a6a1d5](https://github.com/informalsystems/quint/pull/1115/commits/1a6a1d563ff9c1ca7d92cb5cf9c10393f2f57a2b)

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 ### Changed
+
+- The `verify` command now manages Apalache server, if no already running
+  instance is detected (#1115)
+
```
shonfeder, line 15: This is a feature add, not a breaking change, iiuc?

```suggestion
### Added

- The `verify` command now manages Apalache server, if no already running
  instance is detected (#1115)
  
### Changed
```

quint/package.json
```diff
@@ -50,18 +50,21 @@
...
     "@grpc/proto-loader": "^0.7.7",
+    "@octokit/request": "^8.1.1",
     "@sweet-monads/either": "^3.0.1",
     "@sweet-monads/maybe": "^3.1.0",
     "@types/line-column": "^1.0.0",
     "@types/seedrandom": "^3.0.4",
     "antlr4ts": "^0.5.0-alpha.4",
     "chalk": "^4.1.2",
     "eol": "^0.9.1",
+    "gunzip-maybe": "^1.4.2",
```
shonfeder, line 61: Still needed?

quint/package.json
```diff
@@ -84,10 +87,14 @@
...
   "devDependencies": {
     "@types/chai": "^4.2.18",
+    "@types/gunzip-maybe": "^1.4.0",
     "@types/json-bigint": "^1.0.1",
     "@types/lodash.isequal": "^4.5.6",
     "@types/mocha": "^8.2.3",
     "@types/node": "^18.16.0",
+    "@types/node-fetch": "^2.6.4",
+    "@types/tar": "^6.1.5",
+    "@types/tar-fs": "^2.0.1",
```
shonfeder, line 97: Again, assuming these are not needed:

```suggestion
```
Also the gunzip above.

quint/src/quintVerifier.ts
```diff
@@ -132,13 +132,11 @@ type CmdExecutor = {
   // Constructs a new client service
   new (url: string, creds: any): CmdExecutor
   run: (req: RunRequest, cb: AsyncCallBack<any>) => void
-  ping: (o: {}, cb: AsyncCallBack<void>) => void
 }
 
 // The refined interface to the CmdExecutor we produce from the generated interface
 type AsyncCmdExecutor = {
   run: (req: RunRequest) => Promise<RunResponse>
-  ping: () => Promise<void>
```
shonfeder, line 141: Nice that this can be removed!

quint/src/quintVerifier.ts
```diff
@@ -242,22 +191,33 @@ async function loadProtoDefViaReflection(): Promise<VerifyResult<ProtoPackageDef
...
+    await (async () => {
+      for (;;) {
+        const grpcChannelState = reflectionClient.getChannel().getConnectivityState(true)
+        if (grpcChannelState == grpc.connectivityState.READY) {
+          return
+        } else {
+          await setTimeout(1000)
+        }
+      }
+    })()
```
shonfeder, line 205: Why does this need to be a thunk if we are juts calling right away?

quint/src/quintVerifier.ts
```diff
@@ -242,22 +191,33 @@ async function loadProtoDefViaReflection(): Promise<VerifyResult<ProtoPackageDef
...
-          call.end()
-          resolve(r)
-        })
-        call.on('error', (e: grpc.StatusObject) => reject(e))
+  // Wait for gRPC channel to come up, with 1sec pauses
+  if (retry) {
+    await (async () => {
+      for (;;) {
+        const grpcChannelState = reflectionClient.getChannel().getConnectivityState(true)
+        if (grpcChannelState == grpc.connectivityState.READY) {
```
shonfeder, line 199: Great that this reliably ensures the service is ready! I did not have luck with this in the past, but maybe the timing taken by having to fetch the schema allows time to ensure the service if fully functional?

CHANGELOG.md
thpani: Do we use these categories in a way that diverges from https://keepachangelog.com/?
They simple state

> `Changed` for changes in existing functionality.

quint/package.json
thpani: Oof, thanks for the careful review! I'm really bad at tracking dependencies 😅
d2045ae9

quint/package.json
thpani: d2045ae9

quint/src/quintVerifier.ts
thpani: This still usually loops ~3 times on my machine while the server is coming up.
I previously was doing this in a busy waiting loop, and that did not work. I suspect that there is a race with some async parts of the gRPC lib, which need runtime cycles to bring up the connection (even if the server is fully up).
I will leave a comment here, to note that we actually need the `setTimeout` to yield to to runtime at this point.

thpani: See e78648f1

quint/src/quintVerifier.ts
thpani: Ah, good catch! I assume "a thunk" = "async"? 😄
Removed the async function wrapper in https://github.com/informalsystems/quint/commit/e78648f1b798148d335c3c22b5925260138d6c3b

CHANGELOG.md
thpani: I will merge this PR, since Igor wanted to send install instructions for the summer school.
LMK if I should follow up with an updated changelog.

quint/src/quintVerifier.ts
shonfeder: But doesn't this mean we actually still have a potential race condition here? I'd rather keep using `ping` then trust our timeout will always save us.

quint/src/quintVerifier.ts
shonfeder: A "thunk" is a computation suspended with a nullary function: https://en.wikipedia.org/wiki/Thunk#Functional_programming

So here I refer to the fact that we have `(() => ...)()` rather than the `async`.

CHANGELOG.md
shonfeder: Ah. I take "changes for existing functionality" to entail breaking changes. Otherwise, none of the categories entail breaking changes, and I also don't know how you can change existing functionality without that possibly breaking previous expectation about the API. That said, I can see how disagreement of the interpretation here is possible.

Apparently there has been a long-onging debate about this, lol: https://github.com/olivierlacan/keep-a-changelog/issues/41

Perhaps we should add a "Breaking changes" section?

I don't think my view is not widely enough shared to require any change to your entry.

quint/src/quintVerifier.ts
thpani: I cannot see a race here.

> I'd rather keep using ping then trust our timeout will always save us.

If the gRPC channel doesn't come up, ping will also fail. What do we gain by adding ping on top?

CHANGELOG.md
thpani: I see – yeah, it looks like the keepachangelog sections are a bit orthogonal to indicating breaking changes. Incidentally, also `Removed` may indicate a breaking change, so I guess in this taxonomy breaking changes are scattered across sections.

If we want to list breaking changes, we can either prefix individual items with a qualifier or break them into a separate section, as you suggested.

PR #1119, opened 2023-08-24
New flattening

### LINKED ISSUE #1102: Wire in the new flattening process
Use the new flattening process to flatten modules before compiling
```[tasklist]
### Tasks
- [ ] Raise proper errors in export scenarions that won't be supported for now
```

bugarela:
Hello :octocat:

Closes #1102.

I'm opening this in a bit of a rush so y'all have potentially more time to review (with your mornings), and that might contribute to releasing this tomorrow.

I was able to put all the pieces together with only a few adaptations.

  - will do tomorrow


bugarela:
I'll check tomorrow what is wrong with the integration with apalache. It did work on the PoC, so it shouldn't be something to worry about

bugarela:
Ok, found the issue. The new flattening process only copies definitions that are actually used. But currently it doesn't know that the `init`, `step` and `inv` definitions are being used.

It is not a problem with `simulate` since we do add `q::init`, `q::step` and `q::inv` definitions to the main module there.

For verify, we could simplify it and just add `run spec = init.then(step.and(inv))`. This would actually be helpful because it will trigger a proper effect check between init and step at least - so if they are updating a different set of state variables, an error will be shown.

konnov:
> For verify, we could simplify it and just add `run q::spec = init.then(all {step, inv})`. This would actually be helpful because it will trigger a proper effect check between init and step at least - so if they are updating a different set of state variables, an error will be shown.

That's a nice idea. I have never thought of using runs for doing that :)

Do you mean that we could do that automatically?

konnov (approved):
Looks great! I only worry about `quint verify`, which I wanted to use next week. Would it be possible to add an auxiliary definition like the one you proposed?

bugarela:
I'm trying to add auxiliary definitions like we do in the simulator, but it's not as straightfoward as I thought. Let's see if I can manage to make it work in the next few hours.

If that doesn't work and you think its a better idea to ship this partially, we could wire the new flattener only for repl/simulation/tests, and use the old one for apalache until I fix that. WDYT @konnov ?

konnov:
> If that doesn't work and you think its a better idea to ship this partially, we could wire the new flattener only for repl/simulation/tests, and use the old one for apalache until I fix that. WDYT @konnov ?

Yep, that's a great idea! Let's do it this way

bugarela:
Actually, I just manage to make it work for all tests and examples in the dashboard :smile: It also makes the `Lightclient.qnt` spec receive a :white_check_mark: on unit tests. I'll add some comments and push the commit. Is not the most elegant solution ever, but its under the same standard as the simulation approach. We can improve both of them with a single solution, most likely.

bugarela:
@konnov do you want to take another look?

konnov (approved):
Looks great! The solution with `q::init` and `q::step` is similar to `quint run`, so it's good for now. I am a bit tired preparing slides, so it may be that I miss some bugs :)

shonfeder:
The ci failure here is #1124. I'm investigating it. It shouldn't block merging this PR, but may interfere with releasing.

bugarela:
Yeah, I've been hitting that in this and in my previous PR since today morning

bugarela:
Ah, I forgot to update the CHANGELOG :woman_facepalming:

PR #1140, opened 2023-09-04
Adding a specification of hashes to the examples

konnov:
I have been thinking about a nice way of specifying hashes in Quint. One reason for not merging #975 yet is that I barely understand hashes in that spec myself. The best way would be introduce uninterpreted terms for hashing, e.g., `h(text)`, `h(h(text))`. However, we do not allow for recursive types. Hence, I had to apply a trick: Wraps a list of bytes with a marker -1, which gets decreased whenever additional hashes are applied.


examples/cryptography/hashes.qnt
```diff
@@ -0,0 +1,100 @@
...
+    input._bytes.foldl((0, false),
+      (p, b) =>
+        if (b == -1) {
+          if (p._2) {
+            // the marker for the end of a hash sequence
+            (p._1 + HASH_LEN, false)
+           } else {
+            // the marker for the start of a hash sequence
+            (p._1, true)
+           }
```
Kukovec, line 84:
```suggestion
          val newLen = p._1 + if (p._2) HASH_LEN else 0 // We add the total hash length at the terminator
          (newLen, not(p._2)) // toggle the "in hash" flag
```

examples/cryptography/hashes.qnt
```diff
@@ -0,0 +1,100 @@
...
+        } else {
+          if (p._2) {
+            // in hash, keep the length
+            (p._1, true)
+          } else {
+            // in plain text
+            (p._1 + 1, false)
+          }
```
Kukovec, line 92:
```suggestion
          val newLen = p._1 + if (p._2) 0 else 1 // if in hash, the total hash length gets added in the terminator
          (newLen, p._2)
```

Kukovec (approved):
LGTM, sugestions are purely for compactness

amiller:
I'm curious if the fact that it's not feasible to find two colliding hashes is expressed / exported as a property somehow? I see how incrementing the hashes by 1 as a concrete representation exhibits this, but not how it could be exposed in general as a property

konnov:
> I'm curious if the fact that it's not feasible to find two colliding hashes is expressed / exported as a property somehow? I see how incrementing the hashes by 1 as a concrete representation exhibits this, but not how it could be exposed in general as a property

This is an excellent question. I have been thinking about the same but have not come up with a good solution yet. If we were using an interactive theorem prover or a model checker that admits quantification over unbounded domains and supports recursive data types, that would be the way to go. For instance, we could write the following assumption:

```bluespec
// in Quint
assume(tuples(D, D).forall((x, y) => x == y iff hash(x) == hash(y)))
// in TLA+
ASSUME(\A x, y \in D: x = y <=> hash(x) == hash(y))
```

I can see how we could enforce this property in [Apalache](https://github.com/informalsystems/apalache/) even though we use a quantifier-free fragment of SMT. It is less obvious to me, how we would do that in the random simulator. However, we could interpret `hash` symbolically in the simulator and avoid any computation at all.

If you have an example of a system where it was done right already, I would be very curious to see. Now that I am thinking about it, maybe we should have a look at ACL2 or something similar.

amiller:
My experience is mainly with an early version of this production of Merkle Trees in F*. An early version defined a lemma `injective_hash`. But, when combined with an assumption that hashes have some fixed length, it seems vulnerable to disproving it by pigeonhole principle. (Not sure how this would matter for quint) 
The current version avoids defining the assumption of collision resistance, instead just leaves it to the Merkle Tree application to show that it can produce a collision.
https://www.fstar-lang.org/tutorial/book/part2/part2_merkle.html

In EasyCrypt they actually define collisions as a game, they have some probability model to work with tho
https://github.com/EasyCrypt/easycrypt/blob/main/theories/crypto/assumptions/CRHash.ec

konnov:
> My experience is mainly with an early version of this production of Merkle Trees in F*. An early version defined a lemma `injective_hash`. But, when combined with an assumption that hashes have some fixed length, it seems vulnerable to disproving it by pigeonhole principle. (Not sure how this would matter for quint)
> The current version avoids defining the assumption of collision resistance, instead just leaves it to the Merkle Tree application to show that it can produce a collision.

Right. The way they do it in F* is similar to what I would imagine in a proof system. I guess, there should be a way to execute code in F*. Would one have to provide some form of refinement of `hash` to make it executable? (I don't know much about F*, so this may be a very naive question :) )

PR #1143, opened 2023-09-05
Add partial completion and rename functionality in LSP/VSCode

thpani:
Output of my 🥧 week:

Add partial support for the completion and rename LSP endpoints in the VSCode plugin.

The logic for completion is a bit funky, because in most cases, when the endpoint is called, we don't have parsable source code. Hopefully we can alleviate this in the future by having an incremental parser.

The other thing I noted was that analysis as currently delayed by some timeout after parsing. While this works for hovering, it gives me mismatching IDs in the parser result and the analysis output when completion is called.
Thus, I'm currently calling analysis synchronously in `completeIdentifier` – I'm sure you have suggestions how to improve this :)


vscode/quint-vscode/server/src/complete.ts
```diff
@@ -0,0 +1,250 @@
...
+}
+
+/**
+ * Find a declaration of `name` by name inside `decl`.
+ */
+function findDeclByNameInDecl(name: string, decl: QuintDeclaration): bigint[] {
+  const visitor = new DeclOfNameFinder(name)
+  walkDeclaration(visitor, decl)
+  return visitor.namesFound
+}
```
bugarela, line 87:
```suggestion
}

```

quint/src/index.ts
```diff
@@ -16,6 +16,8 @@ export { produceDocsById, DocumentationEntry } from './docs'
 export { builtinDocs } from './builtin'
 export * as Builder from '../test/builders/ir'
 export { TypeScheme } from './types/base'
+export * from './ir/IRVisitor'
+export * from './static/callgraph'
```
bugarela, line 20: Is this being used?
```suggestion
```

vscode/quint-vscode/server/src/complete.ts
```diff
@@ -0,0 +1,250 @@
...
+  position: Position,
+  builtinDocs: Map<string, DocumentationEntry>
+): CompletionItem[] {
+  const { modules, table } = parsedData
+
+  // Until we have incremental parsing, try to lookup `triggeringIdentifier` by name
+  const declIds = findDeclByNameAtPos(identifier, position, sourceFile, parsedData)
+
+  // FIXME: use analysisOutputByDocument
+  const [_, aop] = analyzeModules(table, modules)
```
bugarela, line 43: Have you tried using `analysisOutputByDocument`? I'm not sure exactly what the consequences of the delay might be, but we might be fine with it.

We might also be fine with running the analysis when users type `.`. The reason for the delay is that we can't run it every time a user types anything, since that becomes laggy pretty quickly.

bugarela: We could try to trigger the async job here and avoid running it twice, but I'm not sure it's worth the trouble. The real improvement will be with incremental parsing.

thpani: > Have you tried using analysisOutputByDocument? I'm not sure exactly what the consequences of the delay might be, but we might be fine with it.

Yeah, I tried that, but in 99% of cases the analysis output and the parsed data are out of sync by then (i.e., using disjoint sets of IDs).

I guess the delay is fine for hover, because an analysis will be triggered until you move your mouse, but for triggering completion, the LSP request wins the race against the timeout.

thpani: Maybe we should make `analysisOutputByDocument` a Promise? Then I could wait for that.

vscode/quint-vscode/server/src/server.ts
```diff
@@ -260,6 +304,77 @@ export class QuintLanguageServer {
...
+        const decl = findDefinitionWithId(modules, declToRename)
+        if (decl) {
+          // The tricky case: the name is declared as a definition.
+          // In this case, the location pointed by the source map includes its
+          // entire body and quantifiers, while we're only looking to replace
+          // its name. We thus compute an offset, starting from the beginning
+          // of the range indicated through the source map, to take into account
+          // any preceeding qualifiers. The end of the replacement range is
+          // given by the length of the definition's name.
+          const sourceMapOffset = decl.kind.length + (decl.kind === 'def' && decl.qualifier.startsWith('pure') ? 5 : 0)
```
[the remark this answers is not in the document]
thpani: Good point! I have to think this through a bit more. We somehow need to isolate the identifier from its surroundings; in your example simply matching `/a/` would also match the `a` in `val`.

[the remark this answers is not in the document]
thpani: Changed to a regexp in 8cdc9ccd

vscode/quint-vscode/server/src/complete.ts
```diff
@@ -0,0 +1,250 @@
...
+function findDeclByNameAtPos(name: string, pos: Position, sourceFile: string, parsedData: ParserPhase3): bigint[] {
+  // Compute effects that contain the triggering position, from the smallest to the largest.
+  const { modules, sourceMap } = parsedData
+  const results: [bigint, null][] = [...sourceMap.keys()].map(id => [id, null])
+  const effectsContainingPosition = findMatchingResults(sourceMap, results, pos, sourceFile)
+
+  // let result = [] as bigint[]
+  for (let { id } of effectsContainingPosition) {
+    const names = findDeclByNameInsideScope(name, id, modules)
+    // TODO(#693): Collect all names as a workaround for #693, as they may receive a too-general type.
```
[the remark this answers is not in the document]
thpani: Ah yes, I forgot to remove part of this. Cleaned up in ea5de295

vscode/quint-vscode/server/src/complete.ts
```diff
@@ -0,0 +1,250 @@
...
+/**
+ * Search for a declaration of `name`, in increasingly wide effects around `pos`.
+ *
+ * @returns Array of IDs that declare `name`.
+ */
+function findDeclByNameAtPos(name: string, pos: Position, sourceFile: string, parsedData: ParserPhase3): bigint[] {
+  // Compute effects that contain the triggering position, from the smallest to the largest.
+  const { modules, sourceMap } = parsedData
+  const results: [bigint, null][] = [...sourceMap.keys()].map(id => [id, null])
+  const effectsContainingPosition = findMatchingResults(sourceMap, results, pos, sourceFile)
```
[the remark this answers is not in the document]
thpani: Ah, I think I was confused early on, because the comments in [`resultsOnPosition()`](https://github.com/informalsystems/quint/blob/5a36609ceb45178f6dd7069acdf45f9b1337c2c4/vscode/quint-vscode/server/src/reporting.ts#L145) refer to effects.

Maybe we should clean it up there as well?

vscode/quint-vscode/server/src/complete.ts
thpani: I tried to combine this with a promise, but I couldn't find a way to use the same analysis action both with a timeout and as a promise.

I'm off next week, maybe you have an idea here, otherwise I'll take another stab at it when I'm back!

bugarela: I gave this a shot and I think I got something pretty OK, please take a look and see if you  agree: https://github.com/informalsystems/quint/pull/1143/commits/ea330e09bfef2ef64de825d489acac3d7f6afd19

thpani: This looks great! Thanks for taking a shot at it! 🙏

PR #1146, opened 2023-09-05
Prevent variables from reseting when `__repl__` module is present

bugarela:
Hello :octocat: 

This PR fixes a problem reported by @konnov where the REPL was not behaving correctly when users evaluated import statements in the REPL (as opposed to loading a module through `quint -r file.qnt::module`. I could reproduce a bug where values of state variables were not persisting between evaluations - which was happening due to a change in our incremental compilation that makes it so declarations were sometimes re-evaluated. Re-evaluation should be fine, but our compiler was overwriting registers for variables with new ones - which we now prevent from happening.


quint/io-cli-tests.md
```diff
@@ -177,12 +177,14 @@ error: typechecking failed
 
 <!-- !test in repl loads a file -->
 ```
-echo "import counters.*" | quint -r ../examples/language-features/counters.qnt 2>&1 | tail -n +3
+echo -e "import counters.* \n init \n n" | quint -r ../examples/language-features/counters.qnt 2>&1 | tail -n +3
```
konnov, line 180: Nice that we have a test for this behavior now!

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -333,6 +333,12 @@ export class CompilerVisitor implements IRVisitor {
   }
 
   exitVar(vardef: ir.QuintVar) {
+    // In the process of incremental compilation, we might revisit the same var
+    // definition. Don't overwrite the register if that happens.
+    if (this.context.has(kindName('var', vardef.id))) {
+      return
+    }
```
konnov, line 340: Oh, that's a good catch!

konnov (approved):
Cool! Thanks a lot for fixing this!

PR #1158, opened 2023-09-12
Introduce new frames on actions

### LINKED ISSUE #854: The simulator should introduce a new frame on a new action
Consider the following REPL session:

```
>>> .verbosity 3
.verbosity=3
>>> var x: int
>>> action init = x' = 0
>>> action step = x' = x + 1
>>> init.then(step).then(step)
true

 [Frame 0]
 q::input() => true
 ├─ init() => true
 ├─ step() => true
 └─ step() => true

>>>
```

Similar to `quint run`, every invocation of an action should produce its own frame.

### LINKED ISSUE #1133: trace output by `test` disregards the last action
I have a suspicion that `quint test` does not execute the last action. Here is an example:

```
module lastActionInRun {
    var s: str

    action A = 
        s' = "A"
    
    action B = 
        s' = "B"

    
    run t1 = A.then(A).then(B)

    run t2 = A.then(A).then(B).then(s'=s)

    }
```

Upon running `quint test --match=t1 --output=out{#}{}.itf.json lastActionInRun.qnt`, I get two states, `A-A`.

Upon running `quint test --match=t2 --output=out{#}{}.itf.json lastActionInRun.qnt`, I get three states, `A-A-B`

konnov:
Closes #854. This PR fixes the long-standing issue of frames not being introduced in runs when they are executed in REPL. The goal was to fix #1133. However, this would require further refactoring to collect traces in the trace listener. Hence, I have decided to stop here and continue with #1133 once this PR is merged.

There are two major improvements in this PR:

 - [x] Introduction of `onNextState`, which allows us to switch to the next top-level frame:
 
```ts
  /**
   * This callback is called whenever an execution proceeds to the next state,
   * e.g., once A has been evaluated in A.then(B).
   * @param oldState the old state that is about to be discarded
   * @param newState the new state, from which the execution continues
   */
  onNextState(oldState: EvalResult, newState: EvalResult): void
```

 - [x] Introduction of the top-level operator `_`, which normally gets replaced with its child. However, in some cases like evaluating `2 + 3`, we need `_` to produce the verbose output like `_ => 5`.


quint/io-cli-tests.md
```diff
@@ -686,6 +689,35 @@ exit $exit_code
 ]
 ```
 
+### Test does not mean assignments (#1133)
```
shonfeder, line 692: I don't know how to parse the title of this test.

quint/io-cli-tests.md
```diff
@@ -686,6 +689,35 @@ exit $exit_code
...
+
+See: https://github.com/informalsystems/quint/issues/1133
+
+FIXME
```
shonfeder, line 696: Could you add a note indicating what is meant to be fixed?

quint/io-cli-tests.md
```diff
@@ -686,6 +689,35 @@ exit $exit_code
...
+exit_code=$?
+echo "BEGIN"
+# This test should have 3 states (FIXME: it does not!)
+cat out_0_t1.itf.json | jq '.states' | grep "s" | wc -l | grep 3
+rm out_0_t1.itf.json
+# This test should have 4 states (FIXME: it does not!)
+cat out_1_t2.itf.json | jq '.states' | grep "s" | wc -l | grep 4
+rm out_1_t2.itf.json
+echo "END"
+exit $exit_code
```
shonfeder, line 711: Since this is not actually testing the output but just the correct construction of the trace, should this perhaps just be in the typescript tests?

In otherwords, given the parsing logic required and what it being tested, I don't understand how/why its a good fit for a black box test.

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -35,9 +35,12 @@ import * as ir from '../../ir/quintIr'
 import { RuntimeValue, rv } from './runtimeValue'
 import { ErrorCode } from '../../quintError'
 
-import { lastTraceName } from '../compile'
+import { inputDefName, lastTraceName } from '../compile'
 import { unreachable } from '../../util'
 
+// Internal names in the compiler, which have special treatment.
+// For some reason, if we replace 'q::input' with inputDefName, everything breaks.
```
shonfeder, line 42: I'm curious, what breaks specifically?

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -252,10 +255,16 @@ export class CompilerVisitor implements IRVisitor {
...
       boundValue = {
         eval: () => {
-          this.execListener.onUserOperatorCall(app)
-          const r = unwrappedValue.eval()
+          if (app.opcode === inputDefName) {
+            this.execListener.onUserOperatorCall(evalApp)
+            // do not return from '_', as it may span over multiple frames
+          } else {
+            this.execListener.onUserOperatorCall(app)
+          }
```
shonfeder, line 266: I'm not sure how to read this comment. It seems like if `if` branch "does not return" then neither would the `else` branch?

quint/src/runtime/trace.ts
```diff
@@ -218,16 +226,26 @@ class TraceRecorderImpl implements TraceRecorder {
     // https://github.com/informalsystems/quint/issues/747
     if (verbosity.hasUserOpTracking(this.verbosityLevel)) {
       const newFrame = { app: app, args: [], result: none(), subframes: [] }
-      if (this.frameStack.length > 0) {
+      if (this.frameStack.length == 0) {
+        // this should not happen, as there is always bottomFrame, but let's handle it
+        this.frameStack = [newFrame]
```
shonfeder, line 231: If its an erroneous state as far as we know, should it not raise an error?

quint/io-cli-tests.md
shonfeder: Seeing how many FIXMEs are in this test: is the test testing anything useful currently? Or is just a sekeleton of a test that should be failing as it, but will be expanded to test something in the future? If the latter, I suggest leaving this test out of the PR.

shonfeder (approved):
With the caveat that I don't have deep understanding of this code, LGTM, modulo my few questions and minor suggestions.

quint/io-cli-tests.md
konnov: It's a test I have added when I started working on #1133. Since it is the next issue on my list, I would prefer keeping it in the imperfect shape rather than stashing it somewhere on my disk.

quint/io-cli-tests.md
konnov: It is testing the actual JSON file, which @ivan-gavran was inspecting by hand

quint/src/runtime/impl/compilerImpl.ts
konnov: Plenty of tests. I tried to understand, how this was possible, but could not. It really looked like magic to me

quint/src/runtime/impl/compilerImpl.ts
konnov: Good point. I will fix that.

quint/src/runtime/trace.ts
konnov: in principle, yes, but the trace output is a secondary feature, so I would prefer not to fail on that.

PR #1161, opened 2023-09-13
Ensure tests are present in the main module for flattening

### LINKED ISSUE #1150: Imported tests are not run by `quint test`
With the new flattener, unused definitions are not flatenned, and therefore tests might not be found if they are not declared in the main module. Examples: tendermint and `denomTrace.qnt`

bugarela:
Hello :octocat: 

This fixes #1150. In order to ensure that flattening will not get rid of unused test definitions, we now add an auxiliary definition to reference any unused test definitions, similar to what we are doing with `verify` and `simulate`.

In order to properly find the definitions with all necessary metadata, I added a field to name resolution return value, including any unused definition - since this values don't make it to the lookup table and, otherwise, would be lost after the name resolution stage.

### DIFF
--- a/quint/src/cliCommands.ts
+++ b/quint/src/cliCommands.ts
@@ -26,9 +26,9 @@ import {
 
 import { Either, left, right } from '@sweet-monads/either'
 import { EffectScheme } from './effects/base'
-import { LookupTable } from './names/base'
+import { LookupTable, UnusedDefinitions } from './names/base'
 import { ReplOptions, quintRepl } from './repl'
-import { OpQualifier, QuintEx, QuintModule } from './ir/quintIr'
+import { OpQualifier, QuintEx, QuintModule, QuintOpDef, qualifier } from './ir/quintIr'
 import { TypeScheme } from './types/base'
 import lineColumn from 'line-column'
 import { formatError } from './errorReporter'
@@ -55,6 +55,7 @@ interface OutputStage {
   // the modules and the lookup table produced by 'parse'
   modules?: QuintModule[]
   table?: LookupTable
+  unusedDefinitions?: UnusedDefinitions
   // the tables produced by 'typecheck'
   types?: Map<bigint, TypeScheme>
   effects?: Map<bigint, EffectScheme>
@@ -80,6 +81,7 @@ const pickOutputStage = ({
   warnings,
   modules,
   table,
+  unusedDefinitions,
   types,
   effects,
   errors,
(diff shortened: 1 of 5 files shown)


quint/src/cliCommands.ts
```diff
@@ -331,6 +335,37 @@ export async function runTests(prev: TypecheckedStage): Promise<CLIProcedure<Tes
...
+    d => d.kind === 'def' && options.testMatch(d.name)
+  )
+
+  // Define name expressions referencing each test that is not referenced elsewhere, adding the references to the lookup
+  // table
+  const args: QuintEx[] = unusedTests.map(defToAdd => {
+    const id = testing.idGen.nextId()
+    testing.table.set(id, defToAdd)
+    const namespace = defToAdd.importedFrom ? qualifier(defToAdd.importedFrom) : undefined
+    const name = namespace ? [namespace, defToAdd.name].join('::') : defToAdd.name
```
shonfeder, line 351: This seems like a slightly convoluted conditional test, that has to go through whether the temp variable `namespace` is undefined? iiuc, the point of the two lines, iiuc, is that the name should be qualified if `detToAdd.importedFrom` is truthy. This logic could be expressed a bit more directly, imo, via something like

```suggestion
    const name = defToAdd.importedFrom ? [qualifier(defToAdd.importedFrom), defToAdd.name].join('::') : defToAdd.name
```

shonfeder (approved):
LGTM

quint/src/cliCommands.ts
bugarela: > iiuc, is that the name should be qualified if detToAdd.importedFrom is truthy.

This is a correct if but not iff. `qualifier()` might return undefined as well, since not every import definition introduces a qualifier. So in your suggestion, we'd have to add an additional check for undefined, which I believe will make it more convoluted.

quint/src/names/base.ts
```diff
@@ -68,6 +68,19 @@ export type DefinitionsByModule = Map<string, DefinitionsByName>
  */
 export type LookupTable = Map<bigint, LookupDefinition>
 
+/**
+ * A lazy mapping from module names to the definitions that are available but not used in that module.
```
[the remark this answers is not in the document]
bugarela: Because it is a function, and therefore will only be evaluated if called. The value here is a set difference between all names in a module and the referenced names, so I imagine is not a super cheap operation - and we only need to use this mapping for one specific module (the main module, which we don't know what it is at this point) and in one specific cli command (test).

The strict version would be to define a `Map` instead of a function, at least in my understanding.

PR #1163, opened 2023-09-13
Rename shadowed names to unique names

### LINKED ISSUE #802: Flattening throws error when there is shadowing between different modules
Consider the modules:
```
module A {
  val isPositive(a: int): bool = a > 0
}

module B {
  var a: int
  import A.*
}

```

There is no error in name resolution since the lambda parameter `a` is scoped, and therefore not copied to module `B`. There's no way module `B` can access that name. However, since flattening just copies the entire definitions to the importing module, a conflict will arise in name resolution after flattening.

We could probably ignore this conflict at the flattening stage, but then we would have a problem in the integration with Apalache, since it also doesn't allow for shadowing. So we should ensure there is no shadowing in the flattened module to satisfy Apalache's constraint.

bugarela:
Hello :octocat: 

This fixes #802 by running a pass that renames shadowed names before flattening. We currently don't allow arbitrary shadowing, but it is possible to have shadowing between different modules' definitions, as in the example from the issue.

PS: If we decide to allow arbitrary shadowing, this solution should also work for that. We would only need to remove the current shadowing checks from the name resolution step.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 ### Removed
 ### Fixed
+
+- Fixed a problem where an error was thrown when a name from an importing module
+  shadowed a nested name from the imported module (#802)
+
 ### Security
 
 ## v0.14.2 -- 2023-09-06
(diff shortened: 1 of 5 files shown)


shonfeder (approved):
Looks great.

konnov (approved):
Nice!

PR #1165, opened 2023-09-14
Serialize big integers in ITF traces as an object

konnov:
This PR update the ITF trace format, according to the change made in Apalache, see https://github.com/informalsystems/apalache/pull/2733. As a result, the ITF writer always writes integers in the format `{ '#bigint': 'num' }`. We still parse JS numbers for backwards compatibility.


bugarela (approved):
LGTM!

PR #1166, opened 2023-09-15
Fix our dear #693 type checker bug

bugarela:


examples/cosmwasm/zero-to-hero/vote.qnt
```diff
@@ -491,7 +491,8 @@ module state {
       val poll = state.polls.get(poll_id)
       poll.options.listForall(option =>
         val optionKey = option._1
-        val optionSum = option._2
+        // FIXME: Type annotation below is a workaround, inferred type is too general
+        val optionSum: int = option._2
```
bugarela, line 495: This specific case seems different from the others. I'll look at it in a separate issue (to be created). Every other FIXME of this kind has been removed :smile:

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   shadowed a nested name from the imported module (#802)
 - Fixed a problem where tests were ignored if they are not defined directly in
   the main module - that is, they were imported (#1161)
+- Fixed a type checker bug where the inferred type was to general for nested
```
thpani, line 28:
```suggestion
- Fixed a type checker bug where the inferred type was too general for nested
```

quint/src/types/simplification.ts
```diff
@@ -0,0 +1,42 @@
...
+import { ConcreteRow } from '../ir/quintTypes'
+import { TypeScheme } from './base'
+
+/**
+ * Simplifies a type scheme by flattening all type rows.
+ *
+ * A row like `{ a: int | { b: bool | r } }` becomes
+ * `{ a: int, b: bool | r }`.
+ *
+ * @param typeScheme - The type scheme to me simplified
```
thpani, line 25:
```suggestion
 * @param typeScheme - The type scheme to be simplified
```

thpani (commented):
This is amazing, great work!

Changes to the typechecker LGTM.

Could you explain where we need the row type simplifier?
afaik, nested rows are already flattened on the Apalache side:
https://github.com/informalsystems/apalache/blob/315d43ad3afca3ed3abbb5e8f2610e3bd9bc9929/tla-io/src/main/scala/at/forsyte/apalache/io/quint/QuintTypeConverter.scala#L60-L73

If also pushed a small commit to get rid of a reference to #693 in one of the RFCs.

bugarela:
> Could you explain where we need the row type simplifier? afaik, nested rows are already flattened on the Apalache side: https://github.com/informalsystems/apalache/blob/315d43ad3afca3ed3abbb5e8f2610e3bd9bc9929/tla-io/src/main/scala/at/forsyte/apalache/io/quint/QuintTypeConverter.scala#L60-L73

Interesting, I didn't realize we are already doing this in Apalache.

So... I've added this simplification step because I was hitting an error where Apalache complained about mismatching row types, something like "expected Var or Empty, found Row", which I instantly interpreted as nested rows and jumped to a fix. However, I can't reproduce that anymore :thinking:. Perhaps it was caused by some intermediate state of my other fix or something like that.

While trying to decide if this simplification is worth keeping, I noticed we do similar processes both before printing rows and before unifying rows. So I'll extract it from this PR and just add it in a separate cleanup PR accounting for all these scenarios.

thpani (approved):
Excellent, LGTM!

PR #1169, opened 2023-09-19
Log Apalache distribution-related `verify` messages only at --verbosity >=4

### LINKED ISSUE #1121: Make `quint verify` user friendly
I really like the fact that `quint verify` is working on my example right now, barring #693. However, the current output is not very user-friendly:

```
$ quint verify --max-steps=10 --invariant=noBeepingWhenHeating kettleTest.qnt
Couldn't connect to Apalache, downloading latest supported release
Using existing Apalache distribution in /Users/igor/.quint/apalache-dist-v0.42.0
Launching Apalache server
Shutting down Apalache server
```

These messages are in principle interesting on a first run, but they are always the same. Moreover, whatever spec is getting is checked successfully, the messages stay the same. It would be much better to have the following:

 1. Introduce the option `quint verify --verbosity` similar to `quint run` and print the above messages only on verbosity greater than 4. We have to figure out the levels. I am just following the Java log levels, where 4 would correspond to DEBUG.

1. On success, produce messages similar to `quint run`, which an experienced user may mute via `--verbosity=0`:

```
[ok] No violation found (491ms).
You may increase --max-steps to obtain better guarantees.
Use --verbosity to produce more (or less) output.
```

thpani:
To avoid cluttering output when invoking `quint verify`, log messages related to obtaining and launching Apalache only at `--verbosity` >=4.

Closes #1121

### DIFF
--- a/quint/apalache-dist-tests.md
+++ b/quint/apalache-dist-tests.md
@@ -31,12 +31,10 @@ quint verify ../examples/language-features/booleans.qnt | \
 
 <!-- !test out server not running -->
 ```
-Downloading Apalache distribution...
-Calling Apalache...
+Downloading Apalache distribution... done.
 [ok] No violation found (duration).
 You may increase --max-steps.
 Use --verbosity to produce more (or less) output.
-Calling Apalache...
 [ok] No violation found (duration).
 You may increase --max-steps.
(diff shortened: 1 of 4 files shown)


konnov (approved):
Looks great! Thanks for fixing it!

PR #1171, opened 2023-09-19
Upgrade Github actions

thpani:
Bumps Github actions `actions/checkout@v2` and `actions/setup-node@v2` to `v4` and `v3`, respectively.

Resolves the following deprecatation warning:

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, actions/setup-node@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,8 +22,8 @@ jobs:
     strategy:
       fail-fast: false
     steps:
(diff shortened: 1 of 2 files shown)


shonfeder (approved):
Thanks!

PR #1176, opened 2023-09-20
Pin antlr4ts version

bugarela:
Hello :octocat: 

@rnbguy pointed out that version `0.5.0-dev` of `antlr4ts` is incompatible with quint, and might be matched by our current package specification (according to our understanding of [SemVer item 11.4](https://semver.org/#spec-item-11)). Therefore, I'm pinning this dependency to be safe. 

Update: I used the tool `semver` to check if it really matches, and it does:
```sh
$ semver -r '^0.5.0-alpha.4' '0.5.0-dev'
0.5.0-dev
```

### DIFF
--- a/quint/package-lock.json
+++ b/quint/package-lock.json
@@ -16,7 +16,7 @@
         "@sweet-monads/maybe": "~3.2.0",
         "@types/line-column": "^1.0.0",
         "@types/seedrandom": "^3.0.4",
-        "antlr4ts": "^0.5.0-alpha.4",
+        "antlr4ts": "0.5.0-alpha.4",
         "chalk": "^4.1.2",
         "eol": "^0.9.1",
         "immutable": "^4.3.0",
--- a/quint/package.json
+++ b/quint/package.json
@@ -55,7 +55,7 @@
     "@sweet-monads/maybe": "~3.2.0",
     "@types/line-column": "^1.0.0",
     "@types/seedrandom": "^3.0.4",
-    "antlr4ts": "^0.5.0-alpha.4",
+    "antlr4ts": "0.5.0-alpha.4",
     "chalk": "^4.1.2",
     "eol": "^0.9.1",
     "immutable": "^4.3.0",


thpani (commented):
afaiu, it would not select `-dev`, because it's not published under the `latest` tag?

```
~/s/quint ❯❯❯ npm show antlr4ts
[...]

dist-tags:
alpha: 0.4.1-alpha.0          canary: 0.5.0-alpha.8565fd2b  latest: 0.5.0-alpha.4         
burt: 0.5.0-dev               experiment: 0.4.0-burt.2
```

https://docs.npmjs.com/cli/v10/commands/npm-dist-tag#purpose

bugarela:
@thpani I think that would be true only if we had "latest" in package.json, right? These are the steps @rnbguy sent me to reproduce the issue:

1. `rm package-lock.json` file or node_modules directory
1. change `"antlr4ts": "^0.5.0-alpha.4"` to `"antlr4ts": "^0.5.0-dev"`
1. `npm install` - this will fail because it pulled `antlr4ts@0.5.0-dev` which is not compatible
1. change back to `"antlr4ts": "^0.5.0-alpha.4"`
1. try `npm install` now - and it will still fail as it will still use `antlr4ts@0.5.0-dev`

thpani:
Hmmm, but I guess that only happens because this step

> 3. npm install - this will fail because it pulled antlr4ts@0.5.0-dev which is not compatible

will put the `-dev` tag into the lock file?

If one starts with our package.json, it should not happen, I believe:

> By default, the `latest` tag is used by npm to identify the current version of a package, and `npm install <pkg>` (without any `@<version>` or `@<tag>` specifier) installs the `latest` tag.

bugarela:
Hmm, I think you're right. I tried deleting my `package-lock.json` and running `npm install` again, and it doesn't fetch `0.5.0-dev` with the current setting.

I'll close this then, we can reopen if needed.

Issue #381, opened 2022-11-30
Infer effects for non-determinism of `oneOf`


konnov:
As raised in #377. This needs additional thought and work on the effects checker.

bugarela:
@shonfeder can we do some warning on apalache side for this, while this is not done properly on quint?

PR #1178, opened 2023-09-21
Fix and update ewd840 spec

### LINKED ISSUE #581: Typechecking ewd840
I just found that our integration test on typechecking ewd840 was pointing to Paxos. Typechecking ewd840 does not work:

```
not ok 9 EWD840 - Types & Effects
...
temporal SpecWFNext = Init and always(Next.orKeep(vars) and Next.weakFair(vars))
```

Temporarily disabled this test. It can be fixed later.

bugarela:
Hello :octocat: 

This fixes #581 by:
  1. Commenting out property `AllNodesTerminateIfNoMessages` which has a negation of an action, not allowed by quint's effect system
  2. Commenting out `and init` section of `SpecWFstep` since `init` is an action in quint and therefore cannot be used as a predicate

We should think about whether we want so support this stuff in quint and, if so, how.

In order to make this work with `verify`, I also had to:
  - Fix one error where `oneOf()` was called in a `val` (and not `nondet`). See https://github.com/informalsystems/quint/issues/381
  - Change how the range is computed in `Inv` (copied from `EWD840.tla` in the apalache project), as apalache doesn't support non constant values in ranges (`tpos`).

I also made cosmetic changes in comments and renamed `Node` -> `Nodes`.


examples/classic/distributed/ewd840/ewd840.qnt
```diff
@@ -115,109 +104,100 @@ action Deactivate(i) = all {
...
+/// The following property says that eventually all nodes will terminate
+/// assuming that from some point onwards no messages are sent. It is not
+/// supposed to hold: any node may indefinitely perform local computations.
+/// However, this property is verified if the fairness condition WF_vars(step)
+/// is used instead of only WF_vars(System) that requires fairness of the
+/// actions controlled by termination detection.
+///
+/// Quint translation note: we cannot add `init` to this property since, in quint,
+/// it is an action - not a predicate
+temporal SpecWFstep = always(step.orKeep(vars) and step.weakFair(vars)) // and init
```
thpani, line 163: I guess we should check with @konnov on how to treat `init` in temporal formulas, and note this here with a reference to an issue.

examples/classic/distributed/ewd840/ewd840.qnt
```diff
@@ -115,109 +104,100 @@ action Deactivate(i) = all {
...
+/// Next-state relation: disjunction of above actions
+action step = any {
   System,
   Environment,
 }
 
 val vars = (active, color, tpos, tcolor)
 
-// temporal Spec = Init and always(Next.orKeep(vars) and System.weakFair(vars))
+// temporal Spec = init and always(step.orKeep(vars) and System.weakFair(vars))
```
thpani, line 123: I assume we already had this commented out because of `init`?
Probably good to treat it the same as below (i.e., just comment out `init`, or whatever we decide to do there)

thpani (commented):
LGTM, but I would like to get @konnov's view on commenting out `init`.

examples/classic/distributed/ewd840/ewd840.qnt
```diff
@@ -115,109 +104,100 @@ action Deactivate(i) = all {
...
+/// However, this property is verified if the fairness condition WF_vars(step)
+/// is used instead of only WF_vars(System) that requires fairness of the
+/// actions controlled by termination detection.
+///
+/// Quint translation note: we cannot add `init` to this property since, in quint,
+/// it is an action - not a predicate
+temporal SpecWFstep = always(step.orKeep(vars) and step.weakFair(vars)) // and init
+
+// The following property is not expressible in quint (for now), as it negates an action
+// temporal AllNodesTerminateIfNoMessages = (
```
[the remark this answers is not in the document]
bugarela: Ah, yes, I haven't noticed that. Will fix.

examples/classic/distributed/ewd840/ewd840.qnt
bugarela: @konnov pinging you here since you probably missed it!

konnov: ah, I think it is the fifth time I am staring at it. Let me open an issue.

konnov: Let's maybe refer to the issue and comment out the whole formula?

bugarela: Thanks! Ok, will do

thpani (approved):
Looks great! Thanks for revising this spec 🎉

PR #1179, opened 2023-09-22
one line fix

konnov:
A one-liner that fixes the parameters in a Tendermint instance: The same process should not be both correct and faulty

### DIFF
--- a/examples/cosmos/tendermint/TendermintModels.qnt
+++ b/examples/cosmos/tendermint/TendermintModels.qnt
@@ -12,7 +12,7 @@ module TendermintModels {
   ) as n4_f1 from "./TendermintTest"
 
   import TendermintTest(
-    Corr = Set("p1", "p2", "p3"),
+    Corr = Set("p1", "p2"),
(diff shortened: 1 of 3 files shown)


examples/cosmos/tendermint/TendermintModels.qnt
```diff
@@ -12,7 +12,7 @@ module TendermintModels {
   ) as n4_f1 from "./TendermintTest"
 
   import TendermintTest(
-    Corr = Set("p1", "p2", "p3"),
+    Corr = Set("p1", "p2"),
```
shonfeder, line 15: Should there be some property/assert that says `Corr` and `Faulty` need to be disjoint?

shonfeder (approved):
LGTM, but integration tests are failing.

examples/cosmos/tendermint/TendermintModels.qnt
konnov: Ah, good, point. I have added an assumption

PR #1188, opened 2023-09-28
Add --random-transitions to run Apalache symbolic simulator

### LINKED ISSUE #1185: Add support for `apalache simulate`
Currently, `quint verify` calls `apalache-mc check`. It would be great, if we could also call `apalache-mc simulate`, so we could utilize Apalache even for relative complex specs.

Since `quint run` does random simulation, it would be confusing to call this mode "simulation". Perhaps, we could call it something like `quint verify --randomized`. The essential difference is that this mode would choose transitions at random, but evaluate them symbolically, whereas `quint run` chooses transitions at random and evaluates them concretely.

@thpani wdyt?

thpani:
Add a `--random-transitions` flag to `verify`, to run Apalache symbolic simulator instead of BMC.

Closes #1185 

- [ ] ~Tests added for any new code~

### DIFF
--- a/quint/src/cliCommands.ts
+++ b/quint/src/cliCommands.ts
@@ -644,6 +644,9 @@ export async function verifySpec(prev: TypecheckedStage): Promise<CLIProcedure<V
       next: 'q::step',
       inv: args.invariant ? ['q::inv'] : undefined,
       'temporal-props': args.temporal ? ['q::temporalProps'] : undefined,
+      tuning: {
+        'search.simulation': args.randomTransitions ? 'true' : 'false',
+      },
     },
   }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 
 ### Added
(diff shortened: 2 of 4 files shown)


quint/src/cliCommands.ts
```diff
@@ -644,6 +644,9 @@ export async function verifySpec(prev: TypecheckedStage): Promise<CLIProcedure<V
       next: 'q::step',
       inv: args.invariant ? ['q::inv'] : undefined,
       'temporal-props': args.temporal ? ['q::temporalProps'] : undefined,
+      tuning: {
+        'search.simulation': args.randomTransitions ? 'true' : 'false',
+      },
```
shonfeder, line 649: :ok_hand:

shonfeder (approved):
Lovely how easy it is to expose this kind if extension :)

I guess an alternative would be to just document how to use `--aplache-config` for this setting. But I can see this being used enough to warrant the shorthand

quint/src/cliCommands.ts
konnov: I second that. It's truly amazing :)

konnov:
I am already using it! Thanks a lot!

Issue #1091, opened 2023-08-04
Weird behaviour of 'boolean' versus 'boolean == true'


p-offtermatt:
Hi!

I get weird behaviour with an action.

This is fine:

```
module weird_parsing {
    var channel: List[int]

    action CoolAction(boolean: bool): bool = 
    any {
        all {
            boolean == true,
            channel' = channel
        },
        all {
            not(boolean),
            channel' = channel
        }
    }
}
```

However, this gives me an error:

```
module weird_parsing {
    var channel: List[int]

    action CoolAction(boolean: bool): bool = 
    any {
        all {
            boolean,
            channel' = channel
        },
        all {
            not(boolean),
            channel' = channel
        }
    }
}
```

Note the difference is simply `boolean` breaks  vs `boolean == true` works.
The error I get is 
```
Expected [channel] and [] to be the same
Trying to unify entities ['channel'] and []
Unifying substitutions with same name: v3
Composing substitutions [e0 |-> Read['channel'] & Update['channel'],v1 |-> 'channel',v0 |-> 'channel',e1 |-> Read[v2, 'channel'] & Update['channel'],v4 |-> 'channel',v5 |-> 'channel',e_boolean_101335 |-> Read[v2],e2 |-> Read[v6] & Temporal[v7],v7 |-> ,v3 |-> ,v6 |-> v2,v6 |-> v6,v7 |-> v7,e3 |-> Read['channel'] & Update['channel'],v9 |-> 'channel',v8 |-> 'channel',e4 |-> Read[v2, 'channel'] & Update['channel'],v12 |-> 'channel',v13 |-> 'channel',v11 |-> ,v10 |-> v2] and [e5 |-> Read[v2, 'channel'] & Update['channel', 'channel'],v16 |-> v2, 'channel',v3 |-> 'channel',v14 |-> v2, 'channel',v15 |-> 'channel', 'channel']
Trying to infer effect for operator application in actionAny(actionAll(boolean, assign(channel, channel)), actionAll(not(boolean), assign(channel, channel)))
quint(QNT000)
```

bugarela:
Weird indeed! I'm almost sure that the reason why `boolean == true` makes it work is that the equality operation constrains which effects `boolean` can assume, while in the second example it is pretty much unconstrained until the `not()` application. We should be able to unify those either way, of course, this is definitely a bug. I'll look into it, thanks for reporting!

Is this issue blocking you in any manner?

p-offtermatt:
[...]

PR #1203, opened 2023-10-03
Fix effect substitutions, free names and unification of entities

bugarela:
Hello :octocat:

I finally figured out this substitution composition stuff! This should stabilize the effect checker a lot. Closes https://github.com/informalsystems/quint/issues/1091 as a side effect. The main motivation for the fixes were problems encountered by @p-offtermatt (example in tests).

Turns out that substitution composition is really simple, and I was complicating it for the wrong reasons. As long as substitutions are properly applied to things before running unification on them, we can go back to the simple composition that we had at first, which is the normal thing from the papers.

This PR additions:
1. When trying to unify effect entities like `[v0, v1, v2]` and `[v1, v2]`, instead of throwing an error because we don't do unification of sets, we simplify them to `[v0]` and `[]` which easily results in `v0 |-> `.
2. Free name tracking is now the same as the one for types, since I recently made advancements on the correctness of that. I just copied the code for now, we can try do to a generalization at some point.
3. `applySubstitution` is now called in two new places on inferring effects for applications, ensuring the effects being unified are both updated with the latest substitutions. This makes it so the substitutions resulting from unification can compose with the existing ones with the classic composition algorithm, and therefore I removed the extra complexity I had added to our `compose` (and to `applySubstitutionsToSubstitutions`).
4. `applySubstitution` was fixed, as it was not working properly for substitutions like `[a |-> b, b |-> c]`.
5. internal fresh vars are now prefixed with underscore so they don't collide with the names generated while printing - this was causing bugs now that all the rest is fixed.

From this list, (3), (4) and (5) should also be applicable for types. I'll do that as a followup and check if it solves any of our open issues for the type system.


quint/src/effects/base.ts
```diff
@@ -323,19 +324,21 @@ export function unifyEntities(va: Entity, vb: Entity): Either<ErrorTree, Substit
...
+  } else if (v1.kind === 'union' && v2.kind === 'concrete') {
+    return mergeInMany(v1.entities.map(v => unifyEntities(v, v2)))
+      .map(subs => subs.flat())
+      .mapLeft(err => buildErrorTree(location, err))
+  } else if (v1.kind === 'concrete' && v2.kind === 'union') {
+    return unifyEntities(v2, v1)
+  } else if (v1.kind === 'union' && v2.kind === 'union') {
+    if (intersectionWith(v1.entities, v2.entities, isEqual).length > 0) {
+      const s1 = { ...v1, entities: differenceWith(v1.entities, v2.entities, isEqual) }
+      const s2 = { ...v2, entities: differenceWith(v2.entities, v1.entities, isEqual) }
```
thpani, line 338: Simply an optimization, but given that we already compute the intersection, we could use it to compute the difference and avoid (re-)iterating the entire arrays?

quint/src/effects/base.ts
```diff
@@ -323,19 +324,21 @@ export function unifyEntities(va: Entity, vb: Entity): Either<ErrorTree, Substit
...
+
+      // There was an intersection, try to unify the remaining entities
+      return unifyEntities(s1, s2)
```
thpani, line 341: Not sure if I'm following the code correctly, but won't this call just end up in the same branch (`v1.kind === 'union' && v2.kind === 'union'`) and then return the `left` below?

quint/src/effects/substitutions.ts
```diff
@@ -86,7 +87,14 @@ export function applySubstitution(subs: Substitutions, e: Effect): Either<ErrorT
     }
   }
 
-  return result.map(simplify)
+  return result.map(simplify).chain(r => {
+    if (!isEqual(r, e)) {
+      // Keep reaplying the substitutions until the effect is unchanged.
```
thpani, line 92:
```suggestion
      // Keep re-applying the substitutions until the effect is unchanged.
```

quint/src/effects/substitutions.ts
```diff
@@ -86,7 +87,14 @@ export function applySubstitution(subs: Substitutions, e: Effect): Either<ErrorT
...
+      // Useful when substitutions have a transitive pattern [ a |-> b, b |-> c ]
+      return applySubstitution(subs, r)
```
thpani, line 94: Just to make sure: substitutions can never be cyclic?

Also, if not: do we really need to compute a fixed point here, or can we sort the substitutions topologically and apply them all at once?

thpani (commented):
Code LGTM! I have some more algorithmic questions though

quint/src/effects/base.ts
bugarela: Not if the `flattenUnions` calls in the first two lines of this function are able to perform some simplification. Might not be so obvious from the name, but `flattenUnions` actually transforms `[a]` in `a` (which are equivalent by definition, but the unification algorithm can only handle the latter).

quint/src/effects/base.ts
bugarela: Yep, I meant to do that after my initial validation and ended up forgetting about it. Thanks.

quint/src/effects/substitutions.ts
thpani: > However, I don't see a clear benefit of doing that over this?

idk, I'm always for avoiding recursion if things can be achieved in a single call.
I guess here the overhead does not matter much, because the arrays are small, but even so, small delays add up and I'd avoid them, especially for compile-time checks.

thpani (approved):

quint/src/effects/substitutions.ts
bugarela: Makes sense. I'll keep this for now and improve it a performance-focused pass in the future. There are way more pressing performance problems and, then, I'll find a way to be very careful to not make regressions with the optmizations.

PR #1206, opened 2023-10-05
Add syntax for match and variant operators

### LINKED ISSUE #1082: Add syntax for sum-types

shonfeder:
Closes #1082

This is an alternative to #1093, based on @konnov's preference to use the
existing convention of encoding language constructs as builtin operators with
special interpretations instead of extending our IR.

It adds support for:

- Generating variant value constructors from type definitions
- Parsing match expressions

It also includes some fixes and clean up from the last PR on sum types.

A notable alteration:

I am replacing the existing `match` syntax used for the unsupported record unions with the superseding sum-type syntax agreed upon in #1062. This means a "regression" in our `options.qnt` example, since this used the partially supported union syntax. However, afaiu, this really just reflects more accurately the state of our support for this kind of construct at the moment.


shonfeder:
I'll need to fix the tests before merging. The failing integration tests are straight forward, but I'm also getting a failure on the unit tests due to

```
expected source map to contain all ids
```

I'll have to figure out how and why this can be failing.

quint/src/graphics.ts
```diff
@@ -187,7 +187,7 @@ export function prettyQuintType(type: QuintType): Doc {
       return group([text('{ '), prettyRow(type.fields), text('}')])
     }
     case 'sum': {
-      return group([text('{ '), prettySumRow(type.fields), text('}')])
+      return prettySumRow(type.fields)
```
shonfeder, line 190: This is cleanup from the previous PR adding the sum types. The were wrapped in an extra `{...}`

quint/src/parsing/ToIrListener.ts
```diff
@@ -373,44 +377,88 @@ export class ToIrListener implements QuintListener {
...
-    // I.e., we interpert a variant `A` as `A({})`.
-    if (poppedType === undefined) {
-      const id = this.idGen.nextId()
-      this.sourceMap.set(id, this.loc(ctx))
-      fieldType = unitValue(id)
-    } else {
-      fieldType = poppedType
-    }
-
+    //   const poppedType = this.popType().value
```
bugarela, line 459:
```suggestion
    //
```

quint/src/parsing/ToIrListener.ts
```diff
@@ -887,45 +935,79 @@ export class ToIrListener implements QuintListener {
...
+      params = [{ name, id: this.getId(variant) }]
+    } else {
+      throw new Error('impossible: either _wildCardMatch or _variantMatch must be present')
     }
-    // construct the match expression and push it in exprStack
-    this.pushApplication(ctx, 'unionMatch', matchArgs)
+    const labelStr: QuintStr = { id: caseId, kind: 'str', value: label }
+    const elim: QuintLambda = { id: caseId, kind: 'lambda', qualifier: 'def', expr: caseExpr, params }
+    return [labelStr, elim]
+    // return acc.concat([labelStr, elim])
```
bugarela, line 1010:
```suggestion
```

quint/src/util.ts
```diff
@@ -29,3 +30,17 @@ import JSONbig from 'json-bigint'
 export function unreachable(object: never): never {
   throw new Error(`impossible: non-exhuastive check should fail during type checking ${JSONbig.stringify(object)}`)
 }
+
+/**  A wrapper around lodash zip that ensures all zipped elements are defined
```
bugarela, line 34: I will use this! Thanks

bugarela: Ah, I noticed this was moved, but regardless, I didn't know/remember we have this :smile:

quint/src/generated/Quint.g4
```diff
@@ -176,6 +174,12 @@ expr:           // apply a built-in operator via the dot notation
         |       '{' expr '}'                                        # braces
         ;
 
+// match e { A(a) => e1 | B => e2 | C(_) => e3 | ... | _ => en }
+matchSumExpr: MATCH expr '{' '|'? matchCase+=matchSumCase ('|' matchCase+=matchSumCase)* '}' ;
+matchSumCase: (variantMatch=matchSumVariant | wildCardMatch='_') '=>' expr ;
+matchSumVariant
+    : (variantLabel=simpleId["variant label"]) ('(' (variantParam=simpleId["match case parameter"] | '_') ')')? ;
```
bugarela, line 181: How are these labels used? I see they have spaces, so I suppose it's only for error reporting to users? If so, that's really great - we should really use them more.

bugarela (approved):
This syntax looks super neat!

quint/src/parsing/ToIrListener.ts
shonfeder: Thanks for catching these!

quint/src/util.ts
shonfeder: It was added in 99910e3f3d5523410b92d56ee6106546435c9793 :)

quint/src/generated/Quint.g4
shonfeder: Yep, the string arguments are used for error reporting in

https://github.com/informalsystems/quint/blob/ba946ad88befae93404bef9fd88ac1a283451722/quint/src/generated/Quint.g4#L243-L254

To produce messages like

```
>>> {A::B::c: 3}
syntax error: error: [QNT008] Identifiers in a record cannot be qualified with '::'. Found A::B::c.
{A::B::c: 3}
        ^


>>> type T = A::B(int)
syntax error: error: [QNT008] Identifiers in a variant label cannot be qualified with '::'. Found A::B.
type T = A::B(int)
             ^
```

The ANTLR labels like `variantLabel` and `variantParam`, on the other hand, are used to provide named accessors for the parsed data on the rhs of the `=`, used in

https://github.com/informalsystems/quint/blob/ba946ad88befae93404bef9fd88ac1a283451722/quint/src/parsing/ToIrListener.ts#L996-L1003

shonfeder:
Thank you for the careful review!

Issue #868, opened 2023-05-05
Bad error message when `pure val` is used in a `pure def`


lasarojc:
There was an error and I fixed it thanks to the error message, but the second part of the error message `but operator `hasEntry` doesn't read or update any state variable` seems to contradict the first `pure val operators may not interact with state variables`.

> error: [QNT200] pure val operators may not interact with state variables, but operator `hasEntry` doesn't read or update any state variable. Use pure def instead.

konnov:
@bugarela wdyt?

bugarela:
Ah, I think thats the case where `pure val` is used instead of `pure def`, right? I saw this once but forgot to open an issue. We should improve the message for this case.

lasarojc:
Yes, @bugarela. that was the case. I should have included it in the description.

PR #1208, opened 2023-10-05
Special error for `val` vs `def` mode errors

bugarela:
Hello :octocat: 

Fixes https://github.com/informalsystems/quint/issues/868

This improves error messages when `pure val` is used instead of `pure val`, and when `val` is used instead of `def`.

Before:
```
/home/gabriela/projects/quint/local/test.qnt:40:3 - error: [QNT200] val operators may only read state variables, but operator `foo` reads variables 'x'. Use def instead.
40:   val foo(p: int): bool = {
      ^^^^^^^^^^^^^^^^^^^^^^^^^
41:     x
    ^^^^^
42:   }
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

After:
```
/home/gabriela/projects/quint/local/test.qnt:40:3 - error: [QNT200] val operators may not have parameters, but operator `foo` has 1 parameter. Use def instead.
40:   val foo(p: int): bool = {
      ^^^^^^^^^^^^^^^^^^^^^^^^^
41:     x
    ^^^^^
42:   }
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Also fixes some cases where `val` was used instead of `def`, but no error was flagged.


shonfeder (approved):
looks good :)

PR #1210, opened 2023-10-09
remove the tuple unpacking code

konnov:
This is a follow-up to #1202. Since tuples are rewritten by the parser now, there is no need for special treatment in the simulator. This has also unblocked the old code in the tutorial 3.

### DIFF
--- a/doc/lang.md
+++ b/doc/lang.md
@@ -804,7 +804,7 @@ The latter form `((x, y)) => e2` is equivalent to:
 )
 ```
 
-As a result, the form `((x_1, ..., x_n)) => e_n` is a syntax sugar for tuple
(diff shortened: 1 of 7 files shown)


doc/lang.md
```diff
@@ -790,6 +790,23 @@ _ => e
...
+
+```bluespec
+(t =>
+  val x = t._1
+  val y = t._2
+  e2
+)
+```
+
+As a result, the form `((x_1, ..., x_n)) => e_n` is a syntax sugar for tuple
```
thpani, line 807:
```suggestion
As a result, the form `((x_1, ..., x_n)) => e_n` is syntax sugar for tuple
```

thpani (approved):
Ahh, I thought I'd found all the unpacking code 😂 
Good that you looked it over, thanks for the cleanup!

Issue #631, opened 2023-02-16
Group error classes in one place


konnov:
I am trying to produce an error in the CLI. Currently, we have at least three error classes introduced in various places:
 
 https://github.com/informalsystems/quint/blob/1d67fb43758aa08195fe1a80e1385328d884f33e/quint/src/quintIr.ts#L26
 https://github.com/informalsystems/quint/blob/1d67fb43758aa08195fe1a80e1385328d884f33e/quint/src/quintError.ts#L18
 https://github.com/informalsystems/quint/blob/1d67fb43758aa08195fe1a80e1385328d884f33e/quint/src/quintParserFrontend.ts#L40

We also have adapter functions, e.g.:

https://github.com/informalsystems/quint/blob/1d67fb43758aa08195fe1a80e1385328d884f33e/quint/src/cliCommands.ts#L147

It's getting harder to navigate across error handling. Shall we group all three classes (if we need all three of them) and their adapters in a single file, e.g., `quintIr.ts`?

@bugarela, @shonfeder, wdyt?

bugarela:
Yep, we should centralize those in `quintError` (`quintIr` would be too flooded with error stuff). There's also `ErrorTree`. We should also ensure we are using `QuintError` to represent errors and `ErrorMessage` should be strictly an output format.

PR #1211, opened 2023-10-09
`QuintError` everywhere

### LINKED ISSUE #1014: Static analysis passes should return lists of `QuintError`
Some return `Map<bigint, QuintError>`, some return `[bigint, QuintError][]`, but they should all return `QuintError[]`. In #1011, I've added a `referece` property to `QuintError` to carry the id, so there is no need to return the ids elsewhere.

bugarela:
Hello :octocat: 

Closes #1014, closes https://github.com/informalsystems/quint/issues/631

This isolates `ErrorMessage` type only to the CLI layer (`cliCommands` and `repl`), while everything else uses `QuintError`. `IrErrorMessage` is completely removed.

For that, I had to add new error codes for a lot of compiler errors. We still need to classify errors from parser and type/effect checking - for all, they are `QNT000` by default.


quint/src/parsing/parseErrors.ts
```diff
@@ -0,0 +1,36 @@
+import { QuintError } from '../quintError'
+
+export function undercaseTypeError(id: bigint, name: string): QuintError {
```
konnov, line 3:
```suggestion
export function lowercaseTypeError(id: bigint, name: string): QuintError {
```

konnov (approved):
That's a great piece of refactoring! Thanks for taking care of it!

PR #1213, opened 2023-10-10
print the seed on success

### LINKED ISSUE #1160: `quint run` should print the seed
Sometimes, I would like to repeat the same example, e.g., to generate a json trace of it. We should print the seed when an example is found, not just when there is an error.

konnov:
Closes #1160. This is a small change that prints the seed even if no bug was found. This is quite useful for reproducing long-running executions.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Error messages for `val` vs `def` and `pure val` vs `pure def` errors are clearer (#1208)
+- `quint run` print the seed even if no bug was found (#1213)
```
thpani, line 14:
```suggestion
- `quint run` prints the random seed even if no bug was found (#1213)
```

thpani (approved):
LGTM

PR #1214, opened 2023-10-10
Fix pure defs and vals in tutorial 4

### LINKED ISSUE #1157: Use `pure val` in Tutorial 4
As notice by @pdini, we are using `val` for no good reason. We should only use `pure val` and `pure def` in that tutorial.

konnov:
Closes #1157. In tutorial 4, `pure def` and `pure val` are used where they should be used.


thpani (approved):
👍

PR #1220, opened 2023-10-11
Error recovery between parsing phases

bugarela:
Hello :octocat:

This PR makes it so we accumulate errors between parsing phases, instead of giving up after the first phase that results in error. In order to do that, we stop using `Either` monads, and instead using records with an `errors` field that accumulates the errors.

There are also some changes to `ToIrListener.ts` so we have some error recovery - that is, we produce valid IR even if parsing fails. I'm trying to play with some specs to see if I can cause any of the assertions to fail, but for some reason it is looking pretty stable after my few changes.

There's also a change in the JSON assets for testing, since now we are able to produce `modules` even if there are parsing errors.

Example:
```bluespec
module A {
  val syntaxErrorEx = 1(2)
  val nameErrorEx = syntaxErrorEx + missing
}
```
yields

```
/home/gabriela/projects/quint/local/test.qnt:2:24 - error: [QNT000] extraneous input '(' expecting {'}', 'const', 'var', 'assume', 'type', 'val', 'def', 'pure', 'action', 'run', 'temporal', 'import', 'export', DOCCOMMENT}
2:   val syntaxErrorEx = 1(2)
                          ^

/home/gabriela/projects/quint/local/test.qnt:3:37 - error: [QNT404] Name 'missing' not found
3:   val nameErrorEx = syntaxErrorEx + missing
                                       ^^^^^^^

error: parsing failed
```


quint/test/parsing/quintParserFrontend.test.ts
```diff
@@ -47,50 +46,30 @@ function parseAndCompare(artifact: string): void {
...
-    )
-  } else if (phase2Result.isRight()) {
-    const { modules: modules2, sourceMap } = phase2Result.value
-    const expectedIds = modules2.flatMap(m => collectIds(m)).sort()
-    // Phase 1-2 succededed, check that the source map is correct
-    assert.sameDeepMembers([...sourceMap.keys()].sort(), expectedIds, 'expected source map to contain all ids')
+  const phase1Result = parsePhase1fromText(gen, readQuint(artifact), mainPath.toSourceName())
+  const phase2Result = parsePhase2sourceResolution(gen, resolver, mainPath, phase1Result)
+
+  // const { modules: modules2, sourceMap } = phase2Result
```
konnov, line 52:
```suggestion
```

remove a left-over?

konnov (approved):
This is great refactoring! I believe it makes it easier to understand what's going on. Also, the ability to accumulate errors is nice

PR #1221, opened 2023-10-13
extracting error explanation into a separate parser

konnov:
This is work in progress. I am experimenting with better ways to explain errors with antlr4.


konnov:
There is no reason for keeping this PR open. I've extracted some useful parts out of it.

PR #1228, opened 2023-10-20
[Re-opening] Improve error recoverability in `ToIrListener`

bugarela:
new PR for https://github.com/informalsystems/quint/pull/1223 (which I messed up with a local merge that ruined it forever) - please see that PR's description to understand the changes on this PR.

Also added some further improvements on diagnostics and error reporting, as I experienced some crashing behavior on manual tests last week.


quint/src/parsing/ToIrListener.ts
```diff
@@ -1270,13 +1241,39 @@ export class ToIrListener implements QuintListener {
...
 
-  private undefinedDef(ctx: any): QuintEx {
-    const id = this.getId(ctx)
-    return { id, kind: 'bool', value: true }
+  private undefinedExpr(ctx: any): () => QuintEx {
+    return () => {
+      const id = this.getId(ctx)
+      return { id, kind: 'bool', value: true }
+    }
+  }
```
shonfeder, line 1252: Should we introduce an `UndefinedExpr` node in the IR so we can be very clear about where holes are showing up? I'm worried we could end up with weird bugs where parts of the AST are replaced with `true` and it will be hard for maintainers to find out why that would be happening.

quint/src/parsing/ToIrListener.ts
```diff
@@ -1270,13 +1241,39 @@ export class ToIrListener implements QuintListener {
...
+
+  private undefinedDeclaration(ctx: any): () => QuintDeclaration {
+    return () => {
+      const id = this.getId(ctx)
+      return { id, kind: 'assume', name: '_', assumption: this.undefinedExpr(ctx)() }
```
shonfeder, line 1257: Can we name it something like

```suggestion
      return { id, kind: 'assume', name: `_undefinedDecl${id}`, assumption: this.undefinedExpr(ctx)() }
```

quint/src/parsing/ToIrListener.ts
```diff
@@ -1270,13 +1241,39 @@ export class ToIrListener implements QuintListener {
...
+  private undefinedType(ctx: any): () => QuintType {
+    return () => {
+      const id = this.getId(ctx)
+      return { id, kind: 'bool' }
+    }
+  }
+
+  private undefinedVariant(ctx: any): () => RowField {
+    return () => ({ fieldName: '_', fieldType: this.undefinedType(ctx)() })
   }
```
shonfeder, line 1277: Same question about making undefined forms clearer applies here.

quint/src/parsing/quintParserFrontend.ts
```diff
@@ -116,8 +117,17 @@ export function parsePhase1fromText(
...
+  // wrap this in a try catch, since we are running the parser even if there are
+  // errors after the previous command.
+  try {
+    // walk through the AST and construct the IR
+    ParseTreeWalker.DEFAULT.walk(listener as QuintListener, tree)
+  } catch (e) {
+    if (errors.length === 0) {
+      throw e
+    }
+    // ignore the exception, we already have errors to report
```
shonfeder, line 129: But what if the exception is not obviously explained by the recorded errors? Then we would end up swallowing unknown exceptions, potentially hiding errors.

Instead, I suggest this catch only apply to errors which are from a known family of errors that we are sure will be reflected in the `errors`, and anything else is left to bubble up. In my experience, a catch-all that recovers without a record of the exception inevitably comes back to bite you :D

shonfeder (changes requested):
I have a few questions and minor suggestions.

The only blocking change request in my mind is the exception handling I noted.

quint/src/parsing/quintParserFrontend.ts
bugarela: > But what if the exception is not obviously explained by the recorded errors?

It is not, what is happening is that the generated code breaks internally when the AST is malformed (because of parser errors). The idea is: there is something being reported, you should fix that problem first and then try again. If an exception is thrown after that (which should be impossible), then you can worry about the exception.

It should always be the case that the exception is a direct consequence of one of the errors, and that fixing the errors will make the exception go away.

> Instead, I suggest this catch only apply to errors which are from a known family of errors that we are sure will be reflected in the `errors`

That I cannot do, because I'm introducing this try catch to handle unpredictable (or hard to predict) errors that are internal to the generated code from antlr4ts. Took me some manual testing to find the first case where some error is thrown, and I don't think I could ever find all possible errors that can be thrown.

quint/src/parsing/ToIrListener.ts
bugarela: Hmm that might be a good idea for the next step. For now, the IR with undefined components should only be manipulated in name resolution. After that, we halt the process and report all errors gathered so far. This is not something guaranteed, but I'm pretty confident that we are ensuring that this is the case.

For the next steps, I want the type checker and effect checker to run over this IR as well, and my original plan was to have a check like: if there is an error for the id of this component, don't try to infer type/effect for the component.

Perhaps introducing undefined components to the IR, as you suggest, is a better alternative than relying on the errors. This would be a bigger change tho, as it requires another case in many IR-related functions and tests.

quint/src/parsing/ToIrListener.ts
```diff
@@ -183,19 +194,22 @@ export class ToIrListener implements QuintListener {
...
   exitNondet(ctx: p.NondetContext) {
-    const def = this.declarationStack.pop()
-    const nested = this.exprStack.pop()
+    const def = this.declarationStack.pop() ?? this.undefinedDeclaration(ctx)()
+    const nested = this.exprStack.pop() ?? this.undefinedExpr(ctx)()
+
+    if (def.kind !== 'def') {
+      // only `QuintDef` is allowed in `nondet` expressions
+      return
+    }
```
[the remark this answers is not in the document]
bugarela: This is already ensured by the parser, so we must already have gathered an error for this that will be reported later.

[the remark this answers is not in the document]
shonfeder: Sorry if I'm being a bit dense, but if this is ensured by the parser, then should this not be just an assert? Or are you saying that this is not ensured by the parser, but an erroneous node here will already have an error report from the parser?

quint/src/parsing/quintParserFrontend.ts
shonfeder: I'm not following, maybe we can talk about this a bit tomorrow.

quint/src/parsing/ToIrListener.ts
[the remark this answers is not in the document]
bugarela: In general, the latter. So, if we have an assert, the assertion will throw and omit the properly described error.

For this particular case, I'm not sure if/how this case could be reached, since the grammar requires 'nondet' for this rule. But, if somehow this is possible, the reported error still is much better than an assertion - and there must be a reported error.

quint/src/parsing/ToIrListener.ts
shonfeder: I see. Maybe something to consider for a followup, then.

quint/src/parsing/ToIrListener.ts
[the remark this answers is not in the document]
shonfeder: But, iiuc, if we get here and this is not a `def` them something must be broken with the parser? So assert or exceptions seems justified. Otherwise, e.g., what is something changes upstream that removes generation of an error when this is not a def? Then we would proceed silently and without marking any errer, no?

quint/src/parsing/ToIrListener.ts
bugarela: I like that idea!

quint/src/parsing/ToIrListener.ts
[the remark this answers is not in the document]
shonfeder: Yep, that sounds good. Thanks.

quint/src/parsing/quintParserFrontend.ts
shonfeder: Thanks for talking thru the reasoning here. I am still a bit worried this may hide problems at some point, but less so and not enough to block here.

Perhaps the comment could be expanded to explain the reasoning?

shonfeder (approved):

PR #1230, opened 2023-10-25
Bounded unsigned integer spells

Kukovec:

coauthored w/ @thpani


examples/spells/BoundedUInt.qnt
```diff
@@ -0,0 +1,607 @@
...
+  pure val MIN = 0
+
+  /// The largest value that can be represented by this integer type.
+  pure val MAX = (2^BITS) - 1
+  
+  /// Record-representation of an unsigned bounded integer.
+  /// If the `error` field is nonempty, the record represents an exception of some
+  /// sort happening during computation, such as an overflow.
+  /// Otherwise, the record represents the integer `v`, such that `MIN <= v <= MAX`. 
+  type T = { v: int, error: str }
```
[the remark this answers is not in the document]
shonfeder: I'm wondering if actual use case needs an `error` message string. Would it be possible that having `error: bool` would be more ergonomic and just knowing that there is an error one way or the other would suffice?

examples/spells/BoundedUInt.qnt
```diff
@@ -0,0 +1,607 @@
...
+  /// Unsafe saturating integer addition. 
+  /// Computes `lhs + rhs`, saturating at the numeric bounds instead of overflowing.
+  /// Assumes lhs and rhs are valid (w.r.t. `isValid`).
+  pure def saturatingAddUnsafe(lhs: T, rhs: T): T = {
+    val res = lhs.v + rhs.v
+    { 
+      v: if (res < MAX) res else MAX, 
+      error: "" 
+    }
+  }
```
shonfeder, line 159: Since these don't error out, would it not make more sense to just have it defined over `int`?

examples/spells/BoundedUInt.qnt
```diff
@@ -0,0 +1,607 @@
...
+}
+
+module BoundedUInt_Test {
+  import BoundedUInt.*
+
+  /////////////
+  // CHECKED //
+  /////////////
+
+  pure def MkCInv(x: int, y: int, op: (T,T) => T, expectedErr: str, expectedVal: (int, int) => int): bool = {
```
shonfeder, line 289: - What does `MkCInv` mean? Could we add a comment or a legible name?
- `extectedVal` seems like a slightly confusing name for an operator that is not a value. Would `expectedOp` be clearer?

quint/cli-tests.md
```diff
@@ -260,6 +260,11 @@ fi
 <!-- !test check rareSpells - Syntax/Types & Effects/Unit tests -->
     quint test ../examples/spells/rareSpells.qnt
 
+### OK on test BoundedUint8
+
+<!-- !test check BoundedUint8 - Syntax/Types & Effects/Unit tests -->
+    quint test --main=BoundedUInt8Test ../examples/spells/BoundedUInt.qnt
```
shonfeder, line 266: Tests should get run in the examples dashboard. Whats the added value of adding an additional test here?

shonfeder (commented):
Just a few questions and small suggestions from me :)

Generally looks great.

examples/README.md
```diff
@@ -94,6 +94,7 @@ listed without any additional command line arguments.
 | [solidity/SimpleAuction/SimpleAuction.qnt](./solidity/SimpleAuction/SimpleAuction.qnt) | :white_check_mark: | :white_check_mark: | :white_check_mark: | N/A[^nostatemachine] |
 | [solidity/SimplePonzi/simplePonzi.qnt](./solidity/SimplePonzi/simplePonzi.qnt) | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
 | [spells/basicSpells.qnt](./spells/basicSpells.qnt) | :white_check_mark: | :white_check_mark: | :white_check_mark: | N/A[^nostatemachine] |
+| [spells/BoundedUInt.qnt](./spells/BoundedUInt.qnt) | :white_check_mark: | :white_check_mark: | :x: | N/A[^nostatemachine] |
```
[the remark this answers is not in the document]
Kukovec: IIRC @bugarela changed the checkmarks file manually, could just be an oversight.

examples/spells/BoundedUInt.qnt
```diff
@@ -0,0 +1,607 @@
...
+    x.v >= MIN,
+    x.v <= MAX
+  }
+
+  /// Given a record representation of a bounded unsigned integer, returns
+  /// true iff the record value lies in the `[MIN, MAX]` interval, and the error field is empty.
+  pure def isValid(x: T): bool = and {
+    x.error == "",
+    isInRange(x)
+  }
```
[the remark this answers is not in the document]
Kukovec: Yes, someone could write a `UInt16.MAX` value into a `UInt8` manually, and you'd have no way of knowing until the first OOB.

examples/README.md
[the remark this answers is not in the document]
bugarela: Hmm, weird, because if this was wrong, the CI check should complain about a diff. I'll take a look.

examples/spells/BoundedUInt.qnt
bugarela: I thought the same thing. Seems to me that we'd want all `saturating` and `wrapping` operators to be defined over `int`.

examples/spells/BoundedUInt.qnt
```diff
@@ -0,0 +1,607 @@
...
+    if (xx.isValid() and yy.isValid())
+      if (res.isValid()) res.v == expectedVal(xx.v, yy.v)
+      else res.error == expectedErr
+    else res.error == "out of range"
+  }
+
+  // Checked add
+  pure def CAddInv(x: int, y: int): bool = MkCInv(x,y, checkedAdd, "overflow", (xx, yy) => xx + yy)
+
+  pure val CAddInvsTest = and {
```
bugarela, line 302: I think these tests might be too generalized, in the sense that `CAddInv` (and its equivalents for other operators) re-define a lot of the operation from the real operators. I think if we actually have explicit expectations for each scenario, the tests would be more valuable both as reference material and as enforcers of behavior:

```bluespec
pure val CAddInvsTest = and {
  checkedAdd(UInt(0), UInt(0)).v == 0,
  checkedAdd(UInt(1), UInt(0)).v == 1,
  ...,
  checkedAdd(UInt(1), UInt(MAX)).error == "overflow",
  ...
}
```

examples/spells/BoundedUInt.qnt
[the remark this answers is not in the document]
Kukovec: I can rename it `UintT` (`UInt` clashes with the constructor). As for the error field, I prefer keeping it. You can always call `i.isValid`, and get a boolean evaluation, whereas the error string is more descriptive (and already implemented)

examples/spells/BoundedUInt.qnt
```diff
@@ -0,0 +1,607 @@
...
+  /// Unsafe checked integer addition. 
+  /// Computes `lhs + rhs`, setting the error field to "overflow" if overflow occurred.
+  /// Assumes lhs and rhs are valid (w.r.t. `isValid`).
+  pure def checkedAddUnsafe(lhs: T, rhs: T): T = {
+    val res = lhs.v + rhs.v
+    { 
+      v: res,
+      error: if (res > MAX) "overflow" else ""
+    }
+  }
```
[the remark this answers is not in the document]
Kukovec: All of the "unsafe" methods are named as such, because they perform no validation on their inputs, i.e. they do not check whether `lhs` and `rhs` are valid representations of `UIntX`.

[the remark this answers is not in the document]
Kukovec: For example,
```
>> val a = UInt(-10) // {v: -10, error: "out of range"}
>> val b = UInt(0)
>> checkedAddUnsafe(a,b)
{ v: -10, error: "" } // <-- It is not the case that (-10) > MAX, so error = ""
```

examples/spells/BoundedUInt.qnt
Kukovec: It's an interface question. You want these operations to all be `(UIntT, UIntT) => UIntT` so you can chain them. 
If this returned an int, you couldn't do e.g. `saturatingAdd(saturatingAdd(a,b), c)`

examples/spells/BoundedUInt.qnt
Kukovec: - (M)a(k)e (C)hecked (Inv)ariant.
- Yeah, we wrote that name in the draft originally, then it just never got changed, I'll fix it.

examples/spells/BoundedUInt.qnt
Kukovec: The tests, as they are now, literally check for everything in the above example, except with less duplication, _and_ with error enforcement.
For example, `checkedAdd(UInt(0), UInt(0)).v == 0` would return true on `{v: 0, error: "some error"}`

examples/spells/BoundedUInt.qnt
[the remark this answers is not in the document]
shonfeder: The constructor could be `intToUInt`. I think this is how we are making converters of, e.g., maps to sets: https://uxdesign.cc/designing-for-delight-is-dead-677bca6aebd1

examples/spells/BoundedUInt.qnt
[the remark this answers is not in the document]
shonfeder: How about getting rid of the unsafe/safe operators, and just call `isValid` at the end of every operator application? Then everything is safe, we can cut the code in about half, and we need to perform fewer checks. WDYT?

examples/spells/BoundedUInt.qnt
shonfeder: We are thinking the signature should be `(int, int) => int`, so that would chain fine.

examples/spells/BoundedUInt.qnt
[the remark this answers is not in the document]
Kukovec: There's a few problems:
 1) This basically forces us to go with an `error: bool` approach
 2) This misses cases, such as
```
>> val a = UInt(-10) // {v: -10, error: "out of range"}
>> val b = UInt(10)
>> checkedAddUnsafe(a,b)
{ v: 0, error: "" } // bad input, but magically fixed by math
```

examples/spells/BoundedUInt.qnt
shonfeder: I had the exact some thought as Gabriela here, but was too lazy to mention it :D 

I find these tests are less clear than what Gabiela suggests, and the duplication of the logic makes them less good as tests, imo. In my POV, what Gabiela suggests would provide better checks and better explanatory value.

Code reuse in tests is only helpful if it makes the tests clearer, imo, but with the current set up there is no way for a reader to know which cases are supposed to result in which outcomes except by executing the test helper in their head.

> For example, checkedAdd(UInt(0), UInt(0)).v == 0 would return true on {v: 0, error: "some error"}

The test for this should be `checkedAdd(UInt(0), UInt(0)).error == "the error", right?

examples/spells/BoundedUInt.qnt
Kukovec: For `saturating` alone, yes, but you couldn't combine e.g. `saturatingAdd(checkedAdd(a,b), c)` (which was supposed to be the example above, but I borked it in copypaste)

examples/spells/BoundedUInt.qnt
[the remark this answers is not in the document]
shonfeder: Miss pasted! Sorry. That should be https://github.com/informalsystems/quint/blob/main/quint/src/builtin.qnt#L310

examples/spells/BoundedUInt.qnt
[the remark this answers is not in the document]
shonfeder: Ah I see.  Math is magic :D

examples/spells/BoundedUInt.qnt
bugarela: That makes sense, but is it something people would use? I can't really imagine a scenario where someone would perform a checked operation and then wrapping/saturation on top of that.

Kukovec: The underlying issue is, that Rust uses e.g. `u64` and `Option<u64>` in these ops:
```rust
pub const fn checked_add(self, rhs: u64) -> Option<u64>
pub const fn saturating_add(self, rhs: u64) -> u64
```
which don't quite translate 1-to-1 to our `int` and `T = {v: int, error: str}` types. Specifically, operations which take `u64` as arguments never have to worry about _invalid_ inputs. In this sense, `T` is more than just an option type, since it also plays the role of enforcing (or at least having an interface which allows for the checking of) boundary constraints.
This is not an issue for the wrapping and saturating operations,  _however_, it does mean that `checked` needs to have its arguments either be `T`-typed, or perform input validation on `int` values 
(see the thread under https://github.com/informalsystems/quint/pull/1230#discussion_r1376820888 for examples).

In the rust interface, `x.checkedAdd(y.saturatingAdd(z))` is possible, so if we want it to also be possible here, we need to either have:
```
saturatingAdd: (T,T) => T
checkedAdd: (T,T) => T
```
where `T` plays the role of a hybrid option-and-no-exception-thrown type, or
```
saturatingAdd: (int, int) => int
checkedAdd: (int, int) => T
```
where `T` is purely an option type, but we have to add `int`-validation to every method, since `saturatingAdd(-1,-1)` could never be called in rust, as `-1` is not a `u64` value.

Kukovec: As I understand it, you two seem to prefer the 
```
saturatingAdd: (int, int) => int
checkedAdd: (int, int) => T
```
scenario, so I'll go ahead and change it to behave that way

Kukovec: Although, now that I think about it, I'm not sure I know how to combine input validation and an `int` return type. In `saturatingAdd(-1,-1)` I don't see how you'd alert the user that `-1` is an invalid input. Nor does it make the `intToUInt` constructor clear or sensible.

shonfeder: If the aim of this module is to follow the Rust lib as closely as possible, and they also the same underlying type for saturating and checked values, that's a good reason to keep it, IMO. But please document that motivation, including a link to a motivating rust docs or code.

> This is not an issue for the wrapping and saturating operations, however, it does mean that checked needs to have its arguments either be T-typed, or perform input validation on int values

IMO, this should maybe motivate us to consider introducing type abstraction. This would allow us to prevent construction of values of type `T` by means other than the constructors provided in this module and would solve the needless validation problem.

Kukovec: Hmm, lets maybe wait for @konnov and @thpani  to get back and weigh in on this.

examples/spells/BoundedUInt.qnt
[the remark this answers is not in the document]
konnov: > I'm wondering if actual use case needs an `error` message string. Would it be possible that having `error: bool` would be more ergonomic and just knowing that there is an error one way or the other would suffice?

The `error` field is similar to what was done with the modelling of Decimals in https://github.com/informalsystems/quint-sandbox/blob/main/decimal/decimal.qnt. While a Boolean would suffice, having a string is almost as cheap (from the analysis p.o.v.) as having a Boolean, and it it produces better counterexamples.

examples/spells/BoundedUInt.qnt
thpani: I prefer easy operator chaining over compatibility with the Rust library.
Our goal here should be brief, self-explanatory specs.

examples/spells/BoundedUInt.qnt
thpani: +1 for writing these out in full

examples/spells/BoundedUInt.qnt
konnov: I agree with @bugarela here. It would be easier to understand non-generalized tests. If it takes longer to understand a test than to understand the implementation, then our tests are not that useful.

examples/spells/BoundedUInt.qnt
konnov: yeah, chaining would be good IMO

konnov (commented):
left a few comments

examples/spells/BoundedUInt.qnt
Kukovec: Just to make sure I understand the above, you also prefer
```
saturatingAdd: (int, int) => int
checkedAdd: (int, int) => T
```
?
We can do it like that, but I'll just reiterate https://github.com/informalsystems/quint/pull/1230#discussion_r1380019675, doing this allows for "nonsensical" (in the unsigned-int universe) constructions.

Kukovec: Both ways are "chainable"

examples/spells/BoundedUInt.qnt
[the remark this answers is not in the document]
shonfeder: I'm fine with leaving it as an error message if that has proven useful.

I think the remaining item left for this comment is then the type name.

examples/spells/BoundedUInt.qnt
shonfeder: > Our goal here should be brief, self-explanatory specs

I agree with that goal! But its worth noting that requiring the `saturatinAdd` arguments and return value to be wrapped introduces its own complexity, and leaving them as type `int` makes it clearer when interacting with other int operators, like say the predicates or indexing operations.

Wrapping everything in the `UInt` type, whether or not it is needed, will make it easier to chain these operators and that is clearer, but at the expense of every other point where the returned or supplied values have to be cast to/from an int.

An additional benefit of expecting this to use `Option` combinators for chaining (once sum types are in, but using a makeshift maybe in the meantime) is that it makes it self-explanatory to readers that one is chaining partial computations.

Some example use cases using the two different approaches may be illustrative to track the tradeoffs. Do we have any code in mind that would be a good test case?

examples/spells/BoundedUInt.qnt
shonfeder: This is a nice maxim:

> If it takes longer to understand a test than to understand the implementation, then our tests are not that useful

examples/spells/BoundedUInt.qnt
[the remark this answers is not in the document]
Kukovec: I think 
```
UInt: int => UIntT
```
looks much nicer than
```
intToUInt: int => UInt
```
mainly because, if I have to type `intToUInt` for every integer literal in my spec, that's a lot of line-width bloat.
@konnov @thpani @bugarela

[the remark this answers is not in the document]
shonfeder: True, it looks nicer. If we go with an interface that requires wrapping values, makes sense to keep it short.

examples/spells/BoundedUInt.qnt
thpani: At least from the code where this would be useful that I've seen (Rust in Cosmos, bounded integers in smart contracts), you would firmly stay within the bounded fragment. I've never seen an unwrap to integers.

For this purpose, I believe that a single, initial wrapping of Quint's native `int` into a `UInt`, and then simply remaining in `UInt`, is the most practical.

But I'm happy to consider other examples, or if we have none atm, revise this library when necessary.

bugarela (approved):
I love the new tests! Thank you for all the changes :nail_care:

examples/spells/BoundedUInt.qnt
shonfeder: Makes sense.

shonfeder (approved):
Thanks for making the tests clearer :)

PR #1232, opened 2023-10-27
Add type and effect checking for `match` expressions

### LINKED ISSUE #244: Support discriminated unions in the type system
We need to implement variants and use them to type discriminated unions and their operators (`match`) in the type system.

- [x] Implement type checking and inference for variant construction via the exotic `variant` operator introduced in #1206  
- [ ] Implement type checking and inference for variant elimination via the exotic `match` operator introduced in #1206

shonfeder:
Closes #244

This PR adds support for match expressions (sum-type eliminators) to the type checker and effect checker.

The name of the exotic match operator we "desugar" `match expr { ... }` forms into is now `matchVariant` instead of `match`. This is to avoid the collision we otherwise hit with the `match` keyword.


shonfeder:
The effect inference that is being produced for the expression isn't making much sense to me, nor the unification that's being attempted. I'm hoping @bugarela may be able to provide some pointers, as I have been stuck on this for some time now. I'll plan to touch base on this on Monday when Gabriela is back.

quint/src/effects/builtinSignatures.ts
```diff
@@ -241,12 +241,16 @@ const multipleAritySignatures: [QuintBuiltinOpcode, Signature][] = [
...
+    'matchVariant',
     (arity: number) => {
-      const readVars = times((arity - 1) / 2, i => `r${i}`)
-      const args = readVars.map(r => `Pure, (Pure) => Read[${r}]`)
-      return parseAndQuantify(`(Read[r], ${args.join(', ')}) => Read[${readVars.join(', ')}]`)
+      const eliminatorIdxs = range((arity - 1) / 2)
+      const vars = eliminatorIdxs.map(i => `Read[r${i}] & Update[u${i}]`)
+      const args = vars.map(vs => `Pure, (${vs}) => ${vs}`)
+      const readVars = ['r', ...eliminatorIdxs.map(i => `r${i}`)].join(', ')
+      const updateVars = ['u', ...eliminatorIdxs.map(i => `r${i}`)].join(', ')
```
bugarela, line 252: Hey, I'm not sure what is the problem with the effects you are talking about, but this might be the issue or, at least, part of the issue.
```suggestion
      const updateVars = ['u', ...eliminatorIdxs.map(i => `u${i}`)].join(', ')
```

shonfeder: Thanks for taking a look :) I caught this and changed it last week but hadn't pushed the latest to this branch. But I just had a go of writing up how was stuck (with a clear head) and was able to work everything out. Thanks very much! :duck:

quint/src/ir/quintIr.ts
```diff
@@ -179,7 +179,7 @@ export const builtinOpCodes = [
   'length',
   'map',
   'mapBy',
-  'match',
+  'matchVariant',
```
shonfeder, line 182: Prevents conflict with `match` keyword used in sugared `match foo { ... }` form.

quint/src/ir/quintTypes.ts
```diff
@@ -144,6 +144,19 @@ export type EmptyRow = { kind: 'empty' }
 
 export type Row = ConcreteFixedRow | ConcreteRow | VarRow | EmptyRow
 
+/*
+ * Gives all of a row's field names
+ */
+export function rowFieldNames(r: Row): string[] {
```
shonfeder, line 150: This was a helper I couldn't end up using in this go, but is generally useful IMO.

quint/src/generated/Effect.g4
```diff
@@ -1,12 +1,12 @@
...
+/*,*
  * A grammar of Quint Effects
  *
  * @author: Gabriela Moreira, Informal Systems, 2022
  */
 grammar Effect;
 
 effect:   concrete                                           # concreteEffect
-        | '(' (effect (', ' effect)*)? ')' '=>' effect       # arrowEffect
+        | '(' (effect (',' effect)*)? ')' '=>' effect       # arrowEffect
```
shonfeder, line 9: Allows effect variables to be written without spaces after commas. E.g., we can now write `Read[r0,r1]` as well as `Read[r0, r1]`. Previously, only the latter was allowed.

bugarela (approved):
Looks super clean and easy to follow! Thank you for that :heart:

quint/test/types/inferrer.test.ts
```diff
@@ -159,6 +159,102 @@ describe('inferTypes', () => {
...
+      [18n, '(A(int))'],
+      [24n, 'str'],
+      [26n, 'int'],
+      [19n, 'int'],
+      [20n, 'int'],
+      [21n, 'int'],
+      [25n, '(int) => int'],
+      [27n, 'str'],
+      [29n, '{}'],
+      [22n, 'int'],
```
konnov, line 187: using concrete identifiers in this test looks a bit fragile to me. Also, what is `{}` in this context? Is it the empty record = the unit type, or is it an empty row?

konnov (approved):
@shonfeder, nice to see how your work on sum types is converging into something beautiful :)

quint/src/types/specialConstraints.ts
```diff
@@ -224,3 +224,82 @@ export function variantConstraints(
...
+    // Form a constraint ensuring that the match expression fits the sum-type it is applied to:
+    const matchCaseType = sumType(fields) // The sum-type implied by the match elimination cases.
+    const variantTypeIsMatchCaseType: Constraint = { kind: 'eq', types: [variantType, matchCaseType], sourceId: id }
+
+    // Form a set of constraints ensuring that all the result types of the match cases are the same:
+    // We extract just the types of the eliminator operators, which we've ensured are operators during field validation.
+    const eliminatorOpResultTypes = labelAndElimPairs.map(([_, elim]) => (elim[1] as QuintOperType).res)
+    const resultTypesAreEqual: Constraint[] = []
+    for (const resultType of eliminatorOpResultTypes) {
+      resultTypesAreEqual.push({ kind: 'eq', types: [resultTypeVar, resultType], sourceId: id })
```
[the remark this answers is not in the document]
shonfeder: I take this to be implicit in the rule and just an implementation detail necessitated by our architecture. But I think adding a comment is a good to note that is probably a good idea. Thanks for catching that.

quint/src/types/specialConstraints.ts
```diff
@@ -224,3 +224,82 @@ export function variantConstraints(
...
+    }
+  )
+
+  // TODO: Support more expressive and informative type errors.
+  //       This will require tying into the constraint checking system.
+  //       See https://github.com/informalsystems/quint/issues/1231
+  return mergeInMany(validatedFields).map((fields: [string, QuintType][]): Constraint[] => {
+    // Form a constraint ensuring that the match expression fits the sum-type it is applied to:
+    const matchCaseType = sumType(fields) // The sum-type implied by the match elimination cases.
+    const variantTypeIsMatchCaseType: Constraint = { kind: 'eq', types: [variantType, matchCaseType], sourceId: id }
```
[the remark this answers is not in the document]
shonfeder: Yep: that's right! Good idea :)

quint/test/types/inferrer.test.ts
bugarela: It's fragile but important, until we find a better way to do this. This is also implicitly a test for the hover functionality in LSP, ensuring with high granularity that all types are as expected. It's pretty annoying when something about ids change, but I have had errors flagged by this type of tests a lot of times already and it's often helpful (i.e. if we expected `(int) => int` and got `(_t123) => int`, probably some substitution was not properly applied).

I'd like to keep this.

bugarela: Also, for @konnov's question, types and rows are completely separate things in quint, and these values are being parsed as types, so there is no way this could be a row. I'm just trying to help disambiguate things.

quint/src/types/specialConstraints.ts
```diff
@@ -224,3 +224,82 @@ export function variantConstraints(
...
+
+// The rule for the `match` operator:
+//
+// Γ ⊢ e:(s, c)
+// Γ, x1:(t1, c1') ⊢ e1:(t, c1) ... Γ, xn:(tn, cn') ⊢ en:(t, cn)
+// v is fresh
+// ------------------------------------------------------------------------------
+//  Γ ⊢ match e { i1 : x1 => e1, ..., in : xn => en } : (t, c /\
+//    c1 /\ c1' /\ ... /\ cn /\ cn' /\ cn+1 /\
+//    s ~ < i1 : t1, ..., in : tn > )
```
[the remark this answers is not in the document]
shonfeder: Thanks! I those are dangling from previous updates.

shonfeder:
Thank you for the careful reviews :)

quint/src/types/specialConstraints.ts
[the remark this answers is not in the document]
bugarela: I wouldn't say it is necessitated by our architecture/implementation. It's necessitated in a constraint-based type system, so it's in the type system level more than at implementation only. However, it is only a minor detail.

romac:
🎉 🚀 🥇

PR #1234, opened 2023-11-03
Fix broken link

p-offtermatt:
The link, as is, does not work for me and brings me to a 404. This patch fixes it for me.


p-offtermatt:
Maybe you could i.e. add a markdown link checker to the CI, see https://github.com/cosmos/cosmos-sdk/blob/main/.github/workflows/md-link-checker.yml to avoid problems like this? At least my understanding is this should be very consistently found by something like this

shonfeder (approved):

shonfeder:
Thanks for the fix and the suggestion :)

PR #1237, opened 2023-11-07
Refactoring parser tests

konnov:
Towards #17. I have carved out an important piece of refactoring from #1221: Instead of testing error messages via the JSON output, test them directly in unit tests. This significantly improves our parser tests:

 - It is clear what code is being parsed. One does not have to look into a separate file.
 - The error message is clearly specified. Again, we do not have to look into a JSON file (nobody does that).
 - It's harder to ignore the failing tests, that is, there is no automatic way of agreeing with failing tests (as is happening with our current script).
 - We are testing the error messages, instead of testing the JSON format. There should be but a few tests that actually test the JSON output.

I have only refactored the tests of phase 1. We should understand whether we want to refactor the remaining phases the same way.


konnov:
@bugarela, this test fails in a weird way. I guess, it is the result of recent refactoring. What shall we do about it?

```
     Error: Value is None
      at errorFactory (node_modules/@sweet-monads/maybe/cjs/index.js:105:33)
      at MaybeConstructor.unwrap (node_modules/@sweet-monads/maybe/cjs/index.js:108:15)
      at ToIrListener.exitConst (src/parsing/ToIrListener.ts:118:36)
      at ConstContext.exitRule (src/generated/QuintParser.ts:4074:13)
      at ParseTreeWalker.exitRule (src/tree/ParseTreeWalker.ts:105:7)
      at ParseTreeWalker.walk (src/tree/ParseTreeWalker.ts:57:11)
      at parsePhase1fromText (src/parsing/quintParserFrontend.ts:120:27)
      at parseErrorsFrom (test/parsing/quintParserFrontend.test.ts:36:29)
      at Context.<anonymous> (test/parsing/quintParserFrontend.test.ts:150:20)
      at processImmediate (node:internal/timers:478:21)
```

bugarela:
> @bugarela, this test fails in a weird way. I guess, it is the result of recent refactoring. What shall we do about it?

This is the same problem I've been mentioning, that should be fixed by https://github.com/informalsystems/quint/pull/1228. I'm working on the debug error messages for that and then we can merge it (I believe), thus unblocking this. Alternatively, we can comment out the invalid `const` test and see if that's enough to make the error go away for now, and then I uncomment that test on my PR.

bugarela:
@konnov I merged my PR and updated this branch, and the error is gone. Now there is only a mismatch in the error message.

konnov:
> @konnov I merged my PR and updated this branch, and the error is gone. Now there is only a mismatch in the error message.

thanks @bugarela! I have fixed the remaining unit test. Could you have a look. Then I would be able to faster iterate on #1227.

quint/test/parsing/quintParserFrontend.test.ts
```diff
@@ -113,31 +120,104 @@ describe('parsing', () => {
...
 
-describe('parse errors', () => {
-  it('error in module unit', () => {
-    parseAndCompare('_1002emptyWithError')
+// instead of testing how errors are produced in json,
+// we only test the error messages
+describe('syntax errors', () => {
+  it('unbalanced module definition', () => {
+    const code = 'module empty {'
+    const errors = parseErrorsFrom('module-unbalanced', code)
```
bugarela, line 128: Can we make this source name parameter fixed so a value? I don't see why we need to name the source name for every test differently

quint/test/parsing/quintParserFrontend.test.ts
```diff
@@ -113,31 +120,104 @@ describe('parsing', () => {
...
+    assert.isAtLeast(errors.length, 1)
+    assert.equal(errors[0].message, `mismatched input ''' expecting STRING`)
+    assert.equal(errors[0].code, 'QNT000')
+  })
+})
+
+// Test the JSON error output.  Most of the tests here should migrate to the
+// above test suite called 'syntax errors' or other test suites for the later
+// phases.  It is sufficient to have a few tests that compare JSON. For the
+// rest, we are interested in checking the error messages, not the JSON files.
```
bugarela, line 219:
```suggestion
// Test the JSON error output. Most of the tests here should migrate to the
// above test suite called 'syntax errors' or other test suites for the later
// phases. It is sufficient to have a few tests that compare JSON. For the
// rest, we are interested in checking the error messages, not the JSON files.
```

bugarela (approved):
This is such a great improvement on readability and maintainability, thanks! It does make the tests weaker, and I'm specially concerned about testing if we are highlighting the correct portion of the code. Perhaps we can use the `reportError` facility and then compare the text with `^^` symbols, instead of looking at the locs. 

Anyway, I think this is much better than what we had before and we can iterate on further improvements later!

PR #1224, opened 2023-10-19
Fix problems reporting errors between multiple files

bugarela:
- [ ] Tests added for any new code
- [ ] Documentation added for any new functionality
- [ ] Entries added to the respective `CHANGELOG.md` for any new functionality
- [ ] Feature table on [`README.md`](../README.md#roadmap) updated for any listed functionality

PR #1240, opened 2023-11-08
Release v0.15.0

bugarela:
### Changed

- Error messages for `val` vs `def` and `pure val` vs `pure def` errors are clearer (#1208)
- `quint run` prints the random seed even if no bug was found (#1213)
- Error reporting was changed to show more errors at a time, instead of having a lot of phases (#1220)

### Fixed

- Fixed internal bugs in the effect checker that could cause an incorrect effect
  to be inferred or error to be reported (#1203)
- Fixed propagation of `checker.tuning` Apalache config file key for `quint
  verify` (#1216)
- Fixed a problem where errors in one file were being reported in another file
  that imported it (#1224).
- Fixed a problem where some errors were not being reported in the REPL (#1223)


bugarela:
Ah, forgot to update with main before running the script, but this only missing updates in tests, examples and spells, so it's fine to release without the latest changes.

shonfeder (approved):

PR #1242, opened 2023-11-08
Add support for sum types in the simulator

shonfeder:
Closes #1033

Adds support for sum types to the compiler (and so the simulator):

- Generalizes the REPL so it can compile all the declarations parsed out of its
  input, instead of just the last one (required to enable sum type declarations
  to be entered in the REPL).
- Adds graphics for pretty printing variants.
- Adds evaluation of constructors, requiring a new runtime value to represent
  variants.
- Adds evaluation of match expressions, which eliminate variants.

The history is clean enough that it would probably help to review by commit.

- [x] Tests added for any new code
- [ ] Documentation added for any new functionality
- [ ] Entries added to the respective `CHANGELOG.md` for any new functionality
- [ ] Feature table on [`README.md`](../README.md#roadmap) updated for any listed functionality

PR #1243, opened 2023-11-10
Fix handling of wildcard match cases

shonfeder:
This is a followup to #1242 and #1232.

While the preceding gave basic support for sum type constructors and
eliminators, we did not have support for properly typing "wildcard" or default
match cases.

This PR simplifies the representing and logic around such cases and extends the
constraint generator to give the correct typing for match expressions that
include a default case.

The gist:

A match expression of the form

``` quint
match foo {
  | A(_:t) => bar
  | B(_:u) => baz
  | _      => qux
}
```

Is now typed to apply to a variant `<A : t, B : u | v >` where `v` is an free
row variable. The open row accounts for the fact that the wildcard handles the
rest of the labels, whatever they may be.

The constraint generator is also expanded with error handling to catch cases
where redundant wildcard's or match statements are included.


shonfeder:
After this is in, I'll write a more extensive series of tests using the constructs, along with documentation and release notes announcing the feature.

bugarela (approved):
LGTM

PR #1245, opened 2023-11-13
Remove Deprecated Union Types

shonfeder:
This PR removes support, documentation, tests, and examples using the deprecated union types.

Almost all of this change is either from generated code (from antlr) or mechanical removal of code for handling these types.

The only notable exceptions are

- 6ba5791 which updates Paxos.qnt to use sum types
- ba9c25a which updates lang.md, to document sum types and their use

A forthcoming PR will expand on the documentation in other needed places.


examples/classic/distributed/Paxos/Paxos.qnt
```diff
@@ -142,14 +167,13 @@ module Paxos {
   (* greater than b (thereby promising not to vote in ballot b).             *)
   (***************************************************************************/
   action Phase2a(b, v) = all {
-    not(msgs.filter(m => m.tag == "2a")
-            .exists(m => m.bal == b)),
+    not(msgs.exists(m => match m { Msg2a(m2a) => m2a.bal == b | _ => false })),
     Quorum.exists(Q =>
-      val Q1b = msgs.filter(m => m.tag == "1b")
-                    .filter(m => (m.acc.in(Q)) and (m.bal == b))
```
shonfeder, line 149: Note that this seems to have been an error, because messages tagged "1b" do not have an `acc` field. This may be good evidence of the advantage of the stricter typing discipline of the new implementation of sum-types.

examples/classic/distributed/Paxos/Paxos.qnt
```diff
@@ -34,15 +34,41 @@ module Paxos {
...
+      | Msg1b(m1b) => acc.union(Set(m1b))
+      | _          => acc
+    })
+
+  // Filter a set of messager into just 2a messages
+  def filterMsg2a(ms: Set[MSG]): Set[Msg2aT] =
+    ms.fold(Set(), (acc, m) => match m {
+      | Msg2a(m2a) => acc.union(Set(m2a))
+      | _          => acc
+    })
```
konnov, line 70: `fold` seems to be an overkill for me. Why would not we just use `filter`?

konnov (approved):
Amazing :)

examples/classic/distributed/Paxos/Paxos.qnt
shonfeder: We are mapping values out the type `Set[MSG]` into `Set[Msg2aT]`. There is no way of doing this with a filter, which can reduce the elements of the set, but cannot (given our current type system) reduce the type of the elements in the set. If we had instead exposed the extensibility of the extensible variants we are based on -- i.e, by allowing anonymous sum types -- we could give filter a typing that could refine the type of the collection it works on, in this case to just holding `Msg2a` variants. Without mapping out of `Set[MSG]`, whatever consumed the result of the filter would still need to account for all possible message types.

bugarela (approved):
Great! Only some stylistic comments :nail_care: 

Also, in Apalache we have `VariantFilter`, which I don't think we can implement in quint. We should pay attention to how much this filter patterns will emerge on the next specs we introduce sum types to, and then think if we can/do something about it.

examples/classic/distributed/Paxos/Paxos.qnt
shonfeder: I'm not sure what motivated the naming. The names in this spec are mostly cryptic and inscrutable to me :D I suspect it would be a good idea to do a pass to give less cryptic names at some point, but I'll not tackle name fixes here.

shonfeder:
> in Apalache we have VariantFilter, which I don't think we can implement in quint

I think we could do so, it would just require exposing the extensibilty of the underlying extensible variants. This may be as simple as removing the `isDefined` check from our constraints and adding an equivalent operator, but it goes against the decision for simplicity reached in the course of discussion around the sum type RFC.  This is discussed in https://github.com/informalsystems/quint/blob/main/doc/rfcs/rfc001-sum-types/rfc001-sum-types.org#why-not-support-polymorphic-variants and the generalization is presented in the immediately preceding section.

PR #1246, opened 2023-11-14
add frequently asked questions

konnov:
Since we keep receiving similar questions about several constructs, I have added a page with explanations. In the future, this material should migrate into something more organized such a book.

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -1,0 +1,106 @@
+# Frequently Asked Questions
+
+Here you can find answers to the questions that are often asked by people who
+would like to start with Quint.
+
+### What are spells?
+
+Spells are simply Quint modules that contain often-used definitions. There is
+nothing special about these definitions. They are probably a bit more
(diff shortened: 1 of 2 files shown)


doc/faq.md
```diff
@@ -0,0 +1,106 @@
...
+
+Here you can find answers to the questions that are often asked by people who
+would like to start with Quint.
+
+### What are spells?
+
+Spells are simply Quint modules that contain often-used definitions. There is
+nothing special about these definitions. They are probably a bit more
+well-thought than a definition someone would write on the spot. Check the page
+on [Spells][].
```
thpani, line 11:
```suggestion
nothing special about these definitions. They are probably a bit more
thought-out than a definition someone would write on the spot. Check the page
on [Spells][].
```

thpani (approved):
looks good!

I think the confusion around `pure val` and `const` could be avoided by renaming `const` to `param` (and maybe `pure val` to `const` or `const val`).

PR #1249, opened 2023-11-16
fix the antlr grammar

### LINKED ISSUE #1247: Fix the grammar: an operator definition should use `qualId`
Instead of `normalCallName`, the syntax rule for the operator definition should refer to `qualId`. It should not be possible to use the builtin operators in new definitions.

https://github.com/informalsystems/quint/blob/15596529a44749255219d211ef5e4960d09b198a/quint/src/generated/Quint.g4#L44

konnov:
Closes #1247. The rule for operator definition should accept `qualId`, not `normalCallName`.


quint/src/generated/.antlr/QuintLexer.java
```diff
@@ -1,4 +1,4 @@
-// Generated from /Users/thomas/src/quint/quint/src/generated/Quint.g4 by ANTLR 4.13.1
+// Generated from /Users/igor/devl/informal/quint/quint/src/generated/Quint.g4 by ANTLR 4.13.1
```
konnov, line 1: oh, that's weird. Why does my HOME propagate in the source code :)

konnov:
Oh wow. The tests are failing. Maybe this explains why we had this grammar rule this way. However, it is weird that we have it.

quint/src/generated/.antlr/QuintLexer.java
bugarela: Weird!

bugarela:
Now that I see the behavior, I think we should actually change all usages of `qualId` to `normalCallName` (and perhaps rename those accordingly). I just tried writing `module iff { }` and it gives me a parsing error about a missing identifier. 

I think if a user tries to define something with builtin in operator name, we should consistently report a name conflict, not a parser error.

konnov:
> Now that I see the behavior, I think we should actually change all usages of `qualId` to `normalCallName` (and perhaps rename those accordingly). I just tried writing `module iff { }` and it gives me a parsing error about a missing identifier.
> 
> I think if a user tries to define something with builtin in operator name, we should consistently report a name conflict, not a parser error.

I am not sure I understand your reasoning here. Why would be expect `module iff { ... }` to be a syntactically correct code? Parsers typically reject things that do not fit into the language, e.g., `module + { }` would not work either.

bugarela:
> I am not sure I understand your reasoning here. Why would be expect `module iff { ... }` to be a syntactically correct code? Parsers typically reject things that do not fit into the language, e.g., `module + { }` would not work either.

I don't really care if it is syntatically correct or not. What I'm saying is that this message:
```
/home/gabriela/projects/quint/local/test.qnt:2:3 - error: [QNT101] Built-in name 'iff' is redefined in module 'A'
2:   val iff = 1
     ^^^^^^^^^^^

error: parsing failed
```

is better than this:
```
/home/gabriela/projects/quint/local/test.qnt:1:8 - error: [QNT000] missing IDENTIFIER at 'iff'
1: module iff {
          ^^^

error: parsing failed
```

because it gives me more information. If I have no idea that `iff` is a built in name, it would be quite hard to understand the problem in the second error message.

konnov:
> If I have no idea that iff is a built in name, it would be quite hard to understand the problem in the second error message.

Yeah, better error names would definitely help. However, many programming languages have a predefined set of keywords, which cannot be used as identifiers. I don't think that any compiler would give a super-friendly error message on something like `class + object == if`.

konnov:
For instance, if I feed the following to the rust compiler:

```rust
struct + fn == if
```

It outputs the following error message:

```
error: expected identifier, found `+`
 --> t.rs:1:8
  |
1 | struct + fn == if
  |        ^ expected identifier

error: aborting due to previous error
```

It's quite cryptic, but that is how parsers are working.

konnov:
> Yes, because it is hard/impossible to do that due to grammar conflicts and precedence, right? My point is, if we can easily do it (have the proper error message), we should do it. The change on this PR goes against this direction, so unless there is a good reason for that, I wouldn't do it - not only for making this a syntactic error

It actually works in antlr4, to my surprise. When I tried the same with PEGs (recursive descend), the parser was breaking very badly. We are kind of lucky that antlr4 is so power, even if it produces unreadable error messages.

shonfeder:
Closing in favor of https://github.com/informalsystems/quint/pull/1325

PR #1251, opened 2023-11-17
ignore .DS_Store

konnov:
Just a one-liner

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,4 @@ client/server
 .yalc
 *.vsix
 _build
-**/.DS_Store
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ client/server
 .yalc
 *.vsix
 _build
+**/.DS_Store
```
thpani, line 9: Does this match a top-level `.DS_Store`?

This should be enough
```suggestion
.DS_Store
```

thpani (approved):

Issue #1244, opened 2023-11-10
Document sum types


shonfeder:
Add documentation and release notes about sum types syntax etc.

Issue #1254, opened 2023-11-18
Update cheatsheet with syntax for sum types


shonfeder:
This should be part of #1244, but afaict the process for updating the cheatsheet is not documented and only @konnov knows how this is done, thus this request :)

The info needed should be in `lang.md` or `builtin.qnt`.

Thanks!

PR #1255, opened 2023-11-18
Document sum types

shonfeder:
I added an update everywhere I could find that made sense to me and I knew how
to update. The outstanding work on this
https://github.com/informalsystems/quint/issues/1254.

Let me know if you can think of anything else that should be added to document
the feature at this point.

### DIFF
--- a/quint/src/builtin.qnt
+++ b/quint/src/builtin.qnt
@@ -964,8 +964,52 @@ module builtin {
   // /// pure val s = tuples(Set(1, 2), Set(3, 4))
   // /// assert(s == Set((1, 3), (1, 4), (2, 3), (2, 4)))
   // /// ```
-
   // pure def tuples(a*): (Set[a0], ..., Set[an]) => Set[(a0, ..., an)]
+
+  // /// `variant("Label", expr)` is a value of type `T` when
+  // ///
+  // /// - `T` is a predifined sum type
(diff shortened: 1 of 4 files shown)


quint/src/builtin.qnt
```diff
@@ -964,8 +964,52 @@ module builtin {
...
+  // /// assert(variant("A", 42) == A(42))
+  // /// assert(variant("B", Rec()) == B)
+  // /// assert(variant("C", "foo") == C("foo"))
+  // /// ```
+  // ///
+  // /// Note that this is an internal built-in and users should prefer the generated
+  // /// data constructor matching the label of a variant. E.g., in our example above,
+  // /// users should prefer the equivalent form on the right hand side of the `==`.
+  // def variant(label, x): (str, E) => T
+
```
konnov, line 988: This is an idea for the future. Shall we add a prefix to the names of the internal operators, e.g., `__variant` or `q::variant`. This would send the users a clear message that these operators should be used with care.

konnov (approved):
LGTM! I will take care of #1254

quint/src/builtin.qnt
shonfeder: Not a bad idea, but it will make the normalized pretty printing harder to read I think unless we have custom printers for all the hidden operators?

shonfeder:
This closed #1244

PR #1256, opened 2023-11-20
update the cheatsheet with the sum types

### LINKED ISSUE #1254: Update cheatsheet with syntax for sum types

konnov:
Closes #1254. This is simply an update in the cheatsheet as requested in #1254.


shonfeder:
Two suggestions:

1. The example of applying a constructor uses `Req`, which is not declared:

image

2. I suggest putting sum types right after records, because they are dual.

shonfeder (approved):
Thanks!

PR #1258, opened 2023-11-20
Release v0.16.0

shonfeder:
### Added

- Support for sum types in type checking and simulation (#244).

### Removed

- The long deprecated union types have been removed, in favor of the new sum
  types (#1245).


konnov (approved):
Let's go :)

PR #1260, opened 2023-11-21
Fix the default module in `quint run` and `quint test`

### LINKED ISSUE #1195: Crashing when no main module is given
I am getting the same error as https://github.com/informalsystems/quint/issues/1192, but I think for a different reason, namely for an omitted main module.
The spec is here: https://github.com/cosmos/interchain-security/blob/c291303159ce3c28cf00484e34252e07fd94bffc/tests/difference/core/quint_model/ccv_statemachine.qnt#L164

and I am running 

```
quint run --invariant ValidatorUpdatesArePropagated ccv_statemachine.qnt
```

This crashes:

```
TypeError: Cannot read properties of undefined (reading 'id')
    at runSimulator (/Users/offtermatt/projects/quint/quint/dist/src/cliCommands.js:376:63)
    at EitherConstructor.asyncChain (/Users/offtermatt/projects/quint/quint/node_modules/@sweet-monads/either/cjs/index.js:125:16)
    at /Users/offtermatt/projects/quint/quint/dist/src/cli.js:33:39
```

When I run it with a main module, it does not crash:

```
quint run --invariant ValidatorUpdatesArePropagated ccv_statemachine.qnt --main CCVDefaultStateMachine
```

I think there should be a dedicated error message for this case

### LINKED ISSUE #1219: Minor improvement request: Automatically detecting the main module even when it's not the same as the file name
I have a file that contains a single module, e.g. a file called spec.qnt:
```
module Alpha {
...
}
```

When I try to run tests, e.g. 
```
quint test spec.qnt 
```
then Quint complains that it cannot find the main module, because it assumes the main module
has the same name as the file:
```
  spec
error: Cannot find main module
error: Tests failed
```

I'd argue that if there is a file with only a single module in it and I run quint against it, it should just take that one module as the main module. Of course, maybe the current way helps encourage the desired practice of naming the main module the same as the file, but I'm not sure there's a reason for this.

Maybe I am overlooking something here, but my preferred behaviour would be: if the file contains a single module, just use that as the main module. If it contains multiple modules, use the module with the same name as the file name if one exists. If not, complain.

konnov:
Closes #1195. Closes #1219. Both issues were about the behavior of `quint run` and `quint test` when no module was specified. Now the main module is detected as follows:

 - when the flag `--main <name>` is passed, `<name>` is used as the main module name
 - otherwise, if there is only one module in the source file (not considering the imports), then this module's name is used as the main module name
 - otherwise, the main module name is extracted from the filename.

Also, if the main module is not found, the commands report an error instead of failing with a stack trace.


CHANGELOG.md
```diff
@@ -8,10 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 
 ### Added
+
+- Use the module's name when the input has only one module (#1260)
```
shonfeder, line 12: iiuc, something like this would be a clearer statement of the impact of this change:

```suggestion
- When an input file only one module, it will be inferred as the main module  (#1260)
```

shonfeder (approved):

PR #1261, opened 2023-11-22
Add publishing instruction for the language server npm package

bugarela:
Hello :octocat: 

This adds a instruction to "manually" publish the [@informalsystems/quint-language-server](https://www.npmjs.com/package/@informalsystems/quint-language-server) package after publishing the VSCode plugin. We should include this in the script at some point.

P.S. I ran this for `v0.11.0` already, it is all good.


shonfeder:
Thank you for catching this and adding in the docs! Adding it to the script would indeed be a help :)

Does this have be after the vscode plugin? Or only after the quint executable is published?

shonfeder (approved):

PR #1264, opened 2023-11-23
fix: reset trace recorder at beginning of each test run

### LINKED ISSUE #1263: `quint test --output` saves the same trace for all tests
I expect `quint test --output {}.itf.json` to save valid traces for each test. For example,

```bluespec
module manyTests {
  var x: int
  action init = x' = 0

  run firstTest = init.then(x' = 1).then(all { assert(x == 1), x' = x })
  run secondTest = init.then(x' = 2).then(all { assert(x == 2), x' = x })
}
```

should generate ITF traces corresponding to valid runs of `firstTest` and `secondTest`.

But, when using the CLI, the generated traces both have the same state.

```sh
$ quint test --output {}.itf.json manyTests.qnt
  manyTests
    ok firstTest passed 1 test(s)
    ok secondTest passed 1 test(s)

  2 passing (13ms)
$ cat *.itf.json | jq -c .states
[{"#meta":{"index":0},"x":{"#bigint":"0"}},{"#meta":{"index":1},"x":{"#bigint":"1"}}]
[{"#meta":{"index":0},"x":{"#bigint":"0"}},{"#meta":{"index":1},"x":{"#bigint":"1"}}]
```

This trace is not valid for `secondTest`.

The source of this is probably, the `recorder` is not cleared after each `recorder.getBestTrace()` calls.

https://github.com/informalsystems/quint/blob/b361da0a7f999d7c49d716e6ed14cbcf2f64bd62/quint/src/runtime/testing.ts#L85

This was first reported by @hvanz and confirmed by @bugarela, who also provided me the minimal example.

rnbguy:
Closes #1263

Taking the example from the issue,

```bluespec
module manyTests {
  var x: int
  action init = x' = 0

  run firstTest = init.then(x' = 1).then(all { assert(x == 1), x' = x })
  run secondTest = init.then(x' = 2).then(all { assert(x == 2), x' = x })
}
```

After compiling `quint` with the change from this PR, the saved traces correspond to each test case.

```sh
$ quint test --output {}.itf.json manyTests.qnt
  manyTests
    ok firstTest passed 1 test(s)
    ok secondTest passed 1 test(s)

  2 passing (9ms)
$ cat *.itf.json | jq -c .states
[{"#meta":{"index":0},"x":{"#bigint":"0"}},{"#meta":{"index":1},"x":{"#bigint":"1"}}]
[{"#meta":{"index":0},"x":{"#bigint":"0"}},{"#meta":{"index":1},"x":{"#bigint":"2"}}]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Produce proper error messages on invalid module name (#1260)
+- Fix JSON output when running multiple tests (#1264)
 
 ### Security
 
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -1038,3 +1038,20 @@ quint test --main=invalid ./testFixture/_1050diffName.qnt
 error: [QNT405] Main module invalid not found
 error: Tests failed
 ```
+
+### Multiple tests output different json
+
+See [#1264](https://github.com/informalsystems/quint/pull/1264).
+
+<!-- !test in multiple jsons -->
+```
+quint test --output {}.itf.json ./testFixture/_1051manyTests.qnt >/dev/null
+cat firstTest.itf.json secondTest.itf.json | jq -c .states
+rm firstTest.itf.json secondTest.itf.json
+```
+
+<!-- !test out multiple jsons -->
+```
+[{"#meta":{"index":0},"x":{"#bigint":"0"}},{"#meta":{"index":1},"x":{"#bigint":"1"}}]
+[{"#meta":{"index":0},"x":{"#bigint":"0"}},{"#meta":{"index":1},"x":{"#bigint":"2"}}]
+```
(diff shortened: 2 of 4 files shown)


shonfeder:
Thank you for the report and the fix!
This looks great to me: I'll just add a regression test and a fix the lints and we should be good to go.

konnov:
Thanks for the fix. Let me see how it fits in the context.

rnbguy:
@konnov ah. Feel free to make changes in the PR and the branch.

But it works fine for the minimal example and the one (which is more complex) where Hernan first stumbled upon this issue.

shonfeder:
I'll leave this in Igor's hands then. Let me know if you want me to pick up something here!

konnov:
> @konnov ah. Feel free to make changes in the PR and the branch.
> 
> But it works fine for the minimal example and the one (which is more complex) where Hernan first stumbled upon this issue.

Thanks for the fix @rnbguy! I just wanted to make sure I understood why one line was fixing everything. It's a very neat fix :)

I've added a regression test, so we don't hit that issue again.

rnbguy:
Thanks ! Just one thing, this regression will fail again when #1265 is fixed :sweat_smile:

Issue #1266, opened 2023-11-24
Add facility for printing evaluated expressions to the console while executing a spec


romac:
When writing or debugging a specification, it would be useful to me to be able to print expressions to the console as they are evaluated (whether via `test or `run`), much like when writing code in any other language.

Ideally, this `debug` function would return the value it was given so that it can be inserted anywhere within an expression without having to pull the expression to debug out in a statement.

eg. if we have:

```bluespec
val foo = someCall(bar, otherCall(baz))
```

and want to inspect the value of `otherCall(baz)`, we could do

```bluespec
val foo = someCall(bar, debug("result", otherCall(baz))
```

instead of

```bluespec
val res = otherCall(baz)
val _ = debug(res)
val foo = someCall(res)
```

or

```bluespec
val res = otherCall(baz)
val _ = debug("result", otherCall(baz))
val foo = someCall(otherCall(baz))
```

## Questions

- [ ] Output to stdout or stderr?
- [ ] Automatically insert `:` between the description text and the value?

PR #1267, opened 2023-11-24
Add `debug` built-in

romac:
Closes: #1266 

The `debug` built-in is a utility for aiding in debugging specifications by allowing the spec developer to print out a contextual message while executing a specification, whether in a run or in a test, along with a formatted representation of a value.

This can be useful when trying to understand the state of variables at specific points during the spec execution or while tracking down issues.

The `debug` function has the following signature:

```bluespec
def debug(header: str, value: a) -> a
```

The `debug` function prints both header and the value on the same line, preceded by a `>` character, and then returns the given value, which makes it easy to debug nested expressions without having the extract them into a variable first.

Example usage in the REPL:

```bluespec
>>> debug("a number:", 42)
> a number: 42
42

>>> debug("a record:", { foo: 32, bar: true })
> a record: { bar: true, foo: 32 }
{ bar: true, foo: 32 }

>>> debug("large record:", largeRecord)
> large record: {
  hello:
    {
      world:
        { foo: { bar: { baz: 42 }, toto: { tata: { tutu: "Hello, World" } } } }
    }
}
{
  hello:
    {
      world:
        { foo: { bar: { baz: 42 }, toto: { tata: { tutu: "Hello, World" } } } }
    }
}

>>> if (a * debug("nested", a + b) == 10) true else false
> nested 5
true
```

---


bugarela (commented):
I love it! Thanks for getting this done, I'm sure it will be super useful.

I'd do three things before merging:
1. Adding an entry to `CHANGELOG.md`
2. Adding an integration test to `io-cli-tests.md` (i.e. evaluating a quint REPL shell invocation and checking that the debug line is sent to STDOUT)
3. Add an entry to `builtin.qnt` for `debug` with a docstring (and running `npm run api-docs`) to generate a Markdown. This is used for showing overlay documentation in the IDEs.

If you are interested in doing any of those, please go ahead! Otherwise, I can pick it up :smile:

romac:
Thanks a lot for the pointers! I've done all three things, let me know if you want me to add more tests or expand on the documentation.

romac:
I wonder if it would make sense to switch the arguments around, to allow for writing the following in method-call style:

```bluespec
>>> 42.debug("value is")
> value is 42
```

The downside with this order is that when used as a function it now reads less well imho:

```bluespec
>>> debug(42, "value is")
> value is 42
```

I don't feel strongly either way, so will defer to you on that if you don't mind :)

quint/src/builtin.qnt
```diff
@@ -835,6 +835,24 @@ module builtin {
...
+  ///
+  /// ```
+  /// var x: int
+  /// >>> (x' = 0).then(3.reps(i => x' = debug("new x:", x + 1)))
+  /// > new x: 1
+  /// > new x: 2
+  /// > new x: 3
+  /// true
+  /// ```
+  pure def debug(msg, value): (str, a) => a
```
romac, line 854: Can a pure function read from state variables? If not, then we need to remove `pure` here.

quint/src/effects/builtinSignatures.ts
```diff
@@ -223,6 +223,7 @@ const otherOperators = [
   { name: 'reps', effect: parseAndQuantify('(Pure, (Read[r1]) => Read[r2] & Update[u]) => Read[r1, r2] & Update[u]') },
   { name: 'fail', effect: propagateComponents(['read', 'update'])(1) },
   { name: 'assert', effect: propagateComponents(['read'])(1) },
+  { name: 'debug', effect: propagateComponents(['read'])(2) },
```
romac, line 226: If `debug` is indeed pure, do we need to specify that here?

quint/src/builtin.qnt
konnov: a pure function cannot read the state variables. So yeah, it should not be pure.

konnov (commented):
I like this feature! But I have my doubts about the operator name `debug`:

 - It only works for REPL. We add the prefix `q::` to all internal or repl-specific operators.
 - The name `misleading` is a bit misleading. It's not really debugging anything. Quick alternative from `gdb`: `print`, `inspect`, `trace`. Of course, I would prefer them to have the prefix `q::`.

romac:
>  - It only works for REPL. We add the prefix `q::` to all internal or repl-specific operators.

Oh I wasn't aware of that. I'll see if I can figure out how to do that.

>  - The name `debug` is a bit misleading. It's not really debugging anything. Quick alternatives from `gdb`: `print`, `inspect`, `trace`. Of course, I would prefer them to have the prefix `q::`.

Good point! If you ask me I would go with `inspect`:

a) to avoid overloading the "trace" term too much which could be confusing/misleading
b) while `print` might be a good alternative, I think a `print` built-in should allow omitting the value argument altogether and only print the message in that case. Conversely, an argument can probably be made for allowing one to pass only the value without a message to `debug` but I hadn't considered that previously.

konnov:
I also like `inspect` or `q::inspect` :)

romac:
Yeah I meant `q::inspect` :)

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -914,6 +917,15 @@ export class CompilerVisitor implements IRVisitor {
           this.translateAllOrThen(app)
           break
 
+        case 'debug':
+          this.applyFun(app.id, 2, (msg, value) => {
+            let columns = terminalWidth()
+            let valuePretty = format(columns, 0, prettyQuintEx(value.toQuintEx(zerog)))
+            console.log('>', msg.toStr(), valuePretty.toString())
```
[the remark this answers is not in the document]
shonfeder: I'd suggest leaving this as is for now, and replacing it with the proper logging lib along with all our other console prints when we get around to that. This is not a compilation error, so I wouldn't want to put it in the tracker for that.

shonfeder:
Thanks for the useful addition!

We'll need to make sure we strip this out of the IR before sending to alache I think, but we can do that in a followup.

quint/src/builtin.qnt
bugarela: Actually, pure defs can operate on values that read from the state, but it can't add more reads on top of that.

`f` is pure here:
```
var s: int
pure f(x) = x

val foo = f(s)
```

`f` is NOT pure here:
```
var s: int
pure f(x) = x + s

val foo = f(1)
```

So in this case, `debug` can be pure

quint/src/effects/builtinSignatures.ts
bugarela: `propagateComponents(['read'])` is compatible with a pure def, and I believe this is the most appropriate effect. We probably don't want to call `debug` on actions (i.e. `debug("foo", x' = x + 1)`) nor in temporal formulas (i.e. `debug("bar", next(x) == x)`)

bugarela:
I personally like `debug` (or `q::debug`) more than `inspect` or the other alternatives. But I don't feel too strong about it.

quint/src/builtin.qnt
konnov: Ah, I have missed that. You are right @bugarela

konnov:
> How should we proceed here? Should we go with `q::inspect`, or do we want to have further sync/async discussions on this?

I would vote for proceeding like that.

konnov:
@romac do you need help in getting this one merged?

romac:
I was just waiting for the Quint team to decide on the final name, but if/once that's done I am happy to do the renaming if needed and prefix it with the `q::` namespace :)

romac:
I merged `main` in and moved `debug` under the `q::` namespace.

Shall we keep the name or do I rename it to something else?

bugarela (approved):
I talked to the team and we are going with `q::debug`! Thanks for the patience and for all the updates.

romac:
Thanks for helping me see this PR through!

PR #1268, opened 2023-11-24
Toposort on modules

### LINKED ISSUE #1009: Sort modules according to import/export dependencies
Following #1008, we should support out of order modules. For example:

```bluespec
module main {
  pure val myM = sqr(3)
  pure val quadM = 2 * B1::doubleM
  import A.*
  import B(M = myM) as B1
  export B1.*
}

module A {
  pure def sqr(x) = x * x
}

module B {
  const M: int
  pure val doubleM = M + M
}
```

### LINKED ISSUE #1108: Cyclic imports
When I have modules arranged like this, they are fine:

```
module one {
    type Foo = int
}

module two {
    type Boo = int
}

module onePrime {
    import one.*
    import two.*

    var a: Boo
    var b: Foo
}

module twoPrime {
    import one.*
    import two.*

    var c: Boo
    var d: Foo
}
```

However, if I split the modules across two files, one for one and one for two, the imports do not work:

File `moduleOne.qnt`
```
module one {
    type Foo = int
}

module onePrime {
    import one.*
    import two.* from "./moduleTwo"

    var a: Boo
    var b: Foo
}
```
File `moduleTwo.qnt`
```
module two {
    type Boo = int
}


module twoPrime {
    import one.* from "./moduleOne"
    import two.*

    var c: Boo
    var d: Foo
}
```

This gives me a cyclic import error:

`Cyclic imports: '/Users/offtermatt/projects/interchain-security/tests/difference/core/quint_model/test/moduleTwo.qnt' imports '/Users/offtermatt/projects/interchain-security/tests/difference/core/quint_model/test/moduleOne.qnt' imports '/Users/offtermatt/projects/interchain-security/tests/difference/core/quint_model/test/moduleTwo.qnt'`

This seems like a bug to me, since there is no actual cyclic import when resolving down to the module level.

 The reason why I want to keep the modules separated like this is that in my example, module one is something like "staking_types", module onePrime is something like "staking_implementation". The split exists to avoid cyclic imports. Splitting each module into its own file works, but I'd prefer the case where the types and implementation modules can live in the same file.

konnov:
Closes #1108. Closes #1009. This is a refactoring of module sorting in phase 2. The new code simply delegates module sorting to the toposort. I find this solution to be much nicer and the code to be much cleaner.


quint/src/parsing/quintParserFrontend.ts
```diff
@@ -233,8 +213,79 @@ export function parsePhase2sourceResolution(
...
+          message: `Two modules have a conflict on the same name: ${mod.name}`,
+          reference: mod.id,
+        }
+        return { errors: [err], modules }
+      }
+      prev = just(mod)
+    }
+    // this line should not be reachable, unless there is a bug above
+    throw new Error('This line should not be reached. Report a bug.')
+  }
```
shonfeder, line 253: Shouldn't we do this during the same operation that forms the `namesToModule` map? (Eithir in the reduce, or moving that into this for loop).

 Advantages I see:

1. It would mean we only need to traverse the modules once, using the map lookup on the accumulator to check for duplicates at each addition
2. It would make the error detection clearer at the earliest stage at which the error can be detected

I would also suggest detecting all name clashes, rather than just the first two found. That seems more consistent with our general error reporting style.

shonfeder (approved):

shonfeder:
Looks great. Just one, non-blocking suggestion from me.

quint/src/parsing/quintParserFrontend.ts
```diff
@@ -233,8 +213,79 @@ export function parsePhase2sourceResolution(
...
+  // create the import graph
+  let edges = ImmutMap<bigint, ImmutSet<bigint>>()
+  for (const mod of modules) {
+    let imports = ImmutSet<bigint>()
+    for (const decl of mod.declarations) {
+      if (decl.kind === 'import' || decl.kind === 'instance') {
```
bugarela, line 259: Should this also sort exports?

```suggestion
      if (decl.kind === 'import' || decl.kind === 'instance' || decl.kind === 'export') {
```

bugarela (approved):

quint/src/parsing/quintParserFrontend.ts
konnov: Thanks for the suggestion, @shonfeder. I wanted to ignore it at first, since having three loops instead of one is not really crucial here. But now I have refactored the code and it looks much easier to read :)

quint/src/parsing/quintParserFrontend.ts
konnov: @bugarela I have accepted your suggestion, but then the failing tests showed me that we should not keep track of exports. I have added a clarifying comment. It's good that we have tests!

PR #1269, opened 2023-11-24
update the cheatsheet

konnov:
fix the typo in `select` as reported by @josef-widder


josef-widder (approved):
Thanks for the very fast fix! Looks great.

PR #1271, opened 2023-11-27
An example specification of the Cosmos SDK bank module

konnov:
I've prototyped this specification in January. Since then we have added plenty of good features in the language. It feels to be the right time for merging this spec.

The distinct feature of this example is literate programming. Both the functional specification and the state machine are introduced in the markdown text directly. So the markdown file is the only file that actually needs a review. The text description of the bank module is copied from Cosmos SDK. We do not have to proof-read text issues, if there are any.

This specification could be further improved. For instance, we could more tests. However, it is already large enough. It could still be interesting to the people who are looking at the bank module.


examples/cosmos/bank/bankTest.qnt
```diff
@@ -96,12 +98,13 @@ module bankTest {
...
     }
   }
+
   run sendTest = {
     init
       // the King has bananas and he can send it
       .then(send("king",
                  "donkeykong",
-                 Set("banana").mapBy(d => 2_000)))
+                 Map("banana" -> 2_000)))
```
konnov, line 107: Thanks! This indeed looks much simpler :)

bugarela: I just realized that I need to update the markdown, and then re-generate the file. I'm doing that right now!

bugarela (approved):
I didn't look into much detail, but it looks good to me!

konnov:
Thanks! Then let's merge it. I have enabled automerge.

bugarela:
For some reason, auto-merge didn't work, so I merged it manually

PR #1272, opened 2023-11-27
Fix effect signature for the `match` operator

bugarela:
Hello :octocat: 

The current effect for `match` (or `matchVariant`) is currenlty overpermissive. We should ensure that all possible branches result in the same variable updates, as we do in if-then-else and `any`. This changes the effect signature for `match` so the updates are consistent across eliminators.

### DIFF
--- a/quint/test/effects/inferrer.test.ts
+++ b/quint/test/effects/inferrer.test.ts
@@ -248,12 +248,22 @@ describe('inferEffects', () => {
 
     const [errors] = inferEffectsForDefs(defs)
 
-    errors.forEach(v =>
-      assert.deepEqual(v, {
-        children: [],
-        location: 'Inferring effect for f',
-        message: 'Result cannot be an opperator',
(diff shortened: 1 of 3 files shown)


quint/test/effects/inferrer.test.ts
```diff
@@ -248,12 +248,22 @@ describe('inferEffects', () => {
...
     const [errors] = inferEffectsForDefs(defs)
 
-    errors.forEach(v =>
-      assert.deepEqual(v, {
-        children: [],
-        location: 'Inferring effect for f',
-        message: 'Result cannot be an opperator',
-      })
-    )
+    assert.deepEqual([...errors.values()][0], {
```
bugarela, line 251: Unrelated, but I noticed this could result in a false negative with `forEach`.

shonfeder (approved):
Looks great! Thanks for the fix :D

PR #1276, opened 2023-11-29
Two fixes for Paxos simulation

bugarela:
Hello :octocat: 

This includes 2 changes that make `test` for Paxos get a :white_check_mark: (even though it has no tests, we had failures before):
1. Problem: type aliases no being found in the flattener. Solution: update `IRTransformer` to transform the types inside the sum type fields. There was also a related problem with printing that was fixed, and tests were updated accordingly.
2. Problem: nested definitions inside `assume` were not being flattened correctly, because the way we track `definitionDepth` was not considering `assume` (or instance overrides, which could also introduce the same issue). Solution: `definitionDepth` now is also incremented inside top level declarations. It is a bit tricky, and this is used in 4 different visitors, so instead of adding complexity to all of them, I moved handling of `definitionDepth` to the `IRVisitor` interface and its `walk*` functions.

This has the potential to help @shonfeder's work on making sumtypes work in Apalache, since the problem there seems to be related to (1).


quint/src/effects/inferrer.ts
```diff
@@ -87,7 +87,7 @@ export class EffectInferrer implements IRVisitor {
   private freeNames: { effectVariables: Set<string>; entityVariables: Set<string> }[] = []
 
   // the current depth of operator definitions: top-level defs are depth 0
-  private definitionDepth: number = 0
```
konnov, line 90: I am not sure about what's going on with the variable `definitionDepth` here.  Is it getting decremented anywhere? Also, it looks like the comment is not in sync with the initializer.

quint/src/ir/IRVisitor.ts
```diff
@@ -261,14 +265,23 @@ export function walkDeclaration(visitor: IRVisitor, decl: ir.QuintDeclaration):
...
-    case 'def':
     case 'typedef':
     case 'assume':
       walkDefinition(visitor, decl)
       break
+    case 'def':
+      // depth will be increased inside `walkDefinition`, so we set it to -1 in
+      // order for it to be 0 there
+      visitor.definitionDepth = -1
+      walkDefinition(visitor, decl)
```
konnov, line 283: I think I am a bit lost in the difference between top-level and inner definitions. Would not that code get triggered on both?

konnov (commented):
I feel like incrementing and decrementing the depth counter across multiple classes is a bit a fragile solution. Would not it make sense to introduce higher-level methods like `init()`, `enterDef`, `exitDef`?

quint/src/effects/inferrer.ts
bugarela: It is manipulated by the `walk*` functions, which is how the visitors are supposed to be used. The comment is correct, if you look at `this.definitionDepth` at some point and its value is 0, then you are in a top level definition.

I don't like this initialization here, but I couldn't think of a better way to do it. I wonder if extending this from a class (on top of implementing the interface) is better or worse.

quint/src/ir/IRVisitor.ts
bugarela: Top-level definitions are declarations, while inner definitions are NOT declarations. This is inside `walkDeclaration`, so the code would only be called right before walking top level definitions.

shonfeder:
This fixes the problem on the Apalache side, according to a test I just ran locally. :woman_dancing:.

CHANGELOG.md
```diff
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Topological sorting of modules (#1268)
 - The effect checker will now check for consistency of updates across different
   cases inside `match` (#1272)
+- Fix problems in the integration of sum types in `run` and `test` commands (#1276)
```
shonfeder, line 24:
```suggestion
- Fix problems in the integration of sum types in `run`, `test`, and `verify` commands (#1276)
```

shonfeder (approved):

quint/src/ir/IRVisitor.ts
```diff
@@ -312,9 +325,14 @@ export function walkDeclaration(visitor: IRVisitor, decl: ir.QuintDeclaration):
  * @returns nothing, any collected information has to be a state inside the IRVisitor instance.
  */
 export function walkDefinition(visitor: IRVisitor, def: ir.QuintDef): void {
+  if (visitor.definitionDepth !== undefined) {
+    visitor.definitionDepth++
+  }
```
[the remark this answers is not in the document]
bugarela: It means that the class (the concrete visitor) doesn't care about `definitionDepth` (it doesn't define or use it).

CHANGELOG.md
shonfeder: Only needs the changes in the or for Apalache. Nothing left on the quint side. :)

bugarela:
I can't think of a better way to implement the `definitionDepth` counter, but here are some more drastic changes to consider:
1. Add the `depth` attribute to `QuintDef` in the IR and set this while parsing. I know @konnov is, in general, against adding stuff to the IR.
2. Lookup the def in the lookup table whenever we need to know its depth. This might affect performance, and we'd also have to add maps from the definition id to "itself" (the entries in the lookup table include additional metadata such as depth) to the lookup table, in order to effectively `.get()` it.
3. Include an additional argument to the `enter/exitDef` and `enter/exitExpr` which yields the depth - or to all enter/exit functions. This would mean that even the visitors that don't care about the depth would need to be updated with that extra argument on the enter/exit functions it declares.

Do you think any of this two approaches are worth implementing? Or perhaps have a better idea to abstract this?

shonfeder:
I suggest going with your approach to fix the bugs, as I don't see it adding more complexity. It is just correcting the implementation and if it makes complexity more apparent it is only the complexity that is already required by our approach.

The other approaches you suggest seem worth considering, but we can address those in a followup issue, exploring the trade offs there.

WDYT?

konnov:
> 1. Add the `depth` attribute to `QuintDef` in the IR and set this while parsing. I know @konnov is, in general, against adding stuff to the IR.

I would like to avoid this one especially. It would be much harder to compose definitions, if we have to keep track of their depth.

konnov:
> 2\. Lookup the def in the lookup table whenever we need to know its depth. This might affect performance, and we'd also have to add maps from the definition id to "itself" (the entries in the lookup table include additional metadata such as depth) to the lookup table, in order to effectively `.get()` it.

This is probably the cleanest solution. After all, this is exactly what we need lookup tables for: To get all additional data without computing it ever again.

PR #1277, opened 2023-11-29
Fix sum type declaration collision

shonfeder:
Closes #1277

The implementation of the `isDefined` constraint was producing invalid type inferences in the presence of row types with the same labels coordinated with incompatible types.

The problem was that the `isDefined` constraint was being solved before other constraints that would have narrowed the type thru substitutions that would have prevented the ambiguity (e.g., by unifying thru the equality constraints induced by operator annotations). This fix takes the very simple course of introducing an ordering over constraints that places `isDefined` constraints lasts, and sorts constraints prior to solving.

A more robust solution should be considered if we encounter similar problems with constraint order again. Alternatives might include:

- Make the typing context available in our typing ruled and dispensing with the `isDefined` constraint.
- Support some form of back-tracking in our constraint solver to ensure we find the MGU consistent with a commutative conjunction constraint.
- Introduce disjunctive constraints and define `isDefined` using it.

Review by commit is probably helpful here, as some incidental cleanup is included and explained in the commit messages.

-------


quint/src/parsing/ToIrListener.ts
```diff
@@ -430,6 +430,10 @@ export class ToIrListener implements QuintListener {
         // This shouldn't be visible to users
         const paramName = `__${fieldName}Param`
 
+        // def declarations and the lambdas they are associated with should share
+        // the same ID
```
bugarela, line 434: Do we really need this? I feel a bit insecure about this changes because right now, every IR component has a distinct ID, and it might be the case that we are assuming this for some things across the code base.

I actually like the idea of making this a general thing, but I'm not 100% sure we can.

Violating this (perhaps unimportant) assumption of distinct ids only in this new case that's not heavily tested yet seems a bit dangerous to me.

shonfeder: I don't think we need this to fix any behavior, but i I introduced it because it is how, when I checked, the other declarations were working already, so i thought I was making making things more consistent! Sounds like we have a misunderstanding one one side or the other. Let me double check.

shonfeder: As per our discussion, I've reverted those changes. Thank you for the careful review and catching this.

I'll also prepare a followup to fix this for other operators.

bugarela (approved):
Looks great!

PR #1280, opened 2023-12-01
Fix operator declaraion lambda IDs

shonfeder:
Thanks to discussion with @bugarela, following the review comment at
https://github.com/informalsystems/quint/pull/1277#discussion_r1410621776 , we
realized that the lambdas constructed for the body of an operator `def`
declaration where using the wrong ID (they where reusing the IDs from the
declaration as a whole). But every IR component is meant to have its own ID, so
this is incorrect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and `import (...)` (#1276)
 - Fix incorrect type checking failure from interference between sum types
   sharing variant labels (#1275)
+- Fix the IDs generated for operator definition bodies (#1280)
 
 ### Security
 


quint/test/types/inferrer.test.ts
```diff
@@ -56,21 +68,24 @@ describe('inferTypes', () => {
...
-      [6n, '(bool) => int'],
-      [7n, 'bool'],
-      [8n, 'bool'],
-      [9n, 'int'],
-      [10n, 'bool'],
-      [11n, 'bool'],
-      [12n, 'int'],
-      [13n, 'int'],
-      [14n, '((bool) => int, bool) => int'],
+      [6n, '∀ t0, t1 . ((t0) => t1, t0) => t1'],
```
bugarela, line 88: These ids have a weird order (7-16, then 1-6). I guess if you feel like fixing, nice, otherwise its fine as well, this is not important.

bugarela (approved):
Thanks for taking care of this!

quint/test/types/inferrer.test.ts
shonfeder: I will. Or mess with it so long as the tests are passing. I wonder why the order is that way!

PR #1282, opened 2023-12-01
Added defs that extract values of a map as a set and as a list

ivan-gavran:
This PR adds to spells (added to `commonSpells`):
 - extracting a set of all values from a map (`mapValuesSet`)
 - extracting a list of all values from map (`mapValuesList`)


examples/spells/commonSpells.qnt
```diff
@@ -17,4 +17,34 @@ module commonSpells {
...
+    assert(mapValuesSet(Map(1 -> 2, 2 -> 3)) == Set(2, 3)),
+    assert(mapValuesSet(Map(1 -> 2, 2 -> 3, 3 -> 2)) == Set(2, 3)),
+  }
+
+   /// Get the list of values from a map.
+   ///
+   /// - @param __map a map from type a to type b
+   /// - @returns the list of all values in the map
+  pure def mapValuesList(__map: a -> b): List[b] = {
+    __map.keys().fold([], (__list, __k) => __list.append(__map.get(__k)))
```
shonfeder, line 40: I have two thoughts here:

Similar to the discussion at https://github.com/informalsystems/quint/pull/1107#discussion_r1300372909, a snag with operators that take sets to lists is that `fold`'s behavior is undefined for non-commutative operations (like `append`), due to sets being unordered.

Perhaps it would suffice if we have the `toList` method from #1107 (fixed as described in the liked comment) together with the `mapValusSet` added here. Then `mapValuesList` would be available just as the composition `m.mapValuesSet().toList(f)` (where `f` is the ordering function needed to ensure well defined behavior), without needing another named operator.

WDYT?

shonfeder (commented):
Thank you for expanding out the common spells!

We are lagging a bit in processing these vital PRs, but I'll aim to get them resolved this coming week.

examples/spells/commonSpells.qnt
konnov: I totally agree with Shon here. It would be much better to isolate the potentially problematic behavior of `fold` in one place

ivan-gavran: I understand the concern, @shonfeder ! However, the problem with using `m.mapValuesSet().toList()` in place of `mapValuesList` is that it removes duplicates. 

Think about values of the mapping being shares, and we wanted to get the sum of all of them. What would be a preferred way to achieve that? (That was the original use-case, and I perhaps jumped too quickly to a solution.)

shonfeder: Ah, I see! I didn't think about that :) 

We could implement this taking the comparator then. Alternatively, we could provide a `foldMap : (acc, (key, value)) => acc => acc` if the intermediate representation as a list is not really essential.

shonfeder: Maybe what would be best here is to have the inverse of `setToMap`: `mapToSet : Map[a, b] -> Set[(a, b)]`. Then one could fold over this as needed. We may need some warning about only using this on maps with a finite domain, but I think it would likely be useful. WDYT, @konnov?

konnov: > Maybe what would be best here is to have the inverse of `setToMap`: `mapToSet : Map[a, b] -> Set[(a, b)]`. Then one could fold over this as needed. We may need some warning about only using this on maps with a finite domain, but I think it would likely be useful. WDYT, @konnov?

Yes, this makes sense to me. This would simply convert a map to a relation. I thought that we already had something like that in Apalache, but we only have it in the opposite direction.

examples/spells/commonSpells.qnt
```diff
@@ -17,4 +17,42 @@ module commonSpells {
...
+    __map.keys().fold(Set(), (__acc, __k) => __acc.union(Set((__k, __map.get(__k)))))
+  }
+
+  run mapToSetTest = all {
+    assert(mapToSet(Map()) == Set()),
+    assert(mapToSet(Map(1 -> 2, 2 -> 3)) == Set((1, 2), (2, 3))),
+    assert(Map("a" -> 1, "b" -> 1, "c" -> 1).mapToSet().fold(0, (__acc, __kv) => __acc + __kv._2) == 3),
+  }
+  
+
```
bugarela, line 57:
```suggestion
```

bugarela (approved):
LGTM

examples/spells/commonSpells.qnt
ivan-gavran: Absolutely. Thanks :)

shonfeder:
Thanks for the contribution and for flagging the need for this, @ivan-gavran!

PR #1291, opened 2023-12-05
Release v0.17.1

bugarela:
### Added

- Add a `debug` built-in function for printing values to stdout (#1266)

### Fixed

- The effect checker now distinguishes variables from different instances (#1290)


romac:
Let's maybe mention that the function is available as `q::debug` instead of just `debug`?

shonfeder (approved):
@romac's suggestion on the release notes makes to me, but otherwise everything looks good :)

PR #1295, opened 2023-12-07
Detect import paths that only differ in capitalization

### LINKED ISSUE #1194: Importing by path breaks on case-insensitive file systems
Take this folder of specs:
https://github.com/cosmos/interchain-security/blob/aa2989e40e6fb4578adb5e6b94752c36670f3cdc/tests/difference/core/quint_model

Importantly, note this import statement:

https://github.com/cosmos/interchain-security/blob/aa2989e40e6fb4578adb5e6b94752c36670f3cdc/tests/difference/core/quint_model/ccv_statemachine.qnt#L4

If I change the line to
```
import Time.* from "./libraries/time"
```

both the vscode plugin and the quint typechecker give me errors (here is the error from the plugin):
```
[QNT102] Module with name 'Time' was already definedquint(QNT000)
```

Any capitalization other than `Time`, i.e. `timE`, `TIME`, ... does not work.
This seems like a bug - the capitalization in the filename does not seem to matter usually, i.e.
this works fine:

```
module Test {
    import Test2 from "./foo/test2"
}
```
and in file `foo/test2.qnt`
```
module Test2 {
}
```
works ok.
Also, now any capitalization works -
```
 import Test2 from "./fOo/TeSt2"
```
works fine. Not sure what is special about my spec to break it.

konnov:
Closes #1194. This PR adds detection of filenames that use different capitalization. If this is the case, the parser flags an error. We have been discussing the option of storing hashes instead of normalized filenames, but this seems to be an overkill at the moment.

### DIFF
--- a/examples/spells/commonSpells.qnt
+++ b/examples/spells/commonSpells.qnt
@@ -17,4 +17,40 @@ module commonSpells {
     assert(setSum(2.to(4)) == 9),
     assert(setSum(Set(-4, 4, 5)) == 5),
   }
+
+   /// Get the set of values of a map.
+   ///
+   /// - @param __map a map from type a to type b
+   /// - @returns the set of all values in the map
+  pure def mapValuesSet(__map: a -> b): Set[b] = {
+    __map.keys().map(k => __map.get(k))
+  }
+
+  run mapValuesSetTest = all {
+    assert(mapValuesSet(Map()) == Set()),
+    assert(mapValuesSet(Map(1 -> 2, 2 -> 3)) == Set(2, 3)),
+    assert(mapValuesSet(Map(1 -> 2, 2 -> 3, 3 -> 2)) == Set(2, 3)),
+  }
+
+   /// Convert a map into a set of pairs coordinating its keys and values
+   ///
+   /// - @param __map a map
+   /// - @returns the set of pairs coordinating the map's keys with its values
+   ///
+   /// ### Examples
+   ///
+   /// ```
+   /// val accountValues: Map[str, int] = Map("a" -> 1, "b" -> 1, "c" -> "1")
+   /// val sumOfValues: int = accountValues.mapToSet().fold(0, (acc, kv) => acc + kv._2)
+   /// assert(sumOfValues == 3)
+   /// ```
+  pure def mapToSet(__map: a -> b): Set[(a, b)] = {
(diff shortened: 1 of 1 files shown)


quint/cli-tests.md
```diff
@@ -381,3 +381,9 @@ Regression test for [#1108](https://github.com/informalsystems/quint/issues/1108
 
 <!-- !test check 1108 -->
     quint parse testFixture/_1052one.qnt
+
+### FAIL on run river::noSolution
```
bugarela, line 385:
```suggestion
### FAIL on parsing filenames with different casing
```

quint/cli-tests.md
```diff
@@ -381,3 +381,9 @@ Regression test for [#1108](https://github.com/informalsystems/quint/issues/1108
...
+
+<!-- !test exit 1 -->
+<!-- !test check parse testFixture/_1060case.qnt - parse case sensitive filenames -->
+    quint parse testFixture/_1060case.qnt
```
bugarela, line 389: Should we also check if the error matches the expected? I think this test would pass even before your change.

quint/src/parsing/quintParserFrontend.ts
```diff
@@ -202,6 +231,7 @@ export function parsePhase2sourceResolution(
           worklist.push([m, pathTrail.concat([importeePath])])
         })
         sourceToModules.set(importeePath.normalizedPath, newModules)
+        visitedPaths.set(importeePath.normalizedPath.toLocaleLowerCase(), importeePath.normalizedPath)
```
thpani, line 234: 1. Why are we using the locale version of `to*LowerCase` here?
2. Can we reuse the consts from above?
```suggestion
        visitedPaths.set(importeeLowerCase, importeeNormalized)
```

thpani (commented):
left one comment, rest LGTM

quint/cli-tests.md
konnov: I also wanted to do that. But it's a bit hard to do, since the `quint parse` outputs absolute paths.

konnov: I've fixed this test by moving it to `io-cli-tests.md` and comparing the outputs.

konnov:
I think I have addressed all of the comments. Please have a look

thpani (approved):
LGTM 👍

PR #1303, opened 2023-12-12
Introduce a new run operator `expect`

### LINKED ISSUE #1252: write post-condition assertions without introducing a new step
@hvanz reported an issue where he is trying to have an assertion after an action, and that produces a stuttering step which is not wanted.

He thought the built in `assert` could be interpreted in such a way where the step with the assertion doesn't make it through the trace. This is not true, but it is a valid idea.

Also https://github.com/informalsystems/quint/issues/1236 might help solving this issue.

konnov:
Closes #1252. Following the discussion raised by @hvanz, I have implemented the operator `A.expect(P)` with the following semantics:

- Evaluate action `A`:
  - When `A`'s result is `false`, emit a runtime error.
  - When `A`'s result is `true`:
    - Commit the variable updates.
    - Evaluate `P`:
      - If `P` evaluates to `false`, emit a runtime error (similar to `assert`).
      - If `P` evaluates to `true`, rollback to the previous state and return `true`.

Its semantics looks a bit complex, but this is basically what we want, if we do not want to have an extra stuttering step that calls `assert(P)`.

Here are the examples from the unit tests:

image


quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1303,6 +1307,57 @@ export class CompilerVisitor implements IRVisitor {
...
+        const savedVarsAfterAction = this.snapshotVars()
+        const savedNextVarsAfterAction = this.snapshotNextVars()
+        const savedTraceAfterAction = this.trace()
+        this.shiftVars()
+        // evaluate P
+        const predResult = pred.eval()
+        // rollback to the previous state in any case
+        this.recoverVars(savedVarsAfterAction)
+        this.recoverNextVars(savedNextVarsAfterAction)
+        this.resetTrace(just(rv.mkList(savedTraceAfterAction)))
```
bugarela, line 1349: Do we really need the rollback if the effect is already preventing updates inside the expect argument? I found the "rollback" explanation in the documentation confusing.

thpani:
> have implemented the operator `A.except(P)` with the following semantics

I think this should say `expect`?
Had me confused for a while 😄

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 
 ### Added
+
+- Add a run operator `A.expect(P)` to evaluate the result of `A` with `P` (#1303)
```
shonfeder, line 12: Something like this would be clearer IMO:


```suggestion
- Add a run operator `A.expect(P)` to test the state predicate `P` in the state resulting from applying action `A` (#1303)
```

doc/lang.md
```diff
@@ -1754,6 +1755,39 @@ to be disabled.
...
+
+The semantics of this operator is as follows:
+
+- Evaluate action `A`:
+  - When `A`'s result is `false`, emit a runtime error.
+  - When `A`'s result is `true`:
+    - Commit the variable updates.
+    - Evaluate `P`:
+      - If `P` evaluates to `false`, emit a runtime error (similar to `assert`).
+      - If `P` evaluates to `true`, rollback to the previous state and return `true`.
```
shonfeder, line 1778: Does 'rollback to the previous state' mean that the 'committed' variable updates are undone? So `A.then(B).expect(P)` would end up in the state after `A` has been applied but before `B`?

quint/src/runtime/impl/compilerImpl.ts
konnov: I have added clarifying comments about why we have to do that in the simulator

konnov:
> > have implemented the operator `A.except(P)` with the following semantics
> 
> I think this should say `expect`? Had me confused for a while 😄

Yeah, I keep typing `except` all the time. This probably comes from TLA+

doc/lang.md
konnov: Yeah, it's exactly that. Not really rolling back to a state. I will fix it here too.

konnov: I have fixed the description

PR #1304, opened 2023-12-12
Change the runtime behavior of `then`

### LINKED ISSUE #594: Introduce `ok()` as a counterpart of `fail()`
Consider the following test:

```
    run sendWithoutMintTest = {
        init.then(send(minter, "bob", 5))
            .fail()
    }
```

Although this test is expected to fail, we cannot guarantee that it fails only after calling `send(...)`. We have to make sure that `init` goes through, which may be non-trivial to show, as `init` is non-deterministic. We have to add `ok()` that complements `fail()` in tests.

konnov:
Closes #594. Similar to the behavior described in #594, @josef-widder and I have noticed that the current semantics of `A.then(B)` is tricky when `A.then(B)` evaluates to `false`. In this case, it is not clear whether `A` or `B` evaluated to `false`. This makes it hard to debug long runs. In general, engineers expect that a sequence of steps cannot just stop somewhere in the middle, which is how `A.then(B)` works now. This looks a bit like a deadlock, especially when we have non-deterministic choices or `any`.

This PR is changing the runtime semantics of `A.then(B)` in the simulator to the behavior that I believe is more consistent with normal programming languages. Namely, if `A` evaluates to `false`, then `A.then(B)` emits a runtime error saying that a computation cannot continue.

**This is a breaking change. Some tests may fail.** But this is good, these tests may catch problems that were not caught previously.

### DIFF
--- a/quint/src/runtime/impl/compilerImpl.ts
+++ b/quint/src/runtime/impl/compilerImpl.ts
@@ -1270,7 +1270,8 @@ export class CompilerVisitor implements IRVisitor {
     for (const action of actions) {
       nactionsLeft--
       result = action.eval()
-      if (result.isNone() || !(result.value as RuntimeValue).toBool()) {
+      const isFalse = result.isJust() && !(result.value as RuntimeValue).toBool()
+      if (result.isNone() || isFalse) {
         // As soon as one of the arguments does not evaluate to true,
         // break out of the loop.
         // Restore the values of the next variables,
@@ -1278,7 +1279,7 @@ export class CompilerVisitor implements IRVisitor {
         this.recoverNextVars(savedValues)
         this.resetTrace(just(rv.mkList(savedTrace)))
 
-        if (kind === 'then' && nactionsLeft > 0 && result.isJust() && !(result.value as RuntimeValue).toBool()) {
+        if (kind === 'then' && nactionsLeft > 0 && isFalse) {
           // Cannot extend a run. Emit an error message.
           const actionNo = actions.length - (nactionsLeft + 1)
           this.errorTracker.addRuntimeError(
--- a/quint/src/runtime/impl/compilerImpl.ts
+++ b/quint/src/runtime/impl/compilerImpl.ts
@@ -1254,8 +1254,11 @@ export class CompilerVisitor implements IRVisitor {
   }
 
   // compute all { ... } or A.then(B)...then(E) for a chain of actions
-  private chainAllOrThen(actions: Computable[], kind: 'all' | 'then',
-      actionId: (idx: number) => bigint): Maybe<EvalResult> {
+  private chainAllOrThen(
+    actions: Computable[],
+    kind: 'all' | 'then',
+    actionId: (idx: number) => bigint
+  ): Maybe<EvalResult> {
     // save the values of the next variables, as actions may update them
     const savedValues = this.snapshotNextVars()
     const savedTrace = this.trace()
@@ -1275,12 +1278,14 @@ export class CompilerVisitor implements IRVisitor {
         this.recoverNextVars(savedValues)
         this.resetTrace(just(rv.mkList(savedTrace)))
 
-        if (kind === 'then' && nactionsLeft > 0
-              && result.isJust() && !(result.value as RuntimeValue).toBool()) {
+        if (kind === 'then' && nactionsLeft > 0 && result.isJust() && !(result.value as RuntimeValue).toBool()) {
           // Cannot extend a run. Emit an error message.
           const actionNo = actions.length - (nactionsLeft + 1)
-          this.errorTracker.addRuntimeError(actionId(actionNo),
-            'QNT513', `Cannot continue in A.then(B), A evaluates to 'false'`)
+          this.errorTracker.addRuntimeError(
+            actionId(actionNo),
+            'QNT513',
+            `Cannot continue in A.then(B), A evaluates to 'false'`
+          )
           return none()
         } else {
           return result
@@ -1309,7 +1314,7 @@ export class CompilerVisitor implements IRVisitor {
     const args = this.compStack.splice(-app.args.length)
 
     const kind = app.opcode === 'then' ? 'then' : 'all'
-    const lazyCompute = () => this.chainAllOrThen(args, kind, (idx) => app.args[idx].id)
+    const lazyCompute = () => this.chainAllOrThen(args, kind, idx => app.args[idx].id)
 
     this.compStack.push(mkFunComputable(lazyCompute))
   }
@@ -1336,7 +1341,7 @@ export class CompilerVisitor implements IRVisitor {
               },
             }
           })
-          return this.chainAllOrThen(actions, 'then', (_) => app.args[1].id)
+          return this.chainAllOrThen(actions, 'then', _ => app.args[1].id)
         })
         .join()
     }
--- a/quint/src/runtime/impl/compilerImpl.ts
+++ b/quint/src/runtime/impl/compilerImpl.ts
@@ -1348,6 +1348,7 @@ export class CompilerVisitor implements IRVisitor {
               },
             }
           })
+          // In case the case of reps, we have multiple copies of the same action. This is why all occurrences have the same id.
           return this.chainAllOrThen(actions, 'then', _ => app.args[1].id)
         })
         .join()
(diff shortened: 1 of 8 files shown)


quint/test/runtime/compile.test.ts
```diff
@@ -80,56 +80,86 @@ function assertVarExists(kind: ComputableKind, name: string, input: string) {
...
+  }
+
+  return right(run)
+}
+
 // Scan the context for a callable. If found, evaluate it and return the value of the given var.
-// Assumes the input has a single callable
-// `callee` is used for error reporting.
-function evalVarAfterCall(varName: string, callee: string, input: string): Either<string, string> {
+// Assumes the input has a single callable `callee` is used for error reporting.
```
thpani, line 104: Can't parse this sentence.

quint/test/runtime/compile.test.ts
```diff
@@ -80,56 +80,86 @@ function assertVarExists(kind: ComputableKind, name: string, input: string) {
...
-    }
+    return callableFromContext(ctx, callee)
+      .mapRight(run => {
+        return run
+          .eval()
+          .map(res => {
+            if ((res as RuntimeValue).toBool() === true) {
+              // extract the value of the state variable
+              const nextVal = (ctx.evaluationState.context.get(kindName('nextvar', varName)) ?? fail).eval()
+              // using if-else, as map-or-unwrap confuses the compiler a lot
```
thpani, line 118: I don't know what to do with this comment.
Can we make it more relevant for the reader?

quint/test/runtime/compile.test.ts
```diff
@@ -80,56 +80,86 @@ function assertVarExists(kind: ComputableKind, name: string, input: string) {
...
-          const s = expressionToString(res.toQuintEx(idGen))
-          const m = `Callable ${callee} was expected to evaluate to true, found: ${s}`
-          return left<string, string>(m)
-        }
+  return evalInContext(input, callback)
+}
+
+// Evaluate a run and return the result.
+function evalRun(callee: string, input: string): Either<string, string> {
+  // use a combination of Maybe and Either.
```
thpani, line 141: Not sure where this applies?

thpani (commented):
Changes LGTM, I've left a few comments on the comments :)

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1273,7 +1277,19 @@ export class CompilerVisitor implements IRVisitor {
         // as evaluation was not successful.
         this.recoverNextVars(savedValues)
         this.resetTrace(just(rv.mkList(savedTrace)))
-        break
+
+        if (kind === 'then' && nactionsLeft > 0 && result.isJust() && !(result.value as RuntimeValue).toBool()) {
```
shonfeder, line 1281: Because of the way `toBool` is implemented, this can hide type errors if they make it through to here. 

https://github.com/informalsystems/quint/blob/main/quint/src/runtime/impl/runtimeValue.ts#L523-L529

```
  toBool(): boolean {
    if (this instanceof RuntimeValueBool) {
      return (this as RuntimeValueBool).value
    } else {
      return false
    }
  }
```

That is, if the `result.value` is, e.g., a string, then the final conjunct will be false, and we'll end up hiding the error. This could make it really hard to catch such a problem.

To work around this, we should add a test of `result.value instanceof RuntimeValueBool`.

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1254,7 +1254,11 @@ export class CompilerVisitor implements IRVisitor {
   }
 
   // compute all { ... } or A.then(B)...then(E) for a chain of actions
-  private chainAllOrThen(actions: Computable[], kind: 'all' | 'then'): Maybe<EvalResult> {
+  private chainAllOrThen(
+    actions: Computable[],
+    kind: 'all' | 'then',
+    actionId: (idx: number) => bigint
```
shonfeder, line 1260: Please document how this is meant to be used and why we need it.

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1273,7 +1277,19 @@ export class CompilerVisitor implements IRVisitor {
         // as evaluation was not successful.
         this.recoverNextVars(savedValues)
         this.resetTrace(just(rv.mkList(savedTrace)))
-        break
+
+        if (kind === 'then' && nactionsLeft > 0 && result.isJust() && !(result.value as RuntimeValue).toBool()) {
+          // Cannot extend a run. Emit an error message.
+          const actionNo = actions.length - (nactionsLeft + 1)
+          this.errorTracker.addRuntimeError(
+            actionId(actionNo),
```
shonfeder, line 1285: The use of `actionId` doesn't make sense to me. If we need to be fetching the id based on the action number, why is the ID hard-coded when used on line 1344?

quint/test/runtime/compile.test.ts
konnov: Yep, that was a weird left-over. Fixed

quint/test/runtime/compile.test.ts
konnov: Also removed

quint/src/runtime/impl/compilerImpl.ts
konnov: The whole code base of the simulator assumes that the IR is well-typed, which is ensured by the type checker. Sure, I will add an extra test, but what I should do in the case this test is failing.

quint/src/runtime/impl/compilerImpl.ts
konnov: When we execute `A.then(B)`, we want the error listener to highlight `A`. In case of `reps` below, we have multiple copies of `A`. This is why all occurences of `A` have the same id.

quint/src/runtime/impl/compilerImpl.ts
konnov: `RuntimeValue` encapsulates the types of the concrete values, so `result.value instanceof RuntimeValueBool` simply would not work here.

quint/src/runtime/impl/compilerImpl.ts
konnov: I have added a detailed documentation to this private method :)

quint/src/runtime/impl/compilerImpl.ts
konnov: I have improved the code a bit. However, adding defensive code to this very method does seem overkill to me. For starters, it would be harder to read. On top of that, we would have to ensure type correctness in the whole compiler then, which is not the purpose of this phase.

shonfeder: If the assumption is that everything is well typed and behavior is undefined otherwise, those conversions should be raising exceptions not returning undefined?

shonfeder: I guess this is a larger scale
Design problem and not limited of this pr, so I'm ok deferring this.

I would not surprised if this end up biting us tho, as we still send stuff that fails type checking to the simulator via the repl and we still have holes in the type system.

quint/src/runtime/impl/compilerImpl.ts
shonfeder: Private doesn't mean we don't need to be able understand what stuff is supposed to do. So thank you for taking time to add a comment.

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1325,7 +1348,7 @@ export class CompilerVisitor implements IRVisitor {
               },
             }
           })
-          return this.chainAllOrThen(actions, 'then')
+          return this.chainAllOrThen(actions, 'then', _ => app.args[1].id)
```
shonfeder, line 1351:
```suggestion
          // In case the case of reps, we have multiple copies of the same action. This is why all occurrences have the same id.
          return this.chainAllOrThen(actions, 'then', _ => app.args[1].id)
```

shonfeder (approved):
Please a comment along the lines I've suggested to re: the fixed id in `reps`. Otherwise, lgtm.

quint/src/runtime/impl/compilerImpl.ts
thpani: Maybe add an explanatory comment?

quint/src/runtime/impl/compilerImpl.ts
konnov: > If the assumption is that everything is well typed and behavior is undefined otherwise, those conversions should be raising exceptions not returning undefined?

Agree. I think the easiest thing we could do is to throw exceptions in runtime values, e.g., `RuntimeValueBool` could throw an exception when it is converted to something else. That would help us to defend from subtle typing errors and prevent us from introducing too much defensive code.

PR #1325, opened 2024-01-03
Report an error when trying to redefine keywords

### LINKED ISSUE #1247: Fix the grammar: an operator definition should use `qualId`

shonfeder:
Closes #1247 

This is a replacement for #1249 that corrects the error in the grammar targeted there while also addressing the usability concern flagged by @bugarela in https://github.com/informalsystems/quint/pull/1249#issuecomment-1816319740

It breaks the tests due to our use of the parser to extract documentation from https://github.com/informalsystems/quint/tree/main/quint/src/builtin.qnt

I propose that we simply produce https://github.com/informalsystems/quint/tree/main/doc/builtin.md manually. My initial thought is that using the parser to parse code that should be illegal doesn't seem justified by the need of deriving a bit of documentation, but we'll discuss when Gabriela is back from vacation.


bugarela (commented):
I find this better than our previous attempt, but I'm still failing to see what's the benefit to this change. I guess it is best to detect errors earlier, but I don't think it justifies decreasing the quality of the error message. Here's what I get now:
```
$ quint parse ../local/test.qnt
/home/gabriela/projects/quint/local/test.qnt:2:10 - error: [QNT101] Built-in name iff cannot be redefined
2:  val iff = 1
            ^

/home/gabriela/projects/quint/local/test.qnt:2:2 - error: [QNT101] Built-in name 'iff' is redefined in module 'a'
2:  val iff = 1
    ^^^^^^^^^^^

error: parsing failed
```

Also, this introduces a sort of special case for "infix" operators (the ones listed under `keywords` in the grammar), while the other built-in operators would have redefinition errors reported with the current name-resolution error. I'd prefer those to behave the same, since we can't expect users to tell the difference between them.

shonfeder:
Oh, if we are already catching all of these, then I don't think I understand what #1247 is about :) if no errors are getting thru now, the. I agree we don't need this addition.

PR #1333, opened 2024-01-12
Fix API documentation for `allLists`

bugarela:
Hello :octocat: 

I had a misunderstanding of the behavior of `allLists` and that was reflected in the docs I wrote. This fixes the doc.


shonfeder (approved):
Thanks for the fix！

PR #1337, opened 2024-01-12
Fix incorrect parsing of qualified type constructors

### LINKED ISSUE #1336: Types constants referenced via qualified access to modules with lowercase names are mistaken for type variables.
Types constants referenced via qualified access to modules with lowercase names are mistaken for type variables.

E.g., this typechecks

```
module e {
  type T = | Foo
}

module m {
  import e

  // This should be an operator that should only be the identity on
  // type e::T, with the sole inhabitant `Foo`:
  def f(x: e::T): e::T = x

  // But it is actually the polymorphic identity operator
  val ex_1: int = f(1)
  val ex_2: str = f("foo")
}
```

Related to #1081

shonfeder:
Closes #1336

The fix is:

- parse two different kinds of identifier tokens: IDs starting with lowercase
  letters and IDs starting with upper case
- factor out the `typeConstOrVar` into two separate rules: one for type
  constants and one for type vars, where the latter can only be produced by a
  lowercase non-qualified `LOW_ID`
  
This will also be needed for
https://github.com/informalsystems/quint/issues/1297, wherein type parameters in
declarations can only be variables.


quint/src/generated/Quint.g4
```diff
@@ -285,7 +289,10 @@ LPAREN          :   '(' ;
 RPAREN          :   ')' ;
 
 // other TLA+ identifiers
```
bugarela, line 291: Unrelated, but I guess we should remove this old comment
```suggestion
```

bugarela (approved):
LGTM! Thanks for taking care of this.

There is a breaking integration test, but it should be only a matter of updating the expected error message.

shonfeder:
Thanks for the review!

Issue #1052, opened 2023-07-17
Remove `compileFromCode` and use more appropriate interfaces


bugarela:
Mentioned in https://github.com/informalsystems/quint/pull/1051

`compileFromCode` re-parses and re-analyzes the code, while previous cli stages should have been responsible for that already. We need more appropriate interfaces for this.

bugarela:
With a proper solution for this, it should be easy to get the `sourceCode` information in all cases in the repl, see https://github.com/informalsystems/quint/pull/1224

PR #1340, opened 2024-01-15
Ensure runtime errors are reported in simulation

### LINKED ISSUE #1339: Sometimes, runtime errors are not reported in simulation
Over the past month, I've struggled multiple times with a hidden runtime error. Unfortunately, I was not able to reproduce this with a minimal example, but the scenario is:
1. I have a call to `set` with a non-existing key in my spec
2. I run `quint run` for that spec, expecting a counterexample
3. No counterexample and no error are reported
4. I inspect the spec until I see my mistake with `set`, and then change it to a `put`.
5. `quint run` now finds the expected counterexample.

This is really bad because
1. It appears as if the simulator can't find any counterexamples, while it would if not for the hidden error with `set`
2. There is no reporting for an actual error that is occurring.

bugarela:
Hello :octocat: 

Fixes #1339 

There is currently some mix-up with conditionals that leads to runtime errors not being reported when the trace recorder doesn't find a violation. This fixes that.

I'm having a hard time coming up with a small example for an integration test, so no test is added here.

Also, in the scenario where the errors weren't being reported, and now are, the errors come from a `compileFromCode` call, and most likely that is the reason for the lack of a location for those errors - since the whole code is re-parsed and new ids are generated. This should be fixed by #1052  


quint/src/cliCommands.ts
```diff
@@ -66,7 +66,7 @@ interface OutputStage {
   failed?: string[]
   ignored?: string[]
   // Test names output produced by 'run'
```
shonfeder, line 68: I know it wasn't introduced in your changes, but it looks like this comment corresponds to the field you are updating. Do you suppose this is out of place or out of date? Or do you know what it means w/r/t `status`?

quint/src/cliCommands.ts
```diff
@@ -66,7 +66,7 @@ interface OutputStage {
...
-  status?: 'ok' | 'violation' | 'failure'
+  status?: 'ok' | 'violation' | 'failure' | 'error'
```
shonfeder, line 69: What does it mean that we know allow `error` as a status in `OutputStage` and that we didn't previously?

quint/src/cliCommands.ts
```diff
@@ -530,11 +530,7 @@ export async function runSimulator(prev: TypecheckedStage): Promise<CLIProcedure
   const mainEnd = prev.sourceMap.get(mainId)!.end!.index
   const result = compileAndRun(newIdGenerator(), mainText, mainStart, mainEnd, mainName, mainPath, options)
 
-  if (result.status === 'error') {
-    const newErrors = result.errors.map(mkErrorMessage(prev.sourceMap))
-    const errors = prev.errors ? prev.errors.concat(newErrors) : newErrors
-    return cliErr('run failed', { ...simulator, errors })
-  } else {
+  if (result.status !== 'error') {
```
shonfeder, line 533: The logic all the way down to 579 is pretty hard to follow :sweat_smile: 

Due to the severity of this issue, I'm OK with merging your fix as is. But I suggest we do an immediate followup to refactor this logic so it is possible to follow, and would propose we use the approach in https://github.com/informalsystems/quint/blob/main/CONTRIBUTING.md#ensure-exhaustive-matches to take advantage of the type system to ensure we're handling all the possible statuses.

I'm happy to do that followup work if you've other things on your plate.

quint/src/cliCommands.ts
shonfeder: Also, how does this relate to the changes here? IIUC, the `runSimulator` function should only be able to return the statuses that are defined in https://github.com/informalsystems/quint/blob/d4b7b0f713bb8f1f693aa2454127f96c771b45f0/quint/src/cliCommands.ts#L148-L151

So I'm not clear on how the status in `OutputStage` comes into play.

quint/src/cliCommands.ts
bugarela: Seems like it is out of date, I'll try to find out more in the history

quint/src/simulation.ts
```diff
@@ -129,12 +129,17 @@ export function compileAndRun(
...
+      status = 'error'
+    }
+
     return {
       status: status,
       vars: evaluationState.vars.map(v => v.name),
       states: topFrame.args.map(e => e.toQuintEx(idGen)),
       frames: topFrame.subframes,
-      errors: ctx.getRuntimeErrors(),
+      errors: errors,
```
shonfeder, line 142: I see: looks like this is the core issue? 

We have not tied construction of `errors` to the correct value for status. Mabye the right fix here would be to define

https://github.com/informalsystems/quint/blob/d4b7b0f713bb8f1f693aa2454127f96c771b45f0/quint/src/simulation.ts#L35

So that it also caries all the data essential to each status? So, e.g., for an error `{kind: 'error', errors: errors}` and remove the `errors` field from

https://github.com/informalsystems/quint/blob/d4b7b0f713bb8f1f693aa2454127f96c771b45f0/quint/src/simulation.ts#L37-L47

This is also something we can consider in a followup.

shonfeder (approved):
LGTM! Good find and quick fix. Let's cut a release?

quint/src/cliCommands.ts
shonfeder: Cool cool. Super low priority. I'm fine with us getting this fix in and cutting a release ASAP, if you are. I agree that this is a very bad and we should patch it up.

bugarela: Seems like it came from here https://github.com/informalsystems/quint/commit/6ebf042f5117e6a28013122f7daf0b359d6447df#diff-229e3647dd4a8f61338fb58a6266e6fca5690a1dcb1a0550cd0c1fba2b372d7dR53

I'd rule this as a copy-paste mistake (copy pasted from the "// Test names output produced by 'test'" and then it was never updated to match the actual meaning of this attribute)

quint/src/cliCommands.ts
bugarela: The `'error'` case comes from here:
https://github.com/informalsystems/quint/blob/gabriela%2Ffix-1339/quint/src/simulation.ts#L35

Previously, we had a separate error reporting for the case of `'error'`, which didn't include the status attribute (lines 534-536 before changes). Now, all errors are reported in the same manner (lines 573-579 after changes), which include the status attribute, which can now be `'error'` since it is not a special case anymore.

I guess I have to update the `SimulatorStage` type as well. It is weird to me that typescript didn't give me a type error here.

bugarela: Ah, I see why we don't need those in `SimulatorStage`: the `SimulatorStage` is only ever returned in case of success (`'ok'`). Otherwise, we call `cliErr`, which results in an `OutputStage`. In fact, we only need to have the `'ok'` option there. This typechecks:

```ts
interface SimulatorStage extends LoadedStage {
  status: 'ok'
  trace?: QuintEx[]
}
```

Anyway, I'll leave this for the future refactoring.

PR #1349, opened 2024-01-17
Improve runtime errors in the simulator

bugarela:
Hello :octocat: 

This is a followup to https://github.com/informalsystems/quint/pull/1340 since I found a (hacky) way to include locations in the reported errors. We should remove this when #1052 is fixed and errors can be reported normally. The reason for this is that it is very hard to find where runtime errors could potentially happen in a big spec, and having the location info makes all the difference.

For this, we compute the locs for the error while we still have access to the `compileFromCode` inputs and outputs, and then manipulate the `QuintError` to include those locs in a hacky way:
1. We remove the (optional) `reference` field, so no locs are included in the final error - since those would be messed up anyaway
2. We then include the locs as text + highlights in the error message, reconstructing a `QuintError` with a pre-formatted message in it's message field.

I added a lot of comments so this is not super shady when we come back to it. The end result (what users see) looks perfectly normal :ninja: 

### DIFF
--- a/quint/src/simulation.ts
+++ b/quint/src/simulation.ts
@@ -20,7 +20,6 @@ import { SourceLookupPath } from './parsing/sourceResolver'
 import { QuintError } from './quintError'
 import { mkErrorMessage } from './cliCommands'
 import { createFinders, formatError } from './errorReporter'
-import { uniqBy } from 'lodash'
 
 /**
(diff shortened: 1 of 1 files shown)


shonfeder (approved):
Great workaround! Thanks for leaving such clear comments :D

PR #1350, opened 2024-01-18
Sf/refactor run cmd

shonfeder:


shonfeder:
I'm still working on this refactoring work, but this little chunk is good enough to be reviewed I think, so marking it as such to break up the code size if the PRs. Review by commit is probably helpful here.

quint/src/runtime/testing.ts
```diff
@@ -74,22 +74,31 @@ export interface TestResult {
...
     return left([{ code: 'QNT405', message: `Main module ${mainName} not found` }])
   }
 
+  const recorder = newTraceRecorder(options.verbosity, options.rng)
   const ctx = compile(compilationState, newEvaluationState(recorder), lookupTable, options.rng.next, main.declarations)
 
+  const ctxErrors = ctx.syntaxErrors.concat(ctx.compileErrors, ctx.analysisErrors)
+  if (ctxErrors.length > 0) {
+    // In principle, these errors should have been caught earlier.
+    // But if they did not, return immediately.
```
bugarela, line 98: I know this comment isn't new, but I'm skeptical that we'd have caught compile errors before trying to compile :thinking:

shonfeder: True, I guess this only applies to the syntax errors and type checking errors, which would have prevented from moving on to this stage in the pipeline.

PR #1352, opened 2024-01-22
Cleaning up branching logic in runSimulator

shonfeder:
Address the cleanups noted in discussion of #1340

### DIFF
--- a/quint/src/cliCommands.ts
+++ b/quint/src/cliCommands.ts
@@ -302,7 +302,7 @@ export async function runRepl(_argv: any) {
  */
 export async function runTests(prev: TypecheckedStage): Promise<CLIProcedure<TestedStage>> {
   // TODO: Logic should be consistent with runSimulator?
-  const verbosityLevel = !prev.args.out ? prev.args.verbosity : 0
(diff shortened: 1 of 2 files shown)


quint/src/cliCommands.ts
```diff
@@ -553,53 +547,60 @@ export async function runSimulator(prev: TypecheckedStage): Promise<CLIProcedure
   const mainEnd = prev.sourceMap.get(mainId)!.end!.index
   const result = compileAndRun(newIdGenerator(), mainText, mainStart, mainEnd, mainName, mainPath, options)
 
-  if (result.status !== 'error') {
-    if (verbosity.hasResults(verbosityLevel)) {
-      const elapsedMs = Date.now() - startMs
+  const elapsedMs = Date.now() - startMs
+
+  switch (result.outcome.status) {
```
shonfeder, line 552: This is now case analysis, so there is one, and only one, branch for each outcome. This also ensures that we'll get static checks to enforce we handle new statuses if they are introduced.

bugarela (approved):
Great improvement!

PR #1355, opened 2024-01-25
Add support for parsing polymorphic type declarations

### LINKED ISSUE #1296: Add IR nodes for type abstraction and type function application

### LINKED ISSUE #1297: Add parser support for polymorphic type declaration

shonfeder:
Closes #1296

Adds support for two new kinds in the types IR:

1. Type abstraction, giving us type-level functions
2. Type application, giving us type-level function application

Closes #1297

Adds support for parsing polymorphic type declarations and type constructor application. This enables forming types of the kinds above as follows:

1. Type abstractions is produced only by type declarations of the following form:

   ```scala
   type T[a,b] = (a, b)
   ```
 
   Similar to operator declarations, there are parsed into the assignment of a name to an abstraction. I.e., in this case, `T = Λ(a, b).(a,b)`

2. Type application generalizes our current special-cased support for `List` and `Set`. E.g., we apply the polymorphic `T` defined above as `T[int, str]`.


shonfeder:
As is always the case with changes to the parser, nearly all the lines in the diff are from updates to fixtures and the generated parser code.

quint/package.json
```diff
@@ -72,6 +62,8 @@
     "compile": "genversion -e src/version.ts && tsc && copyfiles -u 1 ./src/reflection.proto ./src/builtin.qnt ./dist/src/",
     "prepare": "rm -rf ./dist && npm run compile && chmod +x ./dist/src/cli.js",
     "test": "mocha --reporter-option maxDiffSize=0 -r ts-node/register test/*.test.ts test/**/*.test.ts",
+    "test-w": "while inotifywait -r -e close_write ./ ; do npm run test; done",
```
bugarela, line 65: What does this do? I'm curious

quint/src/ir/quintTypes.ts
```diff
@@ -98,6 +98,29 @@ export function sumType(labelTypePairs: [string, QuintType][], rowVar?: string,
...
+ */
+export interface QuintAbsType extends WithOptionalId {
+  kind: 'abs'
+  vars: QuintVarType[] /** The bound variables */
+  body: QuintType /** The body of the abstraction */
+}
+
+/** Type application: (Λτ.Τ)υ
+ *
+ * In system-F, this corresponds to (Λτ.Τ)υ
```
bugarela, line 113: :nail_care: 
```suggestion
 * In System-F, this corresponds to (Λτ.Τ)υ
```

quint/src/parsing/ToIrListener.ts
```diff
@@ -485,6 +514,26 @@ export class ToIrListener implements QuintListener {
...
+      ctx._typeVars
+        .map(
+          _ =>
+            this.popType().unwrap(() =>
+              fail('internal error: type parameter parsed with no type variable')
+            ) as QuintVarType
+        )
+        // The stack stores the variables in reverse order
+        .reverse()
+    )
```
bugarela, line 534: I believe we should use `popMany` here as we are doing for most of the popping in this file. This also has a fallback that enables error recovery (improving our error reporting).
```suggestion
    return popMany(this.typeStack, ctx._typeVars().length, this.undefinedType(ctx)) as QuintVarType[]
```

quint/src/parsing/ToIrListener.ts
```diff
@@ -1047,13 +1096,30 @@ export class ToIrListener implements QuintListener {
...
 
+  // E.g., Result[int, str]
+  exitTypeApp(ctx: p.TypeAppContext) {
+    const id = this.getId(ctx)
+    const args: QuintType[] = ctx._typeArg
+      .map(_ =>
+        // We require that there is one parsed type for each typeArg recorded
+        this.popType().unwrap()
+      ) // The stack stores the arguments in reverse order
+      .reverse()
```
bugarela, line 1107: Same here
```suggestion
    const args = popMany(this.typeStack, ctx._typeArg.length, this.undefinedType(ctx))
```

quint/src/types/base.ts
```diff
@@ -29,6 +29,7 @@ export interface TypeScheme {
 
 export type Signature = (_arity: number) => TypeScheme
 
+// TODO: Why aren't we noting type variables in the schema here?
```
bugarela, line 32: Not sure I understand the question, but this `toScheme` is just a helper to convert a `QuintType` into a `TypeScheme` with no quantified variables.

quint/src/parsing/ToIrListener.ts
```diff
@@ -363,43 +367,68 @@ export class ToIrListener implements QuintListener {
     this.declarationStack.push(def)
   }
 
-  // type Alias = set(int)
+  // E.g.,
+  //
+  // - type Alias = Set(int)
+  // - type Constr[a, b] = (Set(a), Set(b))
```
bugarela, line 373: This was already wrong before your PR, but here's the fix anyway
```suggestion
  // - type Alias = Set[int]
  // - type Constr[a, b] = (Set[a], Set[b])
```

bugarela (approved):
I can't wait to start using this!
Just left a few comments, specially on the `ToIRListener` where we should have fallbacks to enable error recovery on this stage of parsing.

quint/package.json
shonfeder: this will watch all the files in the quint dir and rerun the tests whenever they are saved!

quint/src/parsing/ToIrListener.ts
shonfeder: I'm a bit torn. I like reusing code, ofc. but I also find use of wrap here quite clear and it is nice to just make use of combinators when possible. But I'm more motivated by the question of correctness: this isn't actually something we should repair with a default fallback, because if there are `_typeVars` but there are not types on the stack, than our parser is broken (either ANTLR is not parsing correctly or we have broken logic in the `ToIrListener`) and we want an internal error in that case, *not* to cover over the error with a generated type. WDYT?

quint/src/types/base.ts
shonfeder: But why would we drop the type variables from a quint type if it had them?

quint/src/parsing/ToIrListener.ts
bugarela: > if there are `_typeVars` but there are not types on the stack, than our parser is broken

I find it hard to know things like this for sure, because ANTLR produces weird stuff when  there is a parse error. So, to be sure, I added fallbacks almost everywhere. You probably have a better understanding of the parser than I do, and this assumption seems correct, so I'm fine trusting that and moving forward. Alternatively, we could stay on the safer side and use the fallback - `undefinedType` will produce a debug line when called, so this shouldn't be invisible to us.

quint/src/types/base.ts
bugarela: `typeVariables` in `TypeScheme` corresponds only to universally quantified type variables, not all type variables. Here's one example where we decide what to quantify:

https://github.com/informalsystems/quint/blob/main/quint/src/types/constraintGenerator.ts#L419

and the `toScheme` is used when we don't want to quantify anything.

quint/src/parsing/ToIrListener.ts
shonfeder: Good point: if we are still getting errors, then we won't like lose track of this... 

That said, my understanding is that we cannot ever hit this method unless the generated parser successfully parses a `typeDeclaration` and any successful parse of a type declaration should either have consumed a type (put on the stack) for each type variable in the definition head. If this understanding turns out wrong, I'd like to get a hard failure and an error report rather than have (users including myself) trying to figure out why they have a parse error.

So, I'll stick with this as is since you're OK with it, and I'll be prepared to revise my understanding and replace this if we hit these assertion failures! :)

quint/src/types/base.ts
shonfeder: I see. Thank you for clarifying! I'll replace this comment with something describing that logic. :pray:

shonfeder: Actually, I have a followup :)

Should this only be applied to types which have no free variables? Or is it valid to apply this to an open term?

bugarela: I think so! We should probably enforce that requirement. I'm not 100% sure tho. The requirement makes sense, and looking at its usages on the code, I'm like 80% sure they respect this requirement.

shonfeder: Cool. I added a comment to this effect. Thanks for explaining!

PR #1359, opened 2024-02-01
Remove erroneous occurs check from "effect entity" unfication

### LINKED ISSUE #1356: Effect checker gives invalid cyclical binding error
I have run into a strange error message by the type checker on this specification:

```
module t {
    pure def ok(value: a): { v: a, err: str } = {
        { v: value, err: "" }
    }

    pure def onOk(r: { v: a, err: str },
                  callback: (a) => { v: a, err: str }): { v: a, err: str } = {
        if (r.err == "") {
            callback(r.v)
        } else {
            r
        }
    }

    pure def foo(s: int, bar: (int, int) => { v: int, err: str }): { v: int, err: str} = {
        ok(s)
            .onOk((v2) => bar(v2, 3))
            .onOk((v3) => bar(v3, 4))
    }
}
```

This is what the typechecker is showing:

```sh
$ quint typecheck t.qnt
t.qnt:16:9 - error: [QNT000] Can't bind _v18 to _v18, _v21: cyclical binding
Trying to unify entities [_v18, _v21] and [_v18]
Trying to unify Read[_v18, _v21] and Read[_v18]
Trying to unify (Read[_v18, _v21]) => Read[_v23] and (Read[_v18]) => Read[_v21]
Trying to unify (Read[_v22], (Read[_v22]) => Read[_v23]) => Read[_v22, _v23] and (Read[_v18, _v21], (Read[_v18]) => Read[_v21]) => _e10
Trying to infer effect for operator application in onOk(onOk(ok(s), ((v2) => bar(v2, 3))), ((v3) => bar(v3, 4)))

16:         ok(s)
            ^^^^^
17:             .onOk((v2) => bar(v2, 3))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
18:             .onOk((v3) => bar(v3, 4))
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: typechecking failed
```

I guess, I am just pushing the boundaries of Quint way too much. The error message is quite confusing.

shonfeder:
Closes #1356

The originating issues was caused by fairly subtle error in our unification algorithm for the effect checker, which was making an invalid occurs check. This error was hard to spot, because (I think) we are not used to thinking of unification sets! I think we usually think of unification in terms of the "geometric structure" of the syntactic terms being unified. E.g., when unifying lists, `[A, B] =.= [C, D]` iff `A =.= C, B =.= D`, and `A =.= [1, 2, A]` should fail an occurs check because it would require unifying `A` with a term that includes itself as a sub-term. However, the extensional nature of sets invalidates the "evident" geometric intuition that guides us in the previous examples. `A =.= B ∪ C ∪ A` is true iff `B =.= C =.= A`.

This is also explained in a bit more detail in comments added with this change set.

The PR also includes a regression test and some simplification of the effect inference code, introduced during the debugging process.

### DIFF
--- a/quint/src/effects/base.ts
+++ b/quint/src/effects/base.ts
@@ -178,10 +178,23 @@ function bindEffect(name: string, effect: Effect): Either<string, Substitutions>
 }
 
 function bindEntity(name: string, entity: Entity): Either<string, Substitutions> {
-  if (entityNames(entity).includes(name)) {
-    return left(`Can't bind ${name} to ${entityToString(entity)}: cyclical binding`)
-  } else {
-    return right([{ kind: 'entity', name, value: entity }])
+  switch (entity.kind) {
+    case 'concrete':
+    case 'variable':
+      return right([{ kind: 'entity', name, value: entity }])
+    case 'union':
+      if (entityNames(entity).includes(name)) {
+        // An entity variable (which always stands for a set of state variables)
+        // unifies with the union of entities that may include  include itself,
+        // iff the variables unify.
+        //
+        // I.e.:
+        //
+        //   (v1 =.= v1 ∪ ... ∪ vn) <=> (v1 =.= ... =.= vn)
+        return right(entity.entities.map(e => ({ kind: 'entity', name, value: e })))
+      } else {
+        return right([{ kind: 'entity', name, value: entity }])
+      }
   }
 }
 
--- a/quint/src/effects/base.ts
+++ b/quint/src/effects/base.ts
@@ -81,11 +81,11 @@ export interface StateVariable {
  * entity variable or a combination of other entities.
  */
 export type Entity =
-  /* A list of state variables */
+  /* A set of state variables, represented as a list */
   | { kind: 'concrete'; stateVariables: StateVariable[] }
-  /* A variable representing some entity */
+  /* A variable representing some set of entities */
   | { kind: 'variable'; name: string; reference?: bigint }
-  /* A combination of entities */
+  /* The union of sets of entities, represented as a list */
   | { kind: 'union'; entities: Entity[] }
 
 /*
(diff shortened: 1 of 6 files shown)


quint/src/effects/base.ts
```diff
@@ -177,11 +177,36 @@ function bindEffect(name: string, effect: Effect): Either<string, Substitutions>
...
-function bindEntity(name: string, entity: Entity): Either<string, Substitutions> {
-  if (entityNames(entity).includes(name)) {
-    return left(`Can't bind ${name} to ${entityToString(entity)}: cyclical binding`)
-  } else {
-    return right([{ kind: 'entity', name, value: entity }])
+function bindEntity(name: string, entity: Entity): Substitutions {
+  switch (entity.kind) {
+    case 'concrete':
+    case 'variable':
+      return [{ kind: 'entity', name, value: entity }]
```
bugarela, line 184: Don't we need to check if `name === entity.name` here to avoid cyclical bindings?

quint/src/effects/inferrer.ts
```diff
@@ -149,23 +149,13 @@ export class EffectInferrer implements IRVisitor {
...
-    const resultEffect: Effect = { kind: 'variable', name: this.freshVarGenerator.freshVar('_e') }
-    const arrowEffect = paramsResult
-      .map(params => {
-        const effect: Effect = {
-          kind: 'arrow',
-          params,
-          result: resultEffect,
-        }
+    // Fetch all the effects for the arguments
+    const argEffects = mergeInMany(expr.args.map((a: QuintEx) => this.fetchResult(a.id).map(e => e.effect)))
```
bugarela, line 154: I feel like we do need the `newInstance` here. I'm thinking about a scenario like:

```
def f(x) = x // polymorphic operator
def op(f1: (a) => a, f2: (b) => b, x1: a, x2: b): (a,b) = (f1(x1), f2(x2))
val foo = op(f, f, 1, "a")
```

When fetching the results for the arguments, the first and the second argument (`f` and `f`) need to have different types (i.e. `(t0) => t0` and `(t1) => t1`), while with your approach, they would have the same types and I believe typechecking would fail.

bugarela: ok, this is actually for the effects and not types, but afaiu it's all the same. Perhaps we can't reproduce this issue in the effect system right now, tho. But I still feel like the correct thing here would be to use `newInstance`

quint/src/effects/base.ts
shonfeder: I don't think so. `a =.= a` is ok. It just unifies to an empty substitution. Cyclicality is only a problem when a variable is to unify with a term that includes itself as a (strict) sub-term.

But if you can think of a case that we should be testing for but aren't, I'll add a test to see if we can trigger some problem here.

quint/src/effects/inferrer.ts
shonfeder: Ah, I see. I can revert this part then.

quint/src/effects/base.ts
bugarela: Ah, I see! So there is no such thing as cyclical bindings in our effect system anymore, right?

quint/test/effects/base.test.ts
```diff
@@ -375,48 +375,39 @@ describe('unify', () => {
...
-      assert.isTrue(result.isLeft())
+    it('can unify entities when a single variable in the lhs appears in a union on the rhs', () => {
+      // E.g., given the unification problem
+      //
+      //    v1 =.= v1 ∪ v2
+      //
+      // We should be able to form a valid substitution iff v1 =.= v2, since
+      // this then simplifies to
+      //
+      //   v1 =.= v1
```
bugarela, line 386: Should this be
```suggestion
      //   v1 =.= v1 =.= v2
```

quint/src/effects/base.ts
shonfeder: Not at the level of "entity" variables, because they can only stand for sets of state variables. But there might be something that can happy with effect variables. I haven't looked at that logic as part of this. Could be that we don't have rich enough structures to make that possible tho?

quint/test/effects/base.test.ts
shonfeder: That is the implication, and here I am showing just the simplified form. But maybe I should make that more explicit in the comment.

quint/src/effects/base.ts
bugarela: Oh, right. At the effect level we have operators, so we should have cyclical bindings in things like `e0 =.= (e0) => Read['x']`.

Thanks for helping me get this context back in and understanding the scope of your solution!

bugarela (approved):
Great finding! I've been scared of unification of sets since I first wrote this stuff, so I'm glad you were able to reason about this and implement proper binding.

quint/test/effects/base.test.ts
bugarela: I thought the `v1 =.= v2` part was the most relevant here, since that's the part that can later result in an unification error

quint/src/effects/base.ts
shonfeder: Thanks for talking it thru with me and for the review! That is exactly right w/r/t effect variables. I think what we have here is just a special case due to simplicity of the "entity variables".

PR #1363, opened 2024-02-08
Use same register for variables with the same id

bugarela:
Hello :octocat: 

This fixes a problem I found recently when using multiple modules and no instances. Currently, our registers (which store values for state variables) are indexed by both the variable name and variable id. We already had some code ensuring that, if two variables have the same name (but a different id), they'd have the same register - this is important for instances, and flattening ensures that no two different state variables can have the same name.

Now, I'm adding a very similar piece of code to ensure variables with the same id get the same register, even if they have different names. This part is important for imports, since qualified imports can make it so the same variable is referred by different names.

I'm not confident this will work for all combinations of imports and instances, but it definitely covers more ground than what we currently have.

For testing, I just updated the existing `imports.qnt` example to have the complexity needed to hit the problem (fixed by this).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Erroneous effect checking failure resulting from invalid occurs check. This
   error prevented some valid specs from being simulated or verified (#1359).
+- Compilation problem in some scenarios where a module with state variables is
+  imported by different paths with different namespaces (#1363)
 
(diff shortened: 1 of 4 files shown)


bugarela:
Ok, even though I fixed the scenario in the compiler, it still won't work in apalache. In order to make it work in apalache, we have to fix this in the flattener, which seems like the right thing to do now that I think about it.

I'll close this for now, and open an issue for the flattening problem instead.

shonfeder (approved):

PR #1365, opened 2024-02-08
Support producing multiple traces from runs

### LINKED ISSUE #1360: Add the option to specify a number of (counter)examples to generate
### Discussed in https://github.com/informalsystems/quint/discussions/1138

<div type='discussions-op-text'>

<sup>Originally posted by **ivan-gavran** September  1, 2023</sup>
I would like to start a discussion on a feature that would enable specifying a number of (different) counterexamples to be generated for `quint run` and `quint verify`.

The motivation comes from the usage of Quint in testing: with a specified model, I am interested in generating a (large) number of traces satisfying a property (given as a negated invariant).
I believe the proposed feature could already be achieved for `quint verify` by passing the Apalache parameter `--max-error`.  I don't see how to accomplish the same for `quint run` with current options.

A more general view of this feature request would be to make sampling a first-class citizen of Quint (whereas now it can be accomplished by repeatedly looking for a counterexample). This functionality was one of the features of (now not actively developed) [Modelator](https://github.com/informalsystems/modelator). The question is whether such feature belongs to Quint, or rather to wrappers of Quint (such as Modelator).

</div>

shonfeder:
EDIT by Gabriela: see my comment for updated context

Previous PR description:
--------

Closes #1360

This is a quick workaround to meet the need for this functionality, which has been
voiced in the last few days.

```
quint run --out-itf=out-itf-example.itf.json --n-traces=5 ../examples/solidity/Coin/coin.qnt
```

Here's an example of how to use the new flag:

```
quint run --out-itf=out-itf.itf.json --n-traces=5 ../examples/solidity/Coin/coin.qnt
```

Then there should be 5 traces in a json array in the output file.

-------

- [ ] Update other sub commands to also produce an array of itf traces, rather than a single trace
- [ ] Check with users to be sure this API is suitable to their needs
- [ ] Account for duplicate traces, and ensure traces are unique


bugarela:
Hi Shon! I have a strong need for this right now, so I'll re-open and wrap it up. I hope this is ok, sorry it took me so long.

shonfeder:
No worries at all! I only closed it to prevent leaving unfinished WIP in open PRs. Glad it is here if you need :)

bugarela:
I ended up changing this significantly because, as @ivan-gavran brought to my attention (https://github.com/informalsystems/quint/issues/1448), producing arrays of objects in the ITF file is not compliant with the format, nor will it work with ITF trace viewer and the rust deserializer.

So instead, I'm now producing as many `.itf.json` files as `--n-traces`, adding an index information to the file name.

I also managed to do this "the proper way" now that I got more onboarded into the simulator. Which means we are not running the whole thing multiple times and this should be much more efficient than calling the CLI n times.

There's also one behavior to pay attention to: if `--n-traces=N` is supplied and we find 1 violation, we don't stop there. We continue to look for more counterexamples until we find `N` of them (or `--max-samples` is reached).

quint/src/runtime/trace.ts
```diff
@@ -336,23 +344,28 @@ class TraceRecorderImpl implements TraceRecorder {
...
-
-    if (override) {
-      this.bestTrace = bottom
-      this.bestTraceSeed = this.runSeed
-      this.bestTrace.result = outcome
-      this.bestTrace.args = trace
-    }
+    return this.traces.sort((a, b) => {
+      // Prefer short traces for error, and longer traces for non error.
+      // Therefore, trace a is better than trace b if
```
p-offtermatt, line 349: comment doesn't describe the behaviour correctly afaict

if aOk & bNotOk, then b is better (because errors are preferred), but the comment says "trace a is better than trace b if {}, {}, or b has an error"

p-offtermatt (approved):
LGTM, nice work.
One concern about a comment. afaict the logic there is correct and just the comment is misleading, but the logic should also be double-checked just in case

quint/src/runtime/trace.ts
bugarela: You're right, thanks! Fixed it.

(I thought I had merged this already, completely forgot that this fix was pending, sorry for the delay!)

PR #1370, opened 2024-02-14
Correctly constrain parameter types based on annotations

### LINKED ISSUE #1177: Strange error location given type annotation
Here is my spec: https://github.com/cosmos/interchain-security/blob/4f77d68b9ad06525aa7241f392e6754de364f90c/tests/difference/core/quint_model/ccv.qnt#L357

in v0.9.0 of the VSCode quint extension, I get red squigglies in line 357, under currentState.providerState.
The error message reads
```
Incompatible tails for rows with disjoint fields: {} and { | tail_t491 }
Trying to unify { consumerStatus: (Chain -> str), chainState: ChainState, outstandingPacketsToConsumer: (Chain -> List[VSCPacket]), receivedMaturations: Set[VSCMaturedPacket], sentVSCPackets: (Chain -> Set[VSCPacket]), vscSendTimestamps: ((Chain, VSCId) -> Timestamp), providerValidatorSetChangedInThisBlock: bool, runningVscId: int } and { providerState: t491 | tail_t491 }
```

I find the error location strange - the problem is on line 354, changing currentState to currentState.providerState fixes the problem, see this: https://github.com/cosmos/interchain-security/blob/fa233d655ff0a86073f47bdebe7fce5054b7c6b7/tests/difference/core/quint_model/ccv.qnt#L354

It feels like the red underline should be under currentState, or under isCurrentlyConsumer, but definitely on L354. Is this a bug?

shonfeder:
Fixes #1177

This would also open the way to supporting inline type annotations, like

```ocaml
(x : int) + 1
```

However we discussed an alternative fix for #1177 that may be less invasive, so
I'll see if I can prep a PR for that tomorrow.


shonfeder:
Closed in favor of #1371

Issue #923, opened 2023-06-02
Operator type annotations are ambiguous


shonfeder:
If I have 

```
   def condition(f: int => bool): bool = {
     state.filter(f).size() > 0
   }
```

And decide I want to remove the type annotation on `f` and let that be inferred:

```
   def condition(f): bool = {
     state.filter(f).size() > 0
   }
``

I now have an ill-typed spec:

```
static analysis error: error: [QNT000] Couldn't unify oper and bool
Trying to unify ((int) => bool) => bool and bool

   def condition(f): bool = {
   ^^^^^^^^^^^^^^^^^^^^^^^^^^
     state.filter(f).size() > 0
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   }
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I don't think removing a type annotation in this way should change the meaning of the annotations.

**Comment by konnov** [comment]
I am surprised it is syntax correct. I would expect the parser to report a syntax error in this case.

**Comment by bugarela** [comment]
I don't think it's a parser bug. The parser understands `bool` to be the ML-like type signature, and therefore, there is a type error. We currently do not support partially annotated defs (i.e. `def(a: int, b): int` is not supported). Perhaps we could, but then we would have to come up with a syntax to disambiguate this from the ML-like signatures (as in Shon's example)

**Comment by shonfeder** [comment]
I suggest we drop support this ambiguous, kind of confused, form of annotation entirely. If we want to be able to succinct signatures for operators, rather than interleaving it with the operator parameters, we can introduce a dedicated, unambiguous signature for that. And if we ever introduce the ability to define bindings like

```
def f = (a, ..., c) => foo
```

Then the current annotation approach will work for annotating operator signatures as is.

PR #1371, opened 2024-02-16
Constrain operator parameters by their type annotations

### LINKED ISSUE #1177: Strange error location given type annotation

shonfeder:
Closes #1177

## Reviewing

The PR is smaller than it looks! Almost everything here is either generated by the parser generator or minor cleanup. Review by commit could be helpful.

## The change

The new behavior is easiest to illustrate with reference to the operator
definition used in the regression test added here:

``` scala
pure def foo(s: str): int = {
  val x = 1.in(s) // We SHOULD identify an error here, since s is annotated as a str
  val y = s + 1   // and NOT identify an error here, incorrectly expecting s to be a set
  y
}
```

Prior to this fix, typechecking locates the error at `val y`, because it is not
accounting for the constraint `s : str` before it begins solving constraints in
the body of the declaration:

```
error: [QNT000] Couldn't unify int and set
Trying to unify int and Set[int]
Trying to unify (int, int) => int and (Set[int], int) => _t2

4:         val y = s + 1   // and NOT identify an error here, incorrectly expecting s to be a set
                   ^^^^^

error: typechecking failed
```

With this fix, we now correctly account for the constraint on `s`, so we catch
the misuse at `val x`:

```
error: [QNT000] Couldn't unify set and str
Trying to unify Set[int] and str
Trying to unify (_t0, Set[_t0]) => bool and (int, str) => _t1

3:         val x = 1.in(s) // We SHOULD identify an error here, since s is annotated as a str
                   ^^^^^^^

error: typechecking failed
```

The fix is achieved primarily via a small improvement to the IR (which does not
impact the integration with Apalache): we add an optional `typeAnnotation` field
to `QuintLambdaParam`s. This is a natural fix, since we are already parsing
these out together. 

Along the way there was some minor cleanup and preparation for
https://github.com/informalsystems/quint/issues/923  

This change will also make it very easy to support selective annotation of parameters or return values, and to support annotating params of lambdas.


quint/src/parsing/ToIrListener.ts
```diff
@@ -85,8 +86,6 @@ export class ToIrListener implements QuintListener {
   protected exprStack: QuintEx[] = []
   // the stack of parameter lists
   protected paramStack: QuintLambdaParameter[] = []
-  // stack of names used as parameters and assumptions
-  protected identOrHoleStack: string[] = []
```
shonfeder, line 89: We don't need this stack, because these are just string values we can grab directly.

quint/src/parsing/ToIrListener.ts
```diff
@@ -220,89 +219,79 @@ export class ToIrListener implements QuintListener {
...
     const name = ctx.normalCallName().text
-    const [params, typeTag] = this.processOpDefParams(ctx)
-    // get the definition body
+
+    const params = ctx._annotOperParam
+      .map(_ => popOrFail(this.paramStack, 'violated grammar of annotated AnnotatedOperDef'))
+      .reverse()
+    const res = this.popType().unwrap(() => 'violated grammar of annotated params return type')
+    const args = params.map(p => {
+      assert(isAnnotatedDef(p), 'violated grammar of annotated param type')
```
shonfeder, line 230: These unwraps and asserts are just like the ones in all the `exitType*` methods: the grammar will not bring us to visit this context without these pre-requisites unless we've messed up something badly. It should have no impact on incremental parsing, unless we have a bigger problem that we want to know about.

quint/test/ir/IRFinder.test.ts
```diff
@@ -64,12 +64,15 @@ describe('findParameterWithId', () => {
   const modules = [buildModuleWithDecls(['pure def x(a: int): int = a'])]
 
   it('finds definition for existing id', () => {
-    const def = findParameterWithId(modules, 1n)
-
+    const def = findParameterWithId(modules, 2n)
```
shonfeder, line 67: The number is shifted because the type annotation in the param is now the first leaf we hit while parsing this example.

quint/src/parsing/ToIrListener.ts
```diff
@@ -220,89 +219,79 @@ export class ToIrListener implements QuintListener {
...
 
   // translate a top-level or nested operator definition
-  exitOperDef(ctx: p.OperDefContext) {
+  exitAnnotatedOperDef(ctx: p.AnnotatedOperDefContext) {
     const name = ctx.normalCallName().text
-    const [params, typeTag] = this.processOpDefParams(ctx)
-    // get the definition body
+
+    const params = ctx._annotOperParam
+      .map(_ => popOrFail(this.paramStack, 'violated grammar of annotated AnnotatedOperDef'))
```
bugarela, line 226: `popOrFail` already includes a "violated grammar" prefix, we can remove it from here for clarity
```suggestion
      .map(_ => popOrFail(this.paramStack, 'annotated AnnotatedOperDef'))
```

bugarela (approved):
Looks amazing :sparkles:. I'm so happy you found such a simple fix here, really worth the time invested!

shonfeder:
Thanks for the review and the fix!

PR #1380, opened 2024-02-27
Remove node-fetch dep and put upper bound in node versions

### LINKED ISSUE #1343: Do something about `punycode` deprecation warning
From Node 21, one of our indirect dependencies (`punycode`) is deprecated. We should deal with that.

This is the warning that is shown when someone calls our CLI with node 21:
```
(node:2145150) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
```

shonfeder:
Closes #1343

Two fixes to dependencies are added here:

1. Remove the dependency on node-fetch: this was causing a deprecation warning
  (raised to an error on the latest node) and is no longer needed need now that
  our minimum node version supported is 18.
2. Put upper bound on supported node versions: this bounds supported node
  versions at v20, which is the latest major version to have an LTS release.
  
(1) solves the root cause of #1343 and (2) reduces the chance we'll get but
reports for problems due to unstable versions of node.

### DIFF
--- a/quint/src/quintVerifier.ts
+++ b/quint/src/quintVerifier.ts
@@ -21,7 +21,6 @@ import fs from 'fs'
 import os from 'os'
 // TODO: used by GitHub api approach: https://github.com/informalsystems/quint/issues/1124
 // import semver from 'semver'
-import fetch from 'node-fetch'
 import { pipeline } from 'stream/promises'
 import child_process from 'child_process'
 import * as tar from 'tar'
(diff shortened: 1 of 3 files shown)


quint/src/quintVerifier.ts
```diff
@@ -277,7 +276,7 @@ function downloadAndUnpackApalache(): Promise<VerifyResult<null>> {
   return fetch(url)
     .then(
       // unpack response body
-      res => pipeline(res.body, tar.extract({ cwd: apalacheDistDir(), strict: true })),
+      res => pipeline(res.body!, tar.extract({ cwd: apalacheDistDir(), strict: true })),
```
shonfeder, line 279: I don't think we can realistically get an `undefined` body without instead ending up on in the `error` handler. So the `!` should be safe.

bugarela (approved):
LGTM! Thanks for addressing this :heart:

PR #1381, opened 2024-02-28
WIP Typechecking polymorphic types

shonfeder:
Opening a draft for review, as discussed with @bugarela. 

Still in a very rough state, and I think I may need to rewind at least from 7ec67 too, perhaps trying with a different strategy.


quint/src/parsing/parseErrors.ts
```diff
@@ -25,3 +25,19 @@ export function tooManySpreadsError(id: bigint): QuintError {
     data: {},
   }
 }
+
+export function undeclaredTypeParamsError(id: bigint, unboundTypeVariables: string[]): QuintError {
+  return {
+    code: 'QNT014',
+    message: `the type variables ${unboundTypeVariables.join(', ')} are unbound.
```
bugarela, line 32: Nice error message! Two thoughts for the future:
1. We should probably have a "details" or "explanation" field in the errors to include this second part of this error. I'd also like that for details of type unification errors.
2. This seems like a case where we'll be able to include a `fix` to the error data, and have LSP provide it as a code action, like we do for mode errors.

quint/src/types/constraintGenerator.ts
```diff
@@ -351,48 +372,88 @@ export class ConstraintGeneratorVisitor implements IRVisitor {
...
       const signature = signatureFunction(arity)
       return right(this.newInstance(signature))
     } else {
-      const def = this.table.get(nameId)
+      return this.typeForId(id).mapLeft(
+        (child: ErrorTree): ErrorTree => ({
+          location: `looking up type for ${name}`,
+          children: [child],
+        })
+      )
```
bugarela, line 407: We have a utility for this

```suggestion
      return this.typeForId(id).mapLeft(e => buildErrrorTree(`looking up type for ${name}`, e)
```

shonfeder:
Closing this in favor of #1382 and followups.

PR #1385, opened 2024-03-01
Ensure transformations can't mutate the original object 

shonfeder:
There was some mutation happening via the transformer, which was
wreaking havoc on attempts to complete #1298.

Apparently it was also somehow leading to incorrect printing of sum
type variants. I haven't had time to investigate why, but deep cloning
the original object must be safer and it is fixing a blocking problem in
ongoing work in #1298.

Since the cases in which this causes problems are quite subtle, I havent
been able to figure out how to test this directly.

**NOTE** that this is on top of #1382, and that needs to be merged in first
(and this then rebased). Only [31d5654ee777c5dc21fb234b6e9098b55878d367](https://github.com/informalsystems/quint/commit/462dcfa6eaaa02dd71f0e11c1e85506f059eb853) is new here.

### DIFF
--- a/quint/package-lock.json
+++ b/quint/package-lock.json
@@ -15,6 +15,7 @@
         "@sweet-monads/either": "~3.2.0",
         "@sweet-monads/maybe": "~3.2.0",
         "@types/line-column": "^1.0.0",
+        "@types/lodash.clonedeep": "4.5.0",
         "@types/seedrandom": "^3.0.4",
         "antlr4ts": "^0.5.0-alpha.4",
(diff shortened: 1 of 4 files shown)


bugarela (approved):

bugarela:
I wrote a comment on the commit that is now showing up here at the PR level: https://github.com/informalsystems/quint/commit/462dcfa6eaaa02dd71f0e11c1e85506f059eb853#r139252697

shonfeder:
Thanks for the review here too!

PR #1394, opened 2024-03-05
Allow shadowing

### LINKED ISSUE #1187: Lack of shadowing is confusing
Take this spec:

```
module Core {
    pure val foo =
        val boo = 3
        boo    
}
```

Loading it into repl, I cannot access `boo` (this is what I would expect)
but I also cannot assign a new variable called `boo` (this is not expected)

```
quint -r test.qnt::Core       
Quint REPL 0.14.3
>>> boo
syntax error: error: [QNT404] Name 'boo' not found
boo
^^^


>>> val boo = 3
syntax error: error: [QNT101] Conflicting definitions found for name 'boo' in module 'Core'
val boo = 3
^^^^^^^^^^^

syntax error: error: [QNT101] Conflicting definitions found for name 'boo' in module 'Core'
        val boo = 3
```

This seems related to https://github.com/informalsystems/quint/issues/802, but seems to be a different issue, since it has nothing to do with instancing modules.

Version Info: This is happening after `make`ing quint from the v0.14.3 release branch

bugarela:
Hello :octocat: 

We've been discussing whether to allow shadowing for a while, and this finally does it! Closes #1187 

This is quite easy because we already have a [unshadower](https://github.com/informalsystems/quint/blob/gabriela%2Fallow-shadowing/quint/src/names/unshadower.ts), which we had to write to unshadow names between modules.

So the change is only to our definitions collector, so it is able to collect a stack of definitions by name, instead of a single one. This is not super clean because or imports are quite complicated - we should look for a way of simplifying that in the future.


quint/src/ir/IRVisitor.ts
```diff
@@ -444,6 +444,9 @@ export function walkExpression(visitor: IRVisitor, expr: ir.QuintEx): void {
       break
     }
     case 'lambda':
+      if (visitor.definitionDepth !== undefined) {
```
shonfeder, line 447: Why do we allow this to be undefined?

quint/src/names/base.ts
```diff
@@ -49,7 +50,7 @@ export type LookupDefinition = (QuintDef | ({ kind: 'param' } & QuintLambdaParam
 /**
  * A module's definitions, indexed by name.
  */
-export type DefinitionsByName = Map<string, LookupDefinition & { hidden?: boolean }>
+export type DefinitionsByName = Map<string, (LookupDefinition & { hidden?: boolean })[]>
```
shonfeder, line 53: Looks like `LookupDefinition` already has `hidden?`, so this type can be cleaner

```suggestion
export type DefinitionsByName = Map<string, LookupDefinition[]>
```

See https://github.com/informalsystems/quint/blob/f51cf1aa721caf907f03e45b53f44ccf01e25972/quint/src/names/base.ts#L28-L32

quint/src/names/collector.ts
```diff
@@ -142,13 +144,14 @@ export class NameCollector implements IRVisitor {
       }
 
       // Update the definition to point to the expression being overriden
-      instanceTable.set(param.name, { ...constDef, id: ex.id })
+      constDef.id = ex.id
+      constDef.hidden = false
```
shonfeder, line 148: So we don't have to update the `instanceTable` at all now? Or is that happening somewhere else now?

shonfeder (approved):
LGTM! I think this will be a very nice UX improvement.

quint/src/ir/IRVisitor.ts
bugarela: Because some visitors don't need it, therefore it is an optional attribute. If it was a mandatory attribute, those visitors would have to declare and initialize it.

quint/src/names/collector.ts
bugarela: I'm mutating the values themselves.
Before, we updated the whole value of the map (which was a single def). Now, the values are arrays, so it would be tricky to update the value without accidentally mutating the array. Therefore, I decided to `cloneDeep` the entire map, to be safe. And, since it is safe to mutate things, we can just mutate the definition directly.

quint/src/names/base.ts
bugarela: Thanks! Much better

PR #1397, opened 2024-03-07
Move Apalache client into its own module

### LINKED ISSUE #1396: Factor gRPC interactin out of `quintVerifier`
We will need use the apalache server in the command that allows transforming a quint spec int TLA for #1309. But currently all the code to enable interaction with the apalache server is in the `quintVerifier` module.

To prepare for that work, we should factor our the server interaction code into its own modules.

shonfeder:
This is almost entirely a mechanical change. We just move the Apalache
server manager and client into its own module, and rename a few types to
reflect the generality of the functionality.

This leaves `quintVerifier.ts` super slim, but all of the verify logic
(excluding the IO operations) from the `cliCommand` module should be
moved there eventually.

Closes #1396


quint/src/apalache.ts
```diff
@@ -0,0 +1,430 @@
...
+}
+
+// The structure used to report errors
+type ApalacheError = {
+  explanation: string
+  errors: ErrorMessage[]
+  traces?: ItfTrace[]
+}
+
+export type ApalacheResult<T> = Either<ApalacheError, T>
```
shonfeder, line 61: The only change here aside from moving stuff was to rename these types (that used to have `Verify` instead of `Apalache`).

bugarela (approved):

Issue #359, opened 2022-11-24
Add `compile` sub command


konnov:
Our current choice of `tntc` commands is dictated by the implementation details, rather than by usability.

Instead of having `parse`, `typecheck`, and maybe something else, I was thinking of introducing the `compile` command:

 - It integrates the full compilation cycle: the parser, the type checker, the effects checker, and (possibly) compilation to JS or TLA+
 - The choice of the compilation target may be give via the `--target` option. The current candidates are: `js` and `json`.
 - For fine-grained control, it introduces options, telling the compiler where to stop, e.g. after parsing, after type checking, etc.

@bugarela, @shonfeder, wdyt?

shonfeder:
Copied over from #1278:

I think I agree with the functionality I'm not sure if I agree on the preferred interface.

The main thing I need is output of the stage that includes flattening, and a compile sub command seems like the right place to put that.

As to whether we should merge the parse and typecheck commands into compile, I'm not so sure. I'm not opposed, I just don't see a problem with having distinct subcommands here or what the benefit is of moving it into flags under a single subcommand. I guess our current interface seems a bit more like cargo than go, in that cargo has both build and check?

shonfeder:
Important additional context and motivation from #1278:

parse and typecheck can be used to check an input's validity for syntax and typechecking, respectively, and to produce the intermediate stage of IR resulting from those processes. However, run, test, verify do not operate on the result of typechecking: instead they all require post-processing that includes forms of flattening and inlining.

This stage of the process is not exposed through the CLI currently, even tho it would be helpful for debugging and is probably what other tooling would want to consume.

I propose exposing this stage of the process through a compile subcommand, which will also become the recommended way for users to perform all static analysis and processing to a spec before passing it off to other tools.

konnov:
Makes sense to me!

PR #1399, opened 2024-03-12
Add `compile` subcommand

### LINKED ISSUE #1309: Support integration with TLC
This was something @konnov brought up last week.

It would be great if we can come with some nice UX enabling users of TLC to check quint specs.

To move forward here, we need to do a good bit of discovery work, including:

- How much interest is there in this feature set? (Please add a :+1: to this issue if you need this.)
- What is already possible with our current tool chain?
- What is the minimum amount of work needed for a workable solution?
- What is the best possible UX we could engineer given the relationship between TLA+ and Quint?

shonfeder:
Closes #1309, #359

Both issues include ideas for followup work (e.g., compiling to a
`js` or UX improvements for a TLC integration). But this covers the
core of both.

The main target of this changeset is #1309, but #359 was accomplished with a few
extra lines along the way.

Note that one of the test cases is actually showing undesirable behavior:
flattening is still too aggressive, in that it is clobering main modules that
just instantiate another module. When values are referenced in the importing
module, they are sent along to Apalache as expected.

I'm not sure how we want to fix this.


shonfeder:
I still need to update the changelog and the documentation here.

quint/apalache-tests.md
```diff
@@ -219,3 +219,54 @@ An example execution:
...
+### Test that we can compile a module with imports and instances to TLA+
+
+
+<!-- !test in can convert clockSync3.qnt to TLA+ -->
+```
+quint compile --target tlaplus  ../examples/classic/distributed/ClockSync/clockSync3.qnt | head
+```
+
+
+TODO: This is an incorrect result, we are removing all "unused" declarations
```
bugarela, line 263: I can think of two solutions here, considering our current flattening system:
1. introduce `q::init` and `q::next`, as we do with run and verify, and that'd include all dependencies of those, resulting in a valid state machine module
2. Try to force keep all the names in the main module (clockSync3). Shoudn't be trivial, but it should be possible.

quint/src/apalache.ts
```diff
@@ -136,8 +146,9 @@ type RunRequest = { cmd: string; config: string }
 
 type RunResponse =
   | { result: 'failure'; failure: { errorType: string; data: string } }
-  // The success data also includes the parsed module, but we don't need it
-  | { result: 'success' }
+  // The success data also includes the parsed module, either as JSON
+  // representing the Apalache IR, ir as a TLA string (if the `TLA` command is used)
```
bugarela, line 150:
```suggestion
  // representing the Apalache IR, or as a TLA string (if the `TLA` command is used)
```

quint/src/apalache.ts
```diff
@@ -136,8 +146,9 @@ type RunRequest = { cmd: string; config: string }
...
+  | { result: 'success'; success: string }
```
bugarela, line 151: Should this be something more meaningful?
```suggestion
  | { result: 'success'; parsed_module: string }
```

quint/src/cliCommands.ts
```diff
@@ -604,6 +606,67 @@ export async function runSimulator(prev: TypecheckedStage): Promise<CLIProcedure
...
 
+/** compile a spec and output in the specified `typechecked.args.target`
+ *
+ * The output produced by this command includes the parsed, typechecked, and
+ * flattened data.
+ *
+ * @param typechecked The result of a preceding typechecking stage
+ */
+// TODO This should share most of the logic in `verifySpec`, but the logic in
+// that command would need some significant reworking to make it composable
```
bugarela, line 617: Perhaps, whatever we decide to do on https://github.com/informalsystems/quint/pull/1399/files#r1521344080 might help with this as well.

bugarela (approved):
Looks great, the CLI turned out really intuitive! Let's figure out what to do for the flattening problem, but I'll leave it up to you whether you want to merge this and address that in a followup, or leave this open to iterate here.

quint/apalache-tests.md
shonfeder: I don't know that (1) makes sense for compiling to TLA+. One may want different operators to be available instead, I'd think? I also just think it seems like kind of bad behavior to have specifications changed when you are trying to compile it to a target.

In principle, my current sense is that something like (2) is called for l, and that the current behavior of flattening should be considered semi broken. Afaiu, The idea of unqualified instantiation of a module is that it extends the surrounding module. Flattening should respect this, and but require us to do ad hoc additions to prevent flattening from wiping out all of the declarations we meant to add. WDYT?

quint/src/apalache.ts
bugarela: I see. I didn't realize it was from the proto module. I guess it does make sense to have it more general there. Don't worry about it!

quint/src/cliCommands.ts
shonfeder: Yeah, I think that is exactly right! Most of complication in `verifySpec` is only required due to the flattening problem.

quint/apalache-tests.md
shonfeder: The realization that convinced me that the current behavior of flattening is incorrect when it comes to unqualified instantiation was that we are stripping out state variables which, if you do an unqualified instantiation with the current semantics, you must have meant to be included in the instantiating module.

Let's definitively address this in a followup either way tho.

bugarela (approved):

PR #1407, opened 2024-03-14
Add q::* operators to compiled targets

### LINKED ISSUE #1402: `compile` needs to add special treatment for intended `init`, `step`, and `inv`
This is required to support compilation to apalache

@bugarela pointed out that

- In TLA+, the init definition is a predicate, not an action. So we need to replace assignemtns with equalities (so we need to know what is init when getting it parsed by apalache into TLA+

- TLC expects a .cfg file. If we have a fixed name for `INIT`, `NEXT`, and `INVARIANT`, this will allow us to reuse the same cfg file for running in TLC.

shonfeder:
Closes #1402

This also shares the flattening and extension logic between the
`compile` command and the `verifySpec` command.

Note that this presupposes #1406, and it will need to be rebased after that is merged in.

### DIFF
--- a/doc/quint.md
+++ b/doc/quint.md
@@ -73,8 +73,7 @@ especially useful for debugging complex specifications.
 $  quint compile --help
 quint compile <input>
 
-compile a Quint specification into the target, the output is written to
-stdout
+compile a Quint specification into the target, the output is written to stdout
 
 Options:
   --help       Show help                                               [boolean]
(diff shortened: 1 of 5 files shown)


doc/quint.md
```diff
@@ -91,6 +94,16 @@ Given a quint specification as input, this command parses, resolves imports,
...
 the needed definitions.
 
+The main module is determined as follows: If a module name is specified by
+`--main`, that takes precedence. Otherwise, if there is only one module in the
+input file, that is the main module. Otherwise, the module with the same name as
+the file is taken to be the main module.
+
+The main module must specify a state machine. This means it must either define
+actions named `init` and `step`, specifying the initial state and the
+transition action respectivefly, or suitable actions defined in the main module
```
bugarela, line 104:
```suggestion
transition action respectively, or suitable actions defined in the main module
```

doc/quint.md
```diff
@@ -91,6 +94,16 @@ Given a quint specification as input, this command parses, resolves imports,
...
+must be indicated using the `--init` and `--step` actions.
```
bugarela, line 105:
```suggestion
must be indicated using the `--init` and `--step` options.
```

bugarela (approved):

Issue #378, opened 2022-11-29
Make sure that top-level `nondet` fails with a parse error


konnov:
To this end, I have added a test `testFixture/_1015noToplevelNondet.tnt`. However, it is failing, perhaps, due to the recent change in `Either`.

To enable a parse error

1. Uncomment:

https://github.com/informalsystems/tnt/blob/146ca81569e077b8efefb95758288022f0f225c0/tntc/src/generated/Tnt.g4#L25-L29

2. Generate the new parser via `npm run antlr`.

3. Uncomment the test in:

https://github.com/informalsystems/tnt/blob/146ca81569e077b8efefb95758288022f0f225c0/tntc/test/tntParserFrontend.test.ts#L120-L125

4. Produce the expected result for the fixture:

```sh
tntc parse --out testFixture/_1015noToplevelNondet.json --source-map testFixture/_1015noToplevelNondet.map.json testFixture/_1015noToplevelNondet.tnt
```

5. Run the tests and observe the strange message:

```sh
...
  4) parse errors
       error on top-level nondet:
     AssertionError: expected JSON results to be equal: expected { Object (type, value) } to deeply equal [ Array(1) ]
      at parseAndCompare (test/tntParserFrontend.test.ts:55:10)
      at Context.<anonymous> (test/tntParserFrontend.test.ts:121:5)
      at processImmediate (node:internal/timers:471:21)
```

Issue #1422, opened 2024-04-08
Improve our tests that use fixtures


bugarela:
Ideas:
1. Format test fixture json file as suggested by @p-offtermatt in https://github.com/informalsystems/quint/pull/1421#pullrequestreview-1985582480, this would make the diffs much easier to read (currently impossible to read)
2. Use a proper snapshot lib to handle these.

bugarela:
We could take advantage of `jd`, which shows formatted json difference:
```
git show main:quint/testFixture/_1016nonConstOverride.json | jd testFixture/_1016nonConstOverride.json
@ ["table","8","id"]
- 6
+ 2
```

But we should first check if there are any snapshot tools - I want to avoid having to maintain the tool in house, even if its simple.

PR #1431, opened 2024-05-06
Checks for misuse of `nondet` and `oneOf`

bugarela:
Hello :octocat: 

:warning: Big diff only due to generated grammar files. Actual diff should be reasonable.

Closes https://github.com/informalsystems/quint/issues/378 (it actually fails with an "effect" error, which I think it makes more sense than a parse error, but this has been greatly discussed in the past).

It also solves the most important parts of https://github.com/informalsystems/quint/issues/381, but I still have to see if there's something greater to be accomplished if we make proper effects for non-determinism.

This PR introduces some checks for how `nondet` and `oneOf` can be used. The usage is quite restrictive because we need to be able to convert this to TLA+ for model checking, and the current syntax makes it seem like you can use non determinism everywhere. Hopefully, the IDE/cli errors will help people understand that these are special keywords that can't be used everywhere. Also, since the simulator allows too much, you'd only notice something is wrong when running verify or compile to TLA+.

One of the cases in particular is when `nondet` definitions are top-level, which is not allowed. This was already being caught, but by the parser, and it was really hard to understand the issue by the parsing error. So I changed the parser to normally parse `nondet` as any other qualifier, and then properly report the error later on.

Test fixtures were updated because I'm now including the definitions themselves in the lookup table, so the tables in all fixtures were updated. We have issue https://github.com/informalsystems/quint/issues/1422 to address how hard it is to read the diffs - for now: sorry!


quint/src/quintError.ts
```diff
@@ -67,6 +67,14 @@ export type ErrorCode =
   | 'QNT201'
   /* QNT202: Multiple updates of the same variable */
   | 'QNT202'
+  /* QNT203: 'oneOf' must be used inside a nondet definition */
+  | 'QNT203'
+  /* QNT204: 'oneOf' must be the outtermost expression in a nondet definition */
```
p-offtermatt, line 72:
```suggestion
  /* QNT204: 'oneOf' must be the outermost expression in a nondet definition */
```

quint/src/effects/NondetChecker.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (expr.opdef.qualifier !== 'nondet') {
+      return
+    }
+
+    // body of nondet must be an application of oneOf
+    const body = expr.opdef.expr
+    if (body.kind !== 'app' || body.opcode !== 'oneOf') {
+      this.errors.push({
+        code: 'QNT204',
+        message: `'oneOf' must be the outtermost expression in a nondet definition`,
```
p-offtermatt, line 88:
```suggestion
        message: `'oneOf' must be the outermost expression in a nondet definition`,
```

quint/src/effects/NondetChecker.ts
```diff
@@ -0,0 +1,105 @@
...
+        data: {},
+      })
+    }
+
+    // if the opdef is nondet, the return type must be bool
+    const expressionType = this.types.get(expr.expr.id)
+    if (expressionType?.type.kind !== 'bool') {
+      this.errors.push({
+        code: 'QNT205',
+        message: `nondet bindings can only be used with boolean expressions`,
```
p-offtermatt, line 99: I don't think I understand this error - nondet bindings can only be used with boolean expressions means the right-hand side of nondet variables can only be boolean?
I don't know if there is something that would make it more obvious what this means
I think it's maybe closer to
```suggestion
        message: `nondet bindings can only be used in expressions returning a boolean`,
```

p-offtermatt: for extra debugging help, could also output the actual type that is returned instead of a Boolean, to make it easier to identify whats happening?

p-offtermatt (approved):
LGTM, just a couple of comments on clarity of errors and some spelling.

PR #1436, opened 2024-05-13
Hotfix: Handle cases where table entries or types are missing

bugarela:
Hello :octocat: 

This is a hotfix for a mistake I made in https://github.com/informalsystems/quint/pull/1431.

I thought I could assume all types were present at that point, but as those checks run in the same phase as the type checker, it can be the case that there are type errors and, therefore, missing types. This was causing the tool and the language server to crash in those scenarios, in their latest versions released after the linked PR.

We strive for accumulating as many errors as we can before giving up analysis and reporting them. So this PR tries to report nondet errors when it can (types and lookup table entries are found), and otherwise just safely skip the check since surely a type error is already in the list of errors to be reported on exit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 ### Fixed
 
-- Fixed a bug introduced in v0.19.3 here the analyzer would crash if there were
+- Fixed a bug introduced in v0.19.3 where the analyzer would crash if there were
   some specific type errors (#1436)
 
 ### Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 ### Removed
 ### Fixed
+
+- Fixed a bug introduced in v0.19.3 here the analyzer would crash if there were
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 ### Removed
 ### Fixed
+
+- Fixed a bug introduced in v0.19.3 here the analyzer would crash if there were
```
p-offtermatt, line 16:
```suggestion
- Fixed a bug introduced in v0.19.3 where the analyzer would crash if there were
```

p-offtermatt (approved):
Nit: Tiny typo in the changelog.

Could you maybe add something to the test to ensure that one of the expected error types is actually returned when running that?
Might not be the right place to test this and that this is tested elsewhere, if so feel free to disregard (I'll approve already to unblock you in that case)

bugarela:
> Could you maybe add something to the test to ensure that one of the expected error types is actually returned when running that? 

I'd need an integration test for that, and it's a very specific scenario. Considering there are many other places that rely on this sort of assumption ("an error should have been reported already"), I instead improved some dosctrings to make that more explicit: "if we return an incomplete map of types/effects, there will be accompanying errors". To properly test this behavior, I think the best would be to use PBT and check that is never the case where there are missing types and no errors.

PR #1440, opened 2024-05-17
Convenience Spells

ivan-gavran:
Adding 3  convenience spells:
 - `setAdd`, to add a single element to a set
 - `listSum`, sum of all elements in a list
 - `listContains`, a boolean saying whether a list contains an element

### DIFF
--- a/examples/spells/basicSpells.qnt
+++ b/examples/spells/basicSpells.qnt
@@ -75,6 +75,16 @@ module basicSpells {
     assert(Set() == Set().setRemove(3)),
   }
 
+  pure def setAdd(__set: Set[a], __elem: a): Set[a] = {
+    __set.union(Set(__elem))
+  }
+
+  run setAddTest = all{
+    assert(Set(2, 3, 4) == Set(2, 4).setAdd(3)),
+    assert(Set(3) == Set().setAdd(3)),
+    assert(Set(2,4) == Set(2,4).setAdd(4)),
+  }
+
   /// Test whether a key is present in a map
   ///
   /// - @param __map a map to query
--- a/examples/spells/commonSpells.qnt
+++ b/examples/spells/commonSpells.qnt
@@ -19,6 +19,32 @@ module commonSpells {
   }
 
 
+/// Compute the sum of the values in a list.
+   ///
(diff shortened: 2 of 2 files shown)


examples/spells/commonSpells.qnt
```diff
@@ -19,6 +19,32 @@ module commonSpells {
   }
 
 
+/// Compute the sum of the values in a list.
```
bugarela, line 22:
```suggestion
  /// Compute the sum of the values in a list.
```

examples/spells/basicSpells.qnt
```diff
@@ -75,6 +75,16 @@ module basicSpells {
     assert(Set() == Set().setRemove(3)),
   }
 
+  pure def setAdd(__set: Set[a], __elem: a): Set[a] = {
```
bugarela, line 78: Can you add a doc comment?

examples/spells/commonSpells.qnt
```diff
@@ -19,6 +19,32 @@ module commonSpells {
...
+    __list.foldl(0, ((__sum, __i) => __sum + __i))
+  }
+
+  run listSumTest = all {
+    assert(listSum([]) == 0),
+    assert(listSum([1,1,1,3]) == 6),
+    assert(listSum([-4, 4, 5]) == 5),
+  }
+
+  pure def listContains(__list: List[a], __elem: a): bool =
```
bugarela, line 36: Can you add a doc comment here as well?

examples/spells/basicSpells.qnt
romac: Could this be instead added as a built-in named `add`?

bugarela: Yes, good point! It's actually in the plans to change Quint's type system to allow overloading of names for different types, so we should be able to get rid of all `setRemove`, `mapRemove`, `listRemove` etc and just have `remove` as everyone expects.

romac: > It's actually in the plans to change Quint's type system to allow overloading of names for different types

That would be amazing!

bugarela:
Closing in favor of #1507

Issue #1379, opened 2024-02-26
Automatic generation of test cases for counterexample traces


p-offtermatt:
It would be very useful to have an option to automatically convert a counterexample trace into a test inside your module.

A suggestion for the workflow could be like this:
You run ```quint run .... file.qnt``` and get a counterexample. You then run something like ```quint generate-test```, which automatically generates a test according to the run that just found a violation and puts it into ```file.qnt```

I am not very sure this is feasible: I think a problem is probably how/where to resolve nondeterminism.
If no nondeterminism is resolve, a valid counterexample trace would always just be something like "init.then(step).then(step).then(step).....then(assert(invariant_that_fails))```
Of course, in reality ```step``` probably looks like
```
any {
nondet param = oneOf(..)
nondet param2 = oneOf(..)
action1(param, param2)

nondet param = oneOf(..)
action2(param)
}
```
Then, optimally the test case generator would be smart enough to figure out that the trace can be made deterministic by not simply calling ```step``` over and over, but by calling the actions with the parameters that were chosen in the counterexample.

I think a tricky part would be choosing which level of nondeterminism resolution is appropriate, e.g. somewhere deep in some call, something might be nondeterministic, and resolving that could lead to extremely long tests calling tons of functions. Then one might prefer leaving this deep nondeterminism, and having that resolve via the seed.
I think one could have a parameter governing how deep to resolve nondeterminism, or the crazier solution: try many ways and query an LLM to ask which way it thinks is most appropriate :)

PR #1441, opened 2024-05-20
Automatically produce metadata to be used in MBT tests

bugarela:
Hello :octocat: 

This is already useful on its own, but it is also work towards https://github.com/informalsystems/quint/issues/1379

### Context

For many specs used for Model-Based Testing, we have been creating state variables with the sole purpose of tracking which action was taken, with which parameters, etc. This adds a lot of noise to the spec, and it is unnecessary since Quint itself should be able to produce this information and add it to the trace automatically.

### This PR

I'm introducing a flag for the `run` command called `--mbt`. When this flag is given, the Quint simulator will keep track of two additional variables on the traces it produces:
- `action_taken`: The first action executed by the simulator on each step, reset at every `any` evaluation. That is, if the spec has nested `any` statements, `action_taken` will correspond to the action picked in the innermost `any`.
  - This isn't a really good solution in the scenario where there are nested `any` calls. But  well-written specs don't have those, so I found this heuristic good enough. 
- `nondet_picks`: A record with all `nondet` values that were picked since the last `any` was called (or since the start, if there were no `any` calls in the step).

#### Tech debt?

I don't really like that we are passing a `storeMetadata` flag around so much. I think ideally this control of whether or not to include the metadata in the trace should be done in the trace recorder, but I still have to onboard myself into that code and tbh I'm scared to touch it right now. So I'd rather make a small debt of having these flags to stay safe until I can look at this with more time.

  - Only integration tests because there is no existing unit tests for this level of the code


quint/io-cli-tests.md
```diff
@@ -715,6 +749,102 @@ rm out-itf-example.itf.json
...
+        "null",
+        {
+          "#bigint": "0"
+        }
+      ]
+    ]
+  },
+  "minter": "bob",
+  "nondet_picks": {
+    "amount_189": {
```
bugarela, line 806: For all names in this record, they have this annoying suffix (`_189`) for now due to unshadowing that happens before compilation. I have a fix for this in another branch (PR soon), so this will look like `amount` instead of `amount_189`, unless there are two nested `nondet amount = ...` declarations.

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -984,7 +1031,7 @@ export class CompilerVisitor implements IRVisitor {
     // this function gives us access to the compiled operator later
     let callableRef: () => Maybe<Callable>
 
-    if (lookupEntry === undefined || lookupEntry.kind !== 'param') {
+    if (lookupEntry.kind !== 'param') {
```
bugarela, line 1034: I removed the `lookupEntry === undefined` case because it is impossible (see the last if a few lines above)

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1447,6 +1487,12 @@ export class CompilerVisitor implements IRVisitor {
     // non-deterministically. Instead of modeling non-determinism,
     // we use a random number generator. This may change in the future.
     const lazyCompute = () => {
+      // on `any`, we reset the action taken as the goal is to save the last
+      // action picked in an `any` call
+      this.actionTaken = none()
+      // we also reset nondet picks since they should correspond to the action taken
+      this.nondetPicks = this.emptyNondetPicks()
```
p-offtermatt, line 1494: Not sure about this one. consider this rough example:

```
action step =
any {
   someAction,
   nondet time = oneOf(times)
   timeRelatedAction(time)
}

action timeRelatedAction(time) = 
any {
  AdvanceTime(time),
  RevertTime(time)
}
```

If I get the code correctly, it would reset the fields to None here?

p-offtermatt: Seems I didn't understand the code correctly, because testing locally this is not true and in my example the nondet values are still stored

bugarela: I looked again and you are right, I think my comment there is misleading - I was a bit confused myself yesterday, and this is code I wrote a month ago :grimacing:. I tried removing this reset yesterday and it results in some leftovers between steps.

The thing that makes your example work well is something I missed while preparing the PR: On let-in definitions, as in
```
nondet time = oneOf(times)
timeRelatedAction(time)
```
We first evaluate the body (`timeRelatedAction(time)`) and only then add `time` to `nondetPicks`, and this is why it works. In other words, we construct the `nondetPicks` record while we are going *up* the IR tree in compilation.

This is where it happens:
https://github.com/informalsystems/quint/blob/gabriela%2Fmbt-metadata/quint/src/runtime/impl/compilerImpl.ts#L344. I'll add a comment there with this extra context. Thanks for the careful review :heart:

p-offtermatt: Ah, nice! Thank you!

ivan-gavran:
> @ivan-gavran I added you as a reviewer as well since you have the most context.
> 
> The most important here is to review the behavior, I believe.

Sorry, I missed this one :(

And I am joining the discussion too late, after the PR is already merged, but I'd like to raise one point: the ITF trace adds the info, but does not list `action_taken` or `nondet_picks` under its "vars" field. Thus, they are not showed in the ITF trace viewer.

Would it make sense to add them under "vars"? Or is "vars" used for some other purposes, too, where it would be harmful?

bugarela:
I haven't thought of that! I don't think this is used anywhere else - I should be able to add it and I think it would make sense :smile:

PR #1453, opened 2024-06-03
Use Either and extract code from compiler implementation

bugarela:
Hello :octocat: 

This PR is a refactor I started a while ago and don't want to let rotten, specially since I'll make changes to this code in the next days. The goal is to reduce the size of `compilerImpl.ts`, which currently holds too much responsibility and makes my Emacs laggy :sweat_smile:.

I does two things:
1. Extract code from `compilerImpl.ts` into `base.ts` and `operatorEvaluator.ts`
1. Change some Maybe types into Either. This allows methods to be turned into functions, because with Either, we can have the functions return left(error) when there's an error. Before this change, error reporting was completely stateful, so anything relying on reporting errors had to be inside the big `CompilerVisitor` class.

I really tried not making this bigger than it had to be, so there are some frontiers to those changes. You'll see a bunch of `this.addRuntimeError()` turning into `return left(error)`, but not all of them. We still have plenty of `Maybe` types around, and many more things that need to be extracted.

No behavior should change, except for:
- A fix in random number generation that changed results for the previous seeds
- Some better error reporting


quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1495,25 +1323,26 @@ export class CompilerVisitor implements IRVisitor {
...
+            if (sz === 0n) {
+              return left({ code: 'QNT509', message: `Applied oneOf to an empty set` })
+            }
+            return right(this.rand(sz))
+          } else {
             // An infinite set, pick an integer from the range [-2^255, 2^255).
             // Note that pick on Nat uses the absolute value of the passed integer.
             // TODO: make it a configurable parameter:
             // https://github.com/informalsystems/quint/issues/279
-            .or(just(-(2n ** 255n) + this.rand(2n ** 256n)))
```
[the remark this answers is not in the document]
bugarela: I added an extra call to `rand()` to simulate the old behavior here and tests work with the old seeds, so this confirms that this change was the cause of the difference

quint/src/runtime/impl/base.ts
```diff
@@ -0,0 +1,109 @@
+import { Maybe, just, none } from '@sweet-monads/maybe'
```
bugarela, line 1: These were all extracted from compilerImpl and mostly unchanged. Only new definition is `toMaybe`, which might go away if we get to convert all remaining maybe types into either types.

quint/src/runtime/impl/compilerImpl.ts
```diff
@@ -1122,94 +1040,38 @@ export class CompilerVisitor implements IRVisitor {
...
-    // apply the lambda to a single element of the set
-    const evaluateElem = function (elem: RuntimeValue): Maybe<[RuntimeValue, RuntimeValue]> {
-      // evaluate the predicate against the actual arguments
-      const result = callable.eval([just(elem)])
-      return result.map(result => [result as RuntimeValue, elem])
-    }
-    this.applyFun(sourceId, 1, (iterable: Iterable<RuntimeValue>): Maybe<RuntimeValue> => {
-      return flatMap(iterable, evaluateElem).map(rs => mapResultAndElems(rs))
-    })
+    this.popArgs(2)
```
bugarela, line 1045: starting here, I extracted some methods into functions defined at `operatorEvaluator.ts`. So these methods right now only deal with the stateful part, while the functions are responsible for the (more isolated) logic.

quint/src/runtime/impl/operatorEvaluator.ts
```diff
@@ -0,0 +1,167 @@
...
+          .chain(vs => fun(...vs.map(v => v as RuntimeValue)))
+      } catch (error) {
+        const msg = error instanceof Error ? error.message : 'unknown error'
+        return left({ code: 'QNT501', message: msg, reference: sourceId })
+      }
+    },
+  }
+}
+
+export function applyFold(order: 'fwd' | 'rev', args: Computable[]): Computable {
```
bugarela, line 34: I have re-written this implementation to use typescript's `reduce`/`reduceRight` instead of implementing fold by hand.

quint/test/runtime/compile.test.ts
```diff
@@ -980,7 +973,7 @@ describe('compiling specs to runtime values', () => {
 
       evalRun('run1', input)
         .mapRight(result => assert.fail(`Expected the run to fail, found: ${result}`))
-        .mapLeft(m => assert.equal(m, 'Value of run1 is undefined'))
+        .mapLeft(m => assert.equal(m, "[QNT513] Cannot continue in A.then(B), A evaluates to 'false'"))
```
bugarela, line 976: We now have a proper error message for this.

quint/cli-tests.md
```diff
@@ -327,7 +327,7 @@ fi
 <!-- !test check gradualPonzi - Run progressInv -->
     quint run --invariant=progressInv --main=gradualPonziTest \
       --max-samples=1000 --max-steps=50 \
-      --seed=0x8272b36d6a57f \
+      --seed=0xa7bf730b93981 \
```
bugarela, line 330: For changes in tests, see my comment about the seed change.

CHANGELOG.md
```diff
@@ -12,6 +12,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 ### Removed
 ### Fixed
+
+- Fixed a problem where random numbers where internally produced without being
```
p-offtermatt, line 16:
```suggestion
- Fixed a problem where random numbers were internally produced without being
```
tiny typo :)

p-offtermatt (approved):
Awesome! LGTM, it's a lot of changes but the intention looks great and everything I saw and understood looked good.

Would it make sense to wrap something around Either<QuintError, Foo>? Right now, it seems like you have to pay a lot of attention which is left and which is right, and it seems easy to mess up. I think Maybe looked nicer in that respect, but I understand it has some shortcomings, so maybe introducing smt like Result<Foo> as a shorthand for Either<QuintError, Foo> is a good pattern, then you could have 
```
if(result.isError()) {
 // handle error
}
```
instead of 
```
if(result.isLEft()) {
 // handle error
}
```

bugarela:
> Awesome! LGTM, it's a lot of changes but the intention looks great and everything I saw and understood looked good.
> 
> Would it make sense to wrap something around Either<QuintError, Foo>? Right now, it seems like you have to pay a lot of attention which is left and which is right, and it seems easy to mess up. I think Maybe looked nicer in that respect, but I understand it has some shortcomings, so maybe introducing smt like Result as a shorthand for Either<QuintError, Foo> is a good pattern, then you could have
> 
> ```
> if(result.isError()) {
>  // handle error
> }
> ```
> 
> instead of
> 
> ```
> if(result.isLeft()) {
>  // handle error
> }
> ```

I just tried this, and as I suspected, it is not trivial to do. `isLeft` is a method of a hidden class defined in a library, which uses a typescript type system's feature to ensure the value is or isn't the "left" type after evaluation. I don't think I can really reproduce this without messing with the library itself or re-implementing too much of it in the Quint side. To illustrate it better, this is an example code after changing to my custom `isError`:
```
  if (isError(res)) {
    const errors = [{ code: 'QNT512', message: res.value }] as QuintError[]
    return errSimulationResult(errors)
  } else {
    res.value.eval()
  }
```

and I get this error:
```
src/simulation.ts:132:15 - error TS2339: Property 'eval' does not exist on type 'string | Computable'.
  Property 'eval' does not exist on type 'string'.

132     res.value.eval()
```

That is, even tho in the else branch there is no way that the value is a string (since string is the left element here), the type system cannot know that.

This is the type signature of `isLeft()`, from the library: 
```
isLeft(): this is EitherConstructor<L, R, EitherType.Left>;
```

but `EitherConstructor` is not exposed, so I can't have this same type signature for my `isError`.

PR #1455, opened 2024-06-27
Add command `generate`

konnov:
Hello :octocat:!

I would like to add an experimental contribution, namely, the operator `generate` that mirrors `Apalache!Gen`. This operator should unlock the feature of Apalache that is sometimes used to reason about inductive invariants. This PR:

- [x] adds the operator `generate`
- [x] adds an error for `generate` in the simulator, as it is not clear to me how one would implement that in the simulator
- [x] prepares the ground for translation in Apalache.

### DIFF
--- a/quint/cli-tests.md
+++ b/quint/cli-tests.md
@@ -184,6 +184,11 @@ This example was pointing to Paxos. Now it does not typecheck.
 <!-- !test check coin - Types & Effects-->
     quint typecheck ../examples/solidity/Coin/coin.qnt
 
+### OK on typecheck generate
+
+<!-- !test check generate - Types & Effects -->
+    quint typecheck ../examples/language-features/generate.qnt
+
 ### OK on test SimpleAuction.qnt
 
 <!-- !test check SimpleAuction - Syntax/Types & Effects/Unit tests -->
--- a/quint/src/effects/NondetChecker.ts
+++ b/quint/src/effects/NondetChecker.ts
@@ -92,7 +92,7 @@ export class NondetChecker implements IRVisitor {
     if (body.kind !== 'app' || (body.opcode !== 'oneOf' && body.opcode !== 'generate')) {
       this.errors.push({
         code: 'QNT204',
-        message: `'oneOf' or 'generate' must be the outermost expression in a nondet definition`,
+        message: `'oneOf' and 'generate' must be the outermost expressions in a nondet definition`,
         reference: body.id,
         data: {},
       })
--- a/quint/src/effects/NondetChecker.ts
+++ b/quint/src/effects/NondetChecker.ts
@@ -5,8 +5,8 @@
  * --------------------------------------------------------------------------------- */
 
 /**
- * Checking for the misuse of 'nondet' and 'oneOf'. Necessary to ensure they are
- * compatible with the exists operator from TLA+.
+ * Checking for the misuse of 'nondet', 'oneOf', and 'generate'.
+ * Necessary to ensure they are compatible with the exists operator from TLA+.
  *
  * @author Gabriela Moreira
  *
@@ -32,7 +32,7 @@ export class NondetChecker implements IRVisitor {
   }
 
   /**
-   * Checks declarations for misuse of 'nondet' and 'oneOf'.
+   * Checks declarations for misuse of 'nondet', 'oneOf', and 'generate'.
    *
    * @param types - the types of the declarations
    * @param declarations - the declarations to check
@@ -66,7 +66,7 @@ export class NondetChecker implements IRVisitor {
   }
 
   enterApp(app: QuintApp) {
-    if (app.opcode !== 'oneOf') {
+    if (app.opcode !== 'oneOf' && app.opcode !== 'generate') {
       // nothing to check
       return
     }
@@ -74,7 +74,7 @@ export class NondetChecker implements IRVisitor {
     if (this.lastDefQualifier !== 'nondet') {
       this.errors.push({
         code: 'QNT203',
-        message: `'oneOf' must be used inside a nondet definition`,
+        message: `'${app.opcode}' must be used inside a nondet definition`,
         reference: app.id,
         data: {},
       })
@@ -89,10 +89,10 @@ export class NondetChecker implements IRVisitor {
 
     // body of nondet must be an application of oneOf
     const body = expr.opdef.expr
-    if (body.kind !== 'app' || body.opcode !== 'oneOf') {
+    if (body.kind !== 'app' || (body.opcode !== 'oneOf' && body.opcode !== 'generate')) {
       this.errors.push({
         code: 'QNT204',
-        message: `'oneOf' must be the outermost expression in a nondet definition`,
+        message: `'oneOf' or 'generate' must be the outermost expression in a nondet definition`,
         reference: body.id,
         data: {},
       })
--- a/quint/test/effects/NondetChecker.test.ts
+++ b/quint/test/effects/NondetChecker.test.ts
@@ -61,7 +61,7 @@ describe('checkNondets', () => {
     assert.sameDeepMembers(errors, [
       {
         code: 'QNT204',
-        message: "'oneOf' must be the outermost expression in a nondet definition",
+        message: "'oneOf' and 'generate' must be the outermost expressions in a nondet definition",
         reference: 8n,
         data: {},
       },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 
 ### Added
+
+- Add the operator `generate` to mirror `Apalache!Gen`.
+
 ### Changed
 ### Deprecated
 ### Removed
--- a/examples/language-features/generate.qnt
+++ b/examples/language-features/generate.qnt
@@ -1,0 +1,20 @@
+// demonstrating the use of 'generate'
+module gen {
+  var S: Set[int]
+  var T: Set[{ x: int, y: bool }]
+
+  action init = {
+    nondet iS = generate(10)
+    nondet iT = generate(20)
+    all {
+      S' = iS,
+      T' = iT,
+    }
+  }
+
+  action step =
+    all {
+      S' = S,
+      T' = T,
+    }
+}
--- a/examples/language-features/generate.qnt
+++ b/examples/language-features/generate.qnt
@@ -4,8 +4,8 @@ module gen {
   var T: Set[{ x: int, y: bool }]
 
   action init = {
-    nondet iS = generate(10, Set(Set(0)))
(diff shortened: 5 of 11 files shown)


bugarela:
Hi! Thanks Igor, I'll take a look when I can.

Please ignore the Netlify bot, I'm still not done setting it up

bugarela:
Hey @konnov, I finally took a look at these and I feel like I need more context here. I don't think the name "generate" is very informative, and maybe we should have something more tuned to the particular use case where it is necessary.

A couple of months ago, I introduced an operator called `allListsUpTo` https://github.com/informalsystems/quint/pull/1442 (that only works on the simulator). Do you think that we can approach this by adding an `allSetsUpTo` and mapping those into `Apalache!Gen` calls on the Apalache side? Or did I misunderstood this and it should do something completely different?

konnov:
> Hey @konnov, I finally took a look at these and I feel like I need more context here. I don't think the name "generate" is very informative, and maybe we should have something more tuned to the particular use case where it is necessary.
> 
> A couple of months ago, I introduced an operator called `allListsUpTo` #1442 (that only works on the simulator). Do you think that we can approach this by adding an `allSetsUpTo` and mapping those into `Apalache!Gen` calls on the Apalache side? Or did I misunderstood this and it should do something completely different?

Hey @bugarela, yeah, I can see how `generate` could be confusing. However, it's kind of similar to the generators in property-based testing. It simply generates a value of the proper type (given by the example) that has up to a given number of elements, in case it is a non-scalar type such as a set or a list. The difference from PBT is that generation does not happen at random, but is done by the solver. I can see how using a set instead of a single element could be easier to understand by the people who know TLA+. However, the way `generate` works now is more in line with `oneOf`, which could be also expressed as a quantifier over a set

bugarela:
Ok, so that makes me even more in favor of having the current:
```bluespec
nondet a = Int.allListsUpTo(10).oneOf()
```

and, in addition, the Set version:
```bluespec
nondet a = Int.allSetsUpTo(10).oneOf()
```

I'm using the infinite set `Int` in these examples because iiuc that's the behavior of `Apalache!Gen`. Could we also use `Apalache!Gen` for a bounded set like `Set(1,2,3).allListsUpTo(10).oneOf()`? If not, is it feasible to use a filter, like the  equivalent of `Int.allSetsUpTo(10).filter(s => s.in(Set(1,2,3))).oneOf()`? We could take care of that during the conversion to Apalache IR

konnov:
> 

Well, this would definitely play nice for the set of all integers, as it is a built-in set. Imagine you would like to generate a set of structures, e.g., of the shape `{ x: Set[int], y: int -> bool }`. Now, if you would like to pass a set, you are basically replacing the type with the set. It's very much what TLA+ people are doing when they are writing `TypeOK`. However, this approach does not play well with Apalache, as it sometimes start to unfold the sets. This is exactly what we are trying to avoid with `Apalache!Gen`.

bugarela:
I have to thoughts here:
1. The most natural way for a non TLA+ expert to write this would probably be
```
nondet size = 0.to(10).oneOf()
0.to(size).map(_ =>
  nondet x = Int.oneOf()
  nondet y = setOfMaps(Int, Bool).oneOf()
  { x: x, y: y }
)
```

Of course, this doesn't work. But I'd like to try and get closer to this intuition than introducing a magical `generate` operator. Even if that means adding a keyword.

2. We shouldn't require users to write a dummy value like `{ x: Set(0), y: Map(0 -> false) }`. I'd rather require a type annotation, since Quint is a typed language.

konnov:
> I have to thoughts here:
> 
> 1. The most natural way for a non TLA+ expert to write this would probably be
> 
> ```
> nondet size = 0.to(10).oneOf()
> 0.to(size).map(_ =>
>   nondet x = Int.oneOf()
>   nondet y = setOfMaps(Int, Bool).oneOf()
>   { x: x, y: y }
> )
> ```

This looks really complicated. I agree that this is similar to how one would do it in a property-based framework, since most of them are functional. However, we don't have to do complicated things just because someone did them in the past. Thinking about with a fresh head, I would say that the most natural way for a non-TLA+ person to write this would be something like:

```
fresh a: { x: Set[int], y: int -> bool } up to 10
```

I am kind of making up keywords here. The point is that we essentially have to declare a fresh variable and limit its scope.

> Of course, this doesn't work. But I'd like to try and get closer to this intuition than introducing a magical `generate` operator. Even if that means adding a keyword.
> 
> 2. We shouldn't require users to write a dummy value like `{ x: Set(0), y: Map(0 -> false) }`. I'd rather require a type annotation, since Quint is a typed language.

Yes, this would be great. As an alternative, we could have this form:

```
nondet a: { x: Set[int], y: int -> bool } = generate(10)
```

konnov:
bumped the version of apalache to 0.45.3

konnov:
Hey @bugarela! Any thoughts on this feature? I really need something like to work on an inductive invariant, see [example]( https://github.com/matter-labs/era-consensus/blob/56637c35d23929b45dfdb61a5e508489108f3cec/spec/protocol-spec/simplified/inductive.qnt#L266-L267).

bugarela:
Hi! I guess we agreed on a solution at the end of the last discussion. However, I'm having trouble picturing the use case for this and therefore assessing the priority here, as this doesn't seem trivial to implement.

IIRC, we talked over zoom and you said you'd try to come up with an example to demonstrate where you need this. Are you still willing to do this? We could also jump into a call sometime, whatever works best.

konnov:
> Hi! I guess we agreed on a solution at the end of the last discussion. However, I'm having trouble picturing the use case for this and therefore assessing the priority here, as this doesn't seem trivial to implement.
> 
> IIRC, we talked over zoom and you said you'd try to come up with an example to demonstrate where you need this. Are you still willing to do this? We could also jump into a call sometime, whatever works best.

There is a link to the WIP example in my reply. Sure, happy to jump on a call!

konnov:
@bugarela I have just described the need for value generators in detail in Apalache: https://github.com/apalache-mc/apalache/discussions/3014

I agree that what we currently have in Apalache is a hack, though it's hard to tell how long it would take to implement a better version (proposed there).

konnov:
@bugarela is there any chance to have this PR merged? I think other operators have been added since then and this one is not significantly worse then those

bugarela:
Hi @konnov. This has been in my TODO since we last discuss, but I never decide what do to about it. Sorry about leaving it hanging.

I think we should merge it, while renaming it to `apalache::generate`, to ensure that people who use it are aware this is Apalache-specific. WDYT? I'd still like to think more about this feature and introduce a more complete version of it, but that might take even more time, so this should unblock advanced users for now.

At some point, you suggested:
```
nondet a: { x: Set[int], y: int -> bool } = generate(10)
```

We could also support both (see `q::debug` for a built-in that has an optional argument), and/or add support for this format later on.



And I find it objectively better than the version in the PR. If you think it is a reasonable amount of work make this change, I think it would look much better and even feel closer to Apalache's `Gen` which, from what I gather, also uses the type annotation instead of a value in the parameters.

quint/cli-tests.md
```diff
@@ -184,6 +184,11 @@ This example was pointing to Paxos. Now it does not typecheck.
 <!-- !test check coin - Types & Effects-->
     quint typecheck ../examples/tutorials/coin.qnt
 
+### OK on typecheck generate
+
+<!-- !test check generate - Types & Effects -->
+    quint typecheck ../examples/language-features/generate.qnt
```
bugarela, line 190: Can we have a `quint verify` test as well?

quint/src/effects/NondetChecker.ts
```diff
@@ -89,10 +89,10 @@ export class NondetChecker implements IRVisitor {
 
     // body of nondet must be an application of oneOf
     const body = expr.opdef.expr
-    if (body.kind !== 'app' || body.opcode !== 'oneOf') {
+    if (body.kind !== 'app' || (body.opcode !== 'oneOf' && body.opcode !== 'generate')) {
       this.errors.push({
         code: 'QNT204',
-        message: `'oneOf' must be the outermost expression in a nondet definition`,
+        message: `'oneOf' and 'generate' must be the outermost expressions in a nondet definition`,
```
bugarela:
```suggestion
        message: `the outermost expression in a nondet definition must be either 'oneOf' or 'generate'`,
```

quint/src/effects/NondetChecker.ts
```diff
@@ -89,10 +89,10 @@ export class NondetChecker implements IRVisitor {
 
     // body of nondet must be an application of oneOf
```
bugarela:
```suggestion
    // body of nondet must be an application of oneOf or generate
```

quint/test/effects/NondetChecker.test.ts
```diff
@@ -61,7 +61,7 @@ describe('checkNondets', () => {
     assert.sameDeepMembers(errors, [
       {
         code: 'QNT204',
-        message: "'oneOf' must be the outermost expression in a nondet definition",
+        message: "'oneOf' and 'generate' must be the outermost expressions in a nondet definition",
```
bugarela:
```suggestion
        message: "the outermost expression in a nondet definition must be either 'oneOf' or 'generate'",
```

konnov:
Hey @bugarela! Thanks for coming back to this. Yes, this form would be significantly better:

```
nondet a: { x: Set[int], y: int -> bool } = generate(10)
```

Let me check how hard it would be to implement this one

konnov:
@bugarela I've updated the command syntax to be:

```
nondet x: <type> = generate(<bound>)
<expr>
```

As soon as we have a new release of Apalache that includes https://github.com/apalache-mc/apalache/pull/3138, we can finally make it work.

konnov:
Here is the PR introducing the new version of Apalache: https://github.com/informalsystems/quint/pull/1735

quint/src/builtin.qnt
```diff
@@ -903,6 +903,20 @@ module builtin {
...
   pure def q::debug(msg, value): (str, a) => a
 
+  /// `apalache::generate(n)` generates a value of type `a` while constraining
+  /// all of its children collections to have at most `n` elements. This operator
+  /// must be used together with `nondet`, similar to `oneOf`.
+  ///
+  /// ### Examples
+  ///
+  /// ```quint
+  /// nondet S: Set({a: int, b: Set({ c: str, d: bool })}) = apalache::generate(4)
```
bugarela:
```suggestion
  /// nondet S: Set({ a: int, b: Set({ c: str, d: bool }) }) = apalache::generate(4)
```

bugarela (approved):
Looks great! I tried it locally as well and everything works. Thanks for pushing this

konnov:
Fixed the broken docs and formatting. Please re-run the tests :)

thpani:
@bugarela any chance of cutting a release with this?

bugarela:
@thpani yes, I was planning on doing it today! I'll do it in an hour or two.

bugarela:
@thpani done!

thpani:
@bugarela thank you! 🙏

PR #1465, opened 2024-07-24
Hotfix for high memory usage

### LINKED ISSUE #1456: High memory usage after #1365
My change on https://github.com/informalsystems/quint/pull/1365 made it so memory usage increases a lot during simulation, as we are now storing many traces to select the few best ones latter.

The solution is probably to only store the N best traces, according to how many were requested. Previously, we only stored the single best trace and compared it every time. We can now compare it to the Nth trace, and then only sort the N-best ones when we find a new best one.

(Found by @cason)

bugarela:
Hello :octocat: 

Closes #1456 

In https://github.com/informalsystems/quint/pull/1365 I prioritized performance over memory consumption without thinking too much about it and it was a bad choice. I've heard complains from three people (@josef-widder, @cason and @konnov) about a ridiculous increase in memory consumption by Quint, and this was the cause.

This PR fixes the problem by sacrificing a bit of performance (making a sort operation at every trace) and therefore not having to store all traces found (which would enable a single sorting operation).

Most importantly, the new approach should only impact performance when the new `--n-traces` flag is used. All previously existing operations should no longer be affected by this change. This was the biggest problem, since even when the flag was not used and users wanted a single trace, we were still storing all explored traces and using a lot of memory.

### DIFF
--- a/quint/src/runtime/trace.ts
+++ b/quint/src/runtime/trace.ts
@@ -16,7 +16,6 @@ import { EvalResult } from './runtime'
 import { verbosity } from './../verbosity'
 import { Rng } from './../rng'
 import { rv } from './impl/runtimeValue'
-import { take } from 'lodash'
 
 /**
  * A snapshot of how a single operator (e.g., an action) was executed.
--- a/quint/src/runtime/trace.ts
+++ b/quint/src/runtime/trace.ts
@@ -174,16 +174,14 @@ export interface TraceRecorder extends ExecutionListener {
   clear: () => void
 
   /**
-   * Get the best recorded traces.
-   * @param n the number of traces to get
-   * @returns the best recorded traces.
+   * The best recorded traces.
    */
-  getBestTraces(n: number): Trace[]
+  bestTraces: Trace[]
 }
 
 // a trace recording listener
-export const newTraceRecorder = (verbosityLevel: number, rng: Rng): TraceRecorder => {
-  return new TraceRecorderImpl(verbosityLevel, rng)
+export const newTraceRecorder = (verbosityLevel: number, rng: Rng, tracesToRecord: number = 1): TraceRecorder => {
+  return new TraceRecorderImpl(verbosityLevel, rng, tracesToRecord)
 }
 
 // a private implementation of a trace recorder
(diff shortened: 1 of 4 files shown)


p-offtermatt (approved):
LGTM!
Just summarizing, the difference is that before, onUserOperatorCall would just push all traces in the list, and it would only sort/filter at the end during compileAndRun, versus now every time a trace is produced, the existing traces get sorted by quality and filtered down to the best numberOfTraces many.

Makes sense!

MahtabNorouzi (approved):

MahtabNorouzi (approved):

quint/src/runtime/trace.ts
```diff
@@ -331,10 +327,17 @@ class TraceRecorderImpl implements TraceRecorder {
...
 
-    this.traces.push({ frame: traceToSave, seed: this.runSeed })
+    const traceWithSeed = { frame: traceToSave, seed: this.runSeed }
+
+    this.bestTraces.push(traceWithSeed)
+    this.sortTracesByQuality()
+    // Remove the worst trace (if there more traces than needed)
+    if (this.bestTraces.length > this.tracesToStore) {
+      this.bestTraces.pop()
+    }
```
konnov, line 337: This PR has been merged already. However, here is my concern. Now best traces are sorted on every run, even though only one new trace is added on every run. I don't know how large `this.tracesToStore` would be, but assuming it is `n`, we would have something like `O(n * log(n))` operations spent on sorting in every run. Instead, you could just do one step of insertion sort, assuming that `this.bestTraces` is sorted already. This would give us a simple loop of `n/2` iterations on average.

bugarela: Yes, I'm aware. But as I tried explaining in the description, `n=1` in all existing use cases. It is only greater than 1 if you use the new CLI param `--n-traces`. 

I did look up insertion sort implementations in typescript to use here, but there was nothing very handy and I just wanted to fix the memory issue. This seems like a fun thing to improve and benchmark tho, and my students will need to use this `--n-traces` features by the end of the semester, so I'll take care of performance eventually :grimacing:

konnov: I mean, you don't need any special implementation for insertion sorting. To insert a new element `e`, you just do one iteration over the array `x` and find the index `i` such that `x[i] <= e <= x[i+1]` (of course, you have to take care of the cases `i == 0` and `i == x.length`). Then, you insert `e` at the position `i`. I guess, JS `splice` should be able to do that.

konnov:
thanks for fixing this! It really enables simulation of large specs

PR #1482, opened 2024-08-19
Fix Apalache repository URL

romac:
Apalache has recently moved from the `informalsystems` to the `apalache-mc` organization.


bugarela (approved):
Thank you!

Issue #1484, opened 2024-08-21
In `run` command, option `--n-traces` has no effect unless argument to `--out-itf` ends with `.itf.json`


romac:
If I run Quint on a spec with

```
$ quint run (...) --n-traces 10 --out-itf traces.json
```

I would expect to see 10 traces being generated, but only a single one is output to `traces.json`, thus silently disabling the behavior of `--n-traces`.

I apparently have to specify a `.itf.json` suffix for `--n-traces` to have any effect, which is counterintuitive to me.

My suggestion is to let users choose any filename + extension, and specify the trace number to use in the output using the `{#}` placeholder (much like for the `test` command) instead of having Quint automatically append the trace number to files ending with `.itf.json`.

```
$ quint run (...) --n-traces 3 --out-itf trace_{#}.json
```

The command above would produce `traces_0.json`, `traces_1.json`, `traces_2.json`.

romac:
Note that the `test` command has its own counterintuitive behavior, where it accepts placeholders but those only take effect when the filename ends with `.itf.json`, see #1137.

romac:
I have opened a PR for both issues, but I am happy to discuss alternative solutions.

Issue #1137, opened 2023-09-01
The option --output in quint test requires '.itf.json' extension


ivan-gavran:
If running `quint test` with option `--output` being e.g., `out{#}{}.json`, it complains about getting _Unexpected format in --output_.  (Passing `out{#}{}.itf.json` is fine.)

Why is quint insisting on the extension? Especially given that a similar option `--out-itf` in `quint run` accepts `*.json` files.

konnov:
How does this affect your workflow. This does not seem to be a hinderance of any kind to me.

ivan-gavran:
I noticed it when trying to format the resulting .itf.json in VSCode (which does format .json, but does not .itf.json). There should probably be a VSCode-based fix for my issue, true. 

Regardless of that, I am raising this issue to see whether Quint wants to enforce the extension or if it wants to allow files with any extensions to be generated.

PR #1485, opened 2024-08-21
Change placeholders to `{test}` and `{seq}`, stop enforcing `.itf.json` extension, rename `test --output` to `test --out-itf`

romac:
Closes: #1484
Closes: #1137

## `run`, `test` 
- Change placeholders from `{}` to `{test}` and from `{#}` to `{seq}`

## `run`
- Auto-append trace sequence number to filename if more than one trace is present and `{seq}` is not specified

## `test` 
- Stop enforcing `.itf.json` extension
- Rename `--output` to `--out-itf`, add hidden alias for the former
- Show output (ie. test results) even if `--out-itf` is set

## `verify`
- Add warning if `--out-itf` option contains `{test}` or `{seq}` as those have no effect since Apalache only produces a single trace


romac:
Sure, I'll take a look :)

konnov:
Just a note on `{#}`. I guess, it is my fault introducing it. The issue is that GNU parallel is using the same substitution syntax for the job number. If you plan to use Quint with `parallel`, this is going to be a problem.

bugarela:
> Just a note on `{#}`. I guess, it is my fault introducing it. The issue is that GNU parallel is using the same substitution syntax for the job number. If you plan to use Quint with `parallel`, this is going to be a problem.

Hmm interesting. I think it's a good idea to use the same standard as other program. I guess you can always wrap your quint command in a shell script or something and then call parallel on that to avoid the issue?

bugarela (approved):
LGTM :sparkles:

romac:
Ah good point @konnov, I did not think of that.

I guess the solution @bugarela works, but it's also possible to change the `parallel` replacement string to something else using its `--seqreplace` option: https://www.gnu.org/software/parallel/parallel_tutorial.html

@konnov What do you think, are those workarounds good enough for you?

konnov:
> Ah good point @konnov, I did not think of that.
> 
> I guess the solution @bugarela works, but it's also possible to change the `parallel` replacement string to something else using its `--seqreplace` option: https://www.gnu.org/software/parallel/parallel_tutorial.html
> 
> @konnov What do you think, are those workarounds good enough for you?

Well, it would kind of work, but `parallel` is tough enough alone, so adding complexity on top of it would make things even harder :)

romac:
I agree it's a bit more complex, but you only have to figure out the right invocation of parallel once, then you can re-use it everywhere. On the other hand, changing the variable would involve making a breaking change in the Quint CLI, since we also use `{#}` and `{}` in other commands as well (i.e. `quint test --output`).

That said, I empathize with the goal of building tools (Quint in this case) which work well with other pre-existing tools, as per the UNIX philosophy, so I am not sure how to weigh one against the other.

What do you think, @bugarela?

bugarela:
What about doing something more like what TLC does and just produce a pattern without requiring the user to specify it? It takes away the ability of the user to customize the filenames to their liking, but it makes it easier for new users, which won't have to learn how to write this parameter. 

So if I run
```
quint run --out-itf=foo.json --n-traces=2
```
it will produce `foo1.json` and `foo2.json` (or something like that)

romac:
Agreed, that's perhaps the most user-friendly and pragmatic option 👍

romac:
What shall we do about `quint test --output`?

```
  --output       write a trace for every test, e.g., out{#}{}.itf.json where {}
                 is the name of a test, {#} is the test sequence number [string]
```

Shall we keep the `{}` placeholder, remove `{#}` and automatically append the sequence number like for `run/verify --out-itf`?

romac:
An alternative is to change the placeholders to from `{}` to `{test}` and `{#} to `{seq}` or something like that for `test`, and from `{#}` to `{seq}` for `run` and `verify`, and automatically append the seq number if the `{seq}` parameter is not present.

- `quint run/verify --n-traces 3 --out-itf trace_{seq}.itf.json` would yield `trace_0.itf.json`, `trace_1.itf.json`, `trace_2.itf.json`
- `quint run/verify --n-traces 3 --out-itf trace.itf.json` would yield `trace0.itf.json`, `trace1.itf.json`, `trace2.itf.json`
- `quint test --n-traces 3 --output {test}_{seq}.itf.json` would yield `someTest_0.itf.json`, `someTest_1.itf.json`, `someTest_2.itf.json`
- `quint test --n-traces 3 --output {test}.itf.json` would yield `someTest0.itf.json`, `someTest1.itf.json`, `someTest2.itf.json`

This way we get the best of both worlds, being both user-friendly and configurable.

Also, how about we also change `test --output` to `test --out-itf` for consistency with `run` and `verify`, but keep `--output` as an alias for `--out-itf`?

bugarela:
I think that would be absolutely amazing! It's a good thing to have the templates optionally, and using `seq` and `test` solves the problem of conflict with parallel while also being much clearer. Thanks for thinking this through!

check.sh
```diff
@@ -0,0 +1,20 @@
...
+# Find all .qnt files
+qnt_files=$(find examples -type f -name "*.qnt")
+
+# Initialize a flag to track if any file fails the check
+all_files_valid=true
+
+# Loop through each .qnt file and check its header
+for file in $qnt_files; do
+  if [[ "$(head -n 1 "$file")" != "// -*- mode: Bluespec; -*-" ]]; then
+    echo "File $file does not start with // -*- mode: Bluespec; -*-"
```
bugarela, line 12: Great idea! This will be useful :smile:

quint/io-cli-tests.md
```diff
@@ -864,7 +864,7 @@ rm out-itf-example.itf.json
 
 <!-- !test in run with n-traces itf -->
 ```
-quint run --out-itf=out-itf-example.itf.json --n-traces=3 --max-steps=5 --seed=123  ../examples/tutorials/coin.qnt
+quint run --out-itf='out-itf-example{seq}.itf.json' --n-traces=3 --max-steps=5 --seed=123  ../examples/tutorials/coin.qnt
```
bugarela, line 867: Do we actually need to update this? Isn't the behavior on the old version of the test still the default?

quint/src/cliCommands.ts
```diff
@@ -957,6 +970,65 @@ function isMatchingTest(match: string | undefined, name: string) {
 }
 
 // Derive the verbosity for simulation and verification routines
-function deriveVerbosity(args: { out: string | undefined; outItf: string | undefined; verbosity: number }): number {
-  return !args.out && !args.outItf ? args.verbosity : 0
+function deriveVerbosity(args: { out: string | undefined; verbosity: number }): number {
+  return args.out ? 0 : args.verbosity
+}
+
+const Placeholders = {
```
bugarela, line 977: Can we make it fully capitalized? I don't think we have this patter of capitalization for any values in the code (only classes and types), so it's a bit odd.
```suggestion
const PLACEHOLDERS = {
```

quint/src/cliCommands.ts
```diff
@@ -957,6 +970,65 @@ function isMatchingTest(match: string | undefined, name: string) {
...
+ * @param template the output template
+ * @param index the index of the trace
+ * @param options An object of the form `{ autoAppend: boolean }`
+ * @returns the expanded output template
+ */
+function expandOutputTemplate(template: string, index: number, options: { autoAppend: boolean }): string {
+  if (template.includes(Placeholders.seq)) {
+    return template.replaceAll(Placeholders.seq, index.toString())
+  } else if (options.autoAppend && template.endsWith('.itf.json')) {
+    // Special case for the recommended extension, to avoid adding the index in between `itf` and `json`
```
bugarela, line 1024: I think this is ok, but I guess one alternative would be to add the index right before the first `.` in the file name, so it would work fine with `.itf.json` without having to treat it separately, and, although it doesn't matter a lot, it would work better with arbitrary file names (`foo1.bla.json` is better than `foo.bla1.json` I guess).

check.sh
romac: Woops forgot to remove this, wanted to do this in a GitHub Actions workflow. Will open a separate PR for it.

quint/io-cli-tests.md
romac: My bad, forgot to revert this!

quint/src/cliCommands.ts
romac: Will do

quint/src/cliCommands.ts
romac: Agreed, that's a better solution!

romac: ce713fe4

quint/src/cliCommands.ts
romac: 463d19fe

quint/io-cli-tests.md
romac: 6988f0e0

check.sh
romac: 25de4982

bugarela (approved):
:rocket:

Issue #1457, opened 2024-06-27
Print progress in `run` and `test`


bugarela:
User provides `--max-samples`, and we could print our progress towards trying all those samples. We can use some progress bar lib.

PR #1490, opened 2024-08-30
Add progress bar to `quint test` and `quint run`

romac:
Closes: #1457

### `test`

[asciicast](https://asciinema.org/a/xc0TzITpL1pIVM1fYe9dvW03K)

### `run`

[asciicast](https://asciinema.org/a/TQnPRh54dZOf98tuEOLBus0g8)


romac:
@bugarela What do you think of the result? Shall we add a flag to enable/disable the progress bar? Perhaps we should also only enable it if we are running with a TTY, in order to avoid polluting GitHub actions logs, etc.

bugarela (approved):
I think this is absolutely perfect, it will be very useful (specially with the ETA), and it doesn't seem to pollute the CI tests at all, probably because we always clear the bar when we are done processing, and there should never be a scenario where that doesn't happen.

I think we are all good!
I'll have to add some abstractions to my new version of the simulator to handle this, as it has more abstraction layers than the current one, but I'll figure something out - we should not delay this.

PR #1493, opened 2024-09-02
Relax uppercase check for types qualified with a namespace

romac:
Allows `type foo::Bar = int` whereas it would previously fail with `type names must start with an uppercase letter` because the check was looking at the very first letter of the fully qualified name (ie. `foo::Bar`) instead of the unqualified name (ie. `Bar`).

## Example

```bluespec
module upper {
  type Foo = int
  type foo = int

  type foo::Bar = int
  type foo::bar = int
}
```

### Before

```
upper.qnt:2:3 - error: [QNT007] type names must start with an uppercase letter
2:   type Foo = int
     ^^^^^^^^^^^^^^

upper.qnt:3:3 - error: [QNT007] type names must start with an uppercase letter
3:   type foo = int
     ^^^^^^^^^^^^^^

upper.qnt:5:3 - error: [QNT007] type names must start with an uppercase letter
5:   type foo::Bar = int
     ^^^^^^^^^^^^^^^^^^^

upper.qnt:6:3 - error: [QNT007] type names must start with an uppercase letter
6:   type foo::bar = int
     ^^^^^^^^^^^^^^^^^^^

error: parsing failed
```

### After

```
/Users/romac/Informal/Code/quint/quint/upper.qnt:3:3 - error: [QNT007] type names must start with an uppercase letter
3:   type foo = int
     ^^^^^^^^^^^^^^

/Users/romac/Informal/Code/quint/quint/upper.qnt:5:3 - error: [QNT007] type names must start with an uppercase letter
5:   type foo::bar = int
     ^^^^^^^^^^^^^^^^^^^

error: parsing failed
```

### DIFF
--- a/quint/src/parsing/parseErrors.ts
+++ b/quint/src/parsing/parseErrors.ts
@@ -1,19 +1,21 @@
 import { QuintError } from '../quintError'
 
-export function lowercaseTypeError(id: bigint, name: string): QuintError {
+export function lowercaseTypeError(id: bigint, name: string, prefix: string[]): QuintError {
+  const original = [...prefix, name].join('::')
+  const newName = name[0].toUpperCase() + name.slice(1)
+  const replacement = [...prefix, newName].join('::')
+
   return {
     code: 'QNT007',
     message: 'type names must start with an uppercase letter',
     reference: id,
-    data: name[0].match('[a-z]')
-      ? {
-          fix: {
-            kind: 'replace',
-            original: `type ${name[0]}`,
-            replacement: `type ${name[0].toUpperCase()}`,
-          },
-        }
-      : {},
+    data: {
+      fix: {
+        kind: 'replace',
+        original: `type ${original}`,
+        replacement: `type ${replacement}`,
+      },
+    },
   }
 }
 
--- a/quint/src/parsing/ToIrListener.ts
+++ b/quint/src/parsing/ToIrListener.ts
@@ -294,9 +294,7 @@ export class ToIrListener implements QuintListener {
     const name = ctx.qualId()!.text
     const id = this.getId(ctx)
 
-    if (name[0].match('[a-z]')) {
-      this.errors.push(lowercaseTypeError(id, name))
-    }
+    this.checkForUppercaseTypeName(id, name)
 
     const def: QuintTypeDef = {
       id,
@@ -319,9 +317,8 @@ export class ToIrListener implements QuintListener {
     const type = this.popType().unwrap(() =>
       fail('internal error: type alias declaration parsed with no right hand side')
     )
-    if (name[0].match('[a-z]')) {
-      this.errors.push(lowercaseTypeError(id, name))
-    }
+
+    this.checkForUppercaseTypeName(id, name)
 
     let defWithoutParams: QuintTypeDef = { id: id, kind: 'typedef', name, type }
     const def: QuintTypeDef =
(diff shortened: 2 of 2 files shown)


quint/src/parsing/parseErrors.ts
```diff
@@ -1,19 +1,21 @@
...
-            replacement: `type ${name[0].toUpperCase()}`,
-          },
-        }
-      : {},
+    data: {
+      fix: {
+        kind: 'replace',
+        original: `type ${original}`,
+        replacement: `type ${replacement}`,
+      },
```
bugarela, line 17: This information is used by the LSP as a code action to fix the problem. I don't remember in details how it works, so it's probably a good idea to test it manually.

In order to install a version of the LSP that uses your local version of Quint, you'd need `yalc`. There are some instructions here: https://github.com/informalsystems/quint/blob/main/CONTRIBUTING.md#using-the-vscode-plugin-with-an-unpublished-version-of-quint

This is for VSCode. For vim/emacs, which use the `quint-language-server` package, what I do is:
1. uninstall the package from NPM, if installed `npm uninstall @informalsystems/quint-language-server -g`
2. cd into `vscode/quint-vscode/server` and run `npm install && npm run compile && npm link`

Then, if your vim/emacs is set up to use Quint's language server, it should pick up the new version which uses your local version of Quint. And you can test writing a `type foo::bar` and checking if it gets fixed into `type foo::Bar` after calling the code action.

If you can't or don't want to do all of that, lmk. I already have most of this in place, so I can test it in a few minutes for you.

bugarela (commented):
This is great! I think we can easily add a unit test for this: we can just add a type declaration with the new supported name pattern to `SuperSpec.qnt` and then update the expected fixtures with `npm run update-fixtures`. If all works well, the new fixture will be only slightly different, but still a success - while before, a type like that would cause errors. It is not a really good unit test setup, but at least it will prevent us from regressing from this.

quint/src/parsing/parseErrors.ts
romac: Works as expected :)

Screenshot 2024-09-03 at 15 27 37

Same with no prefix:

Screenshot 2024-09-03 at 15 27 58

romac: Only tested in Neovim so far. If you already have VS Code setup might be easier for you to give it a try. Otherwise I can always set up myself.

bugarela: yey awesome!

romac:
CI is failing with

> ERROR: Generated files are not up to date. Run 'npm run generate'

But when I run the suggested command I don't see any changes in my working copy

quint/src/parsing/parseErrors.ts
bugarela: No need to test it in more than one editor. All of them use the same protocol, so we are safe.

bugarela:
> On CI, the `quint-test-generated-up-to-date` is failing with
> 
> > ERROR: Generated files are not up to date. Run 'npm run generate'
> 
> But when I run the suggested command I don't see any changes in my working copy

Unfortunately this command (or the `npm run update-fixtures` which is part of it) generates also files that are not relevant to us, and are not tracked by git, so what I do is not commit them and run `git clean -f testFixture/*` to remove them from my working copy. I see that you committed those files in your last commit, and that's is likely what is causing the CI to break.

Can you remove those extra files? All of them but the SuperSpec* ones. Then I think it will work.

romac:
> I see that you committed those files in your last commit

Indeed, my bad sorry! Fixed now.

bugarela (approved):
Amazing, thanks again!

Issue #1418, opened 2024-04-02
Internal error while flattening [QNT404] Name 'Listed' not found


eras:
Trying to verify the accompanying model results in error 

```
Error: Internal error while flattening [QNT404] Name 'Listed' not found
    at resolveNamesOrThrow (/home/user/.npm-packages/lib/node_modules/@informalsystems/quint/dist/src/flattening/fullFlattener.js:91:15)
    at flattenModules (/home/user/.npm-packages/lib/node_modules/@informalsystems/quint/dist/src/flattening/fullFlattener.js:62:35)
    at compile (/home/user/.npm-packages/lib/node_modules/@informalsystems/quint/dist/src/cliCommands.js:493:104)
    at EitherConstructor.asyncChain (/home/user/.npm-packages/lib/node_modules/@informalsystems/quint/node_modules/@sweet-monads/either/cjs/index.js:125:16)
    at /home/user/.npm-packages/lib/node_modules/@informalsystems/quint/dist/src/cli.js:58:39
```

Steps to reproduc:
1) Install `quint` version 0.19.1
1) Save the following to `Simple.qnt`:
```
module T {
  type State =
    | Listed

  type File = {
    state : State,
  }

  def listed(files: int -> File) : Set[int] =
    files.keys().filter(x => files.get(x).state == Listed)
}

module Spec {
  import T.*

  const init_files: int -> File

  var files: int -> File

  action init = all {
    files' = init_files
  }

  action step = {
    nondet file_id = oneOf(listed(files))
    files' = files
  }
}

module Simple {
  import T.*
  pure val file1: File = {
    state  : Listed,
  }
  import Spec(init_files = Map(1->file1)).*
}
```
2) `quint verify Simple.qnt`
3) Witness the error

Expected results:
- No error; possibly an explanation what I might be doing incorrectly, though it _feels_ correct to me..

It works if I move the definition `listed` to module `Spec`. It also works if I incorporate the `Simple` module inside `Spec`.

bugarela:
Hi, thanks for reporting! Seems it might be related to the error I've hit in https://github.com/informalsystems/quint/issues/1353.

Unfortunately, Quint's flattening (which converts multiple modules into a single one) is probably our most unstable feature, specially in the presence of constants and instances. I'm glad you found a workaround!

Issue #1364, opened 2024-02-08
Flattening issue when importing variables by different paths with different names


bugarela:
The example on this PR results in a invalid flattened module https://github.com/informalsystems/quint/pull/1363.

I tried fixing this in the compiler, which is enough for the simulator. But Apalache can't handle multiple names for the same variable. The flattener should ensure we always refer to that variable with the same name.

Issue #1353, opened 2024-01-24
Flattening issue with multiple names for the same type constructor


bugarela:
I haven't investigated the problem yet, but this hash breaks flattening. Seems like the same type constructor is available both with the `execute::` namespace and without it, but only one version get's flattened and the other is missing.

https://github.com/informalsystems/security-tooling-incubator/commit/c9ef1b7e4895d50a0468d4749a1cfc890e7fa1b1

Command: `quint run donations.qnt --main=state_machine`

PR #1495, opened 2024-09-02
New evaluator that doesn't depend on flattening

### LINKED ISSUE #288: Extend `_lastTrace` when evaluating actions

### LINKED ISSUE #434: redefining a function gives a "syntax error"
In the repl, 

```
>>> def farenheit(celsius) = celsius * 9 / 5 + 32

>>> def farenheit(celsius) = celsius * 9 / 5 + 32
syntax error: <input>:4:1 - error: Conflicting definitions found for name farenheit in module __REPL
4: def farenheit(celsius) = celsius * 9 / 5 + 32
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<input>:5:1 - error: Conflicting definitions found for name farenheit in module __REPL
5: def farenheit(celsius) = celsius * 9 / 5 + 32
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I get a syntax error, even though my syntax is fine. The error message is super useful; I would just put it in a category other than "syntax".

### LINKED ISSUE #734: Random search should react to SIGINT
Currently, there is no good way to terminate a long random search, but to kill the node process. The tool should react to SIGINT and interrupt the search loop in this case, if it's possible at all in node.

### LINKED ISSUE #830: Refactor code compilation
A follow-up of #800. The current set of functions was incrementally developed, in order to serve the needs of REPL, `quint run`, and `quint test`. The call graph is shown below. We should find a better way of parsing and compiling.

compile

### LINKED ISSUE #924: `<undefined value>` in the REPL when evaluating
Given

```
module Test {
   def condition(f) = {
     Set(1).filter(f)
   }

   def demo = condition(x => x > 0)
}
```

Run with

```
$  echo 'demo' | quint -r Test.qnt::Test
Quint REPL 0.11.1
true

>>> demo
<undefined value>
>>> 
```

It does not occur if we inline the operator call like

```
module Test {
   def condition(f) = {
     Set(1).filter(x => x > 0)
   }

   def demo = condition(x => x > 0)
}
```

In general, we should not allow any undefined values to be given to the user. we should check for this at the last possible place and report an internal error if an undefined value would be given.

This was reported by @otrack.

### LINKED ISSUE #934: Make name resolution incremental

### LINKED ISSUE #1035: Name resolution can fail in the simulator depending on operator order
I found a strange edge case in which name resolution of an imported operator incorrectly fails, but only if the operators are imported in a specific order and the main module is loaded in a specific way :woman_shrugging: 

This issue is light on details because I've opened a draft PR with a failing integration test in #1036.

### LINKED ISSUE #1052: Remove `compileFromCode` and use more appropriate interfaces

### LINKED ISSUE #1058: Use `Either` instead of `Option` in `RuntimeValue`
`RuntimeValue` returns plenty of `none()` values, e.g.:

https://github.com/informalsystems/quint/blob/be8bb62dbc046da134d07f44367c4a60970adb45/quint/src/runtime/impl/runtimeValue.ts#L1330-L1354

We are losing an opportunity to explain what kind of error has happened. If we switch to `Either`, we can provide better feedback in the simulator/REPL.

### LINKED ISSUE #1103: Make (the new) flattening incremental
After https://github.com/informalsystems/quint/issues/1102, our flattening is no longer done incrementally. We should find a way to make that incrementally with this new procedure.

### LINKED ISSUE #1151: Unhandled error if `init` and `step` params in the CLI have invalid syntax.
We should make sure we give proper syntax feedback for those.

### LINKED ISSUE #1184: How to run tests in deep modules?
My spec is this:
https://github.com/cosmos/interchain-security/blob/e09968d94a4e5a84bf144e5f0737e1f529494772/tests/difference/core/quint_model/ccv.qnt#L212

When I run ```quint test ccv.qnt --main CCVDefaultStateMachine```
I would expect the tests in the CCV module to be run.
They are not; I guess because they themselves are not imported in CCVDefaultStateMachine, because CCV is not exported by CCVStatemachineLogic.

Adding `export CCV.*` to CCVStatemachineLogic (e.g. after import CCV.* does not work), when running quint test I get this:
```
Error: Internal error while flattening [QNT404] Name 'PROVIDER_CHAIN' not found,[QNT404] Name 'PROVIDER_CHAIN' not found
    at resolveNamesOrThrow (/Users/offtermatt/projects/quint/quint/dist/src/flattening/fullFlattener.js:90:15)
    at flattenModules (/Users/offtermatt/projects/quint/quint/dist/src/flattening/fullFlattener.js:61:35)
    at runTests (/Users/offtermatt/projects/quint/quint/dist/src/cliCommands.js:248:104)
    at EitherConstructor.asyncChain (/Users/offtermatt/projects/quint/quint/node_modules/@sweet-monads/either/cjs/index.js:125:16)
    at /Users/offtermatt/projects/quint/quint/dist/src/cli.js:33:39
```

If I add `export CCV` without .*, I get this error:
```
Error: Internal error while flattening [QNT404] Name 'UnbondingPeriodPositiveTest' not found,[QNT404] Name 'VscTimeoutPositiveTest' not found,[QNT404] Name 'CcvTimeoutPositiveTest' not found,[QNT404] Name 'CcvTimeoutLargerThanUnbondingPeriodTest' not found,[QNT404] Name 'ProviderIsNotAConsumerTest' not found,[QNT404] Name 'CcvTimeoutKeysTest' not found,[QNT404] Name 'UnbondingPeriodKeysTest' not found
    at resolveNamesOrThrow (/Users/offtermatt/projects/quint/quint/dist/src/flattening/fullFlattener.js:90:15)
    at flattenModules (/Users/offtermatt/projects/quint/quint/dist/src/flattening/fullFlattener.js:73:24)
    at runTests (/Users/offtermatt/projects/quint/quint/dist/src/cliCommands.js:248:104)
    at EitherConstructor.asyncChain (/Users/offtermatt/projects/quint/quint/node_modules/@sweet-monads/either/cjs/index.js:125:16)
    at /Users/offtermatt/projects/quint/quint/dist/src/cli.js:33:39
```

Is there a way to do what I am trying to do, i.e. run the tests in CCV, but with the instantiation I have done through CCVStatemachineLogic from CCVDefaultStateMachine? I am not sure if any of the behaviour here is a bug. At least from a users perspective, the error messages seem weird, but I can't quite tell.

### LINKED ISSUE #1330: Promote effect and type checking errors in repl to errors
Apparently this will fix #428

### LINKED ISSUE #1342: Issue loading modules when importing something in a loaded REPL
Reported by @p-offtermatt 

```
quint -r ccv_model.qnt           
Quint REPL 0.18.0
>>> (node:39269) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
>>> import ccv_model.*
syntax error: error: [QNT405] Module 'ccv_model' not found
import ccv_model.*
```

hash: 7b2357452032ce510e002b325b8016d3f27651bf

Upon some investigation, I believe the problem is that we are not sorting modules incrementally in the repl. In the first compilation, made upon the CLI call, there is no dependency between the `__repl__` module and the `ccv_model` module. Therefore, it is possible to resolve names and evaluate `__repl__` before `ccv_model`. After the import statement, this is no longer possible, but we never re-sort the modules according to the new dependency.

[Note to future self] I've attempted to solve this for a couple of hours today, but couldn't make it work.
- Calling phase 2 on the incremental compilation step results in duplication of modules on the `sourceToModules` map.
- Calling only the sorting results in flattening errors.

### LINKED ISSUE #1439: Simulator crashes with invalid `init` action
If the init action refers to vars, the simulator breaks. Instead, we should properly report/explain the error.

MWE:
```bluespec
module test {
  var x: int

  action init = x' = x
  action step = x' = x
}
```

Results in:
```
$ quint run test.qnt
quint run <input>

Simulate a Quint specification and check invariants

Options:
  --help         Show help                                             [boolean]
  --version      Show version number                                   [boolean]
  --out          output file (suppresses all console output)            [string]
  --main         name of the main module (by default, computed from filename)
                                                                        [string]
  --out-itf      output the trace in the Informal Trace Format to file
                 (suppresses all console output)                        [string]
  --max-samples  the maximum on the number of traces to try
                                                       [number] [default: 10000]
  --max-steps    the maximum on the number of steps in every trace
                                                          [number] [default: 20]
  --init         name of the initializer action       [string] [default: "init"]
  --step         name of the step action              [string] [default: "step"]
  --invariant    invariant to check: a definition name or an expression
                                                    [string] [default: ["true"]]
  --seed         random seed to use for non-deterministic choice        [string]
  --verbosity    control how much output is produced (0 to 5)
                                                           [number] [default: 2]

AssertionError [ERR_ASSERTION]: invalid simulation failed to produce a result
    at compileAndRun (/home/gabriela/projects/quint/quint/dist/src/simulation.js:80:26)
    at runSimulator (/home/gabriela/projects/quint/quint/dist/src/cliCommands.js:407:51)
    at EitherConstructor.asyncChain (/home/gabriela/projects/quint/quint/node_modules/@sweet-monads/either/cjs/index.js:125:16)
    at /home/gabriela/projects/quint/quint/dist/src/cli.js:58:39 {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
```

### LINKED ISSUE #1471: Hidden error when calling `get` with no key inside a fold
similar to https://github.com/informalsystems/quint/issues/1439

A spec like this

```
module test {
  pure val faulty = Map().get("a")

  pure val foo = {
    Set(1,2).fold(0, (a, b) => a + faulty)
  }

  var x: int
  action init = x' = 0
  action step = x' = foo
}
```

Fails with
```
$ quint run test.qnt --max-samples=1
quint run <input>

Simulate a Quint specification and check invariants

Options:
  --help         Show help                                             [boolean]
  --version      Show version number                                   [boolean]
  --out          output file (suppresses all console output)            [string]
  --main         name of the main module (by default, computed from filename)
                                                                        [string]
  --out-itf      output the trace in the Informal Trace Format to file
                 (suppresses all console output)                        [string]
  --max-samples  the maximum number of runs to attempt before giving up
                                                       [number] [default: 10000]
  --n-traces     how many traces to generate (only affects output to out-itf)
                                                           [number] [default: 1]
  --max-steps    the maximum on the number of steps in every trace
                                                          [number] [default: 20]
  --init         name of the initializer action       [string] [default: "init"]
  --step         name of the step action              [string] [default: "step"]
  --invariant    invariant to check: a definition name or an expression
                                                    [string] [default: ["true"]]
  --seed         random seed to use for non-deterministic choice        [string]
  --verbosity    control how much output is produced (0 to 5)
                                                           [number] [default: 2]
  --mbt          (experimental) whether to produce metadata to be used by
                 model-based testing                  [boolean] [default: false]

AssertionError [ERR_ASSERTION]: invalid simulation failed to produce a result
    at /home/gabriela/projects/quint/quint/dist/src/simulation.js:81:30
    at Array.forEach (<anonymous>)
    at compileAndRun (/home/gabriela/projects/quint/quint/dist/src/simulation.js:79:15)
    at runSimulator (/home/gabriela/projects/quint/quint/dist/src/cliCommands.js:422:51)
    at EitherConstructor.asyncChain (/home/gabriela/projects/quint/quint/node_modules/@sweet-monads/either/cjs/index.js:125:16)
    at /home/gabriela/projects/quint/quint/dist/src/cli.js:60:39 {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
```

We need to make sure all of simulation uses Either and not Maybe, and that all errors (left branches) are properly propagated.

bugarela:
Hello :octocat: 

This is a huge PR because it changes the core of our evaluation process to not depend on flattening anymore. I've moved a lot of things around and attempted many different things throughout this process, and I finally have a version that I think is good enough to ship.

Since this is pretty much a rewrite, we are getting rid of a lot of tech debt and fixing many bugs. Here's a list of what I could find:
- Closes #1471 
- Closes #1439
- Closes #1330 
- Closes #1342
- Closes #1058 
- Closes #1052 
- Closes #934 
- Closes #1035 
- Closes #924 
- Closes #830 
- Closes #734
- Closes #288 
- Closes #434 
- Closes #1184
- Closes #1151
- Closes #1103 (as it won't be needed anymore) 

The following issues are consequences of flattening bugs. After this PR, the issues stop happening for `test`, `run`, and `repl`, and only happen in `verify` and `compile` commands, as we still need flattening for Apalache. So the impact of these issues was significantly reduced.
- #1353 
- #1418 
- #1364 

## Change overview

The main change is that, while the old evaluator can evaluate a single flattened module, this version can evaluate expressions, definitions, simulations and tests over the original modules, avoiding the flattening pass (which is slow and buggy).

Related to that, the new version compiles from a starting point: either an expression/definition typed into the REPL, a combiation of init/step/inv given to `simulate`, or from a test definition. It won't even touch code that is not a dependency of this entrypoint. The old evaluator would go through the entire flattened module and compile everything beforehand.

The biggest challenge is to handle constants. This was previously done by the flattener, and now we need to take care of them while simulating. A good example of this is the `instances.qnt` file:
https://github.com/informalsystems/quint/blob/351b52a05ad3e84bbc525f0414161d3a109374b4/examples/language-features/instances.qnt#L38-L45

Evaluating `V1::init` and `V2::init` are two different things, even though both point to the same definition in the lookup table. To make sure we use different constant values for different constants, and that we use different variable registers for variables that come from the same module but through different instances, we need to track a some extra information while building (compiling) expressions.

### Structure

The new evaluator has two main steps:
1. Building (`builder.ts`). Take Quint expressions and build typescript arrow functions that can be called to obtain the value of that expression under a certain context
2. Evaluating (`evaluator.ts). Take the built arrow functions and call them, keeping track of the context. 

## Performance

Although performance changes are not consistent, my evaluation is that it it overall better. 

For the REPL, it is a complete game changer. For the malachite project, evaluating any expression in the REPL would take around 20 seconds, while now it is completely instant.

For runs and tests, it is faster for most cases, sometimes taking half of the time. For some cases, it is slower, but I couldn't find any case where it takes double of the time or more. I want to further improve this version so we can get faster results for all scenarios, but I think this version is good enough to release, specially considering the REPL performance and how many bugs it fixes. See benchmark details below.

<details>
<summary>Benchmarks</summary>
"old version" in the benchmarks is the last released version. Ideally I should benchmark against main, but this is easier. Also, we are measuring simulation/testing time, and performance changes outside of this execution scope are not accounted.

Cases where the new version is better than the old one:
```
quint run --main=ics20Test --invariant=BalanceNonNegative ../examples/cosmos/ics20/ics20.qnt  --seed=0xbf95eeae47ada
old: (50803ms)
new: (25292ms)

quint run --main=prisoners3 --invariant='countInv and safetyInv' ../examples/puzzles/prisoners/prisoners.qnt
old: (8033ms)
new: (4848ms)

quint run --main=prisoners3 --invariant='countInv and safetyInv' ../examples/puzzles/prisoners/prisoners.qnt --max-samples=50000
old: (39185ms)
new: (23397ms)

quint test mysticeti_paper_test.qnt --max-samples=100
old: (24329ms)
new: (12704ms)

quint run mysticeti_paper_test.qnt --init=init_with_dag --max-steps=1 --max-samples=100
old: (9157ms)
new: (4187ms)

PS: The mysticeti ones actually have a flattening bug in the old version and I have to make a workaround

quint test --max-samples 1000 specs/quint/tests/disagreement/disagreementTest.qnt
old: (17998ms)
new: (16574ms)

This is the command that runs on Malachite's CI:
time bash scripts/quint-forall.sh "test --max-samples 100" specs/quint/tests/**/*Test.qnt
new:
Executed in  117.45 secs    fish           external
   usr time  152.82 secs    0.00 micros  152.82 secs
   sys time    6.73 secs  663.00 micros    6.73 secs
old:
Executed in  163.25 secs    fish           external
   usr time  196.00 secs  348.00 micros  196.00 secs
   sys time    7.89 secs  210.00 micros    7.89 secs
```

Cases where the new version is worse than the old one:
```
quint test --max-samples=100 --main TendermintModels ../examples/cosmos/tendermint/TendermintModels.qnt
old: (3653ms)
new: (6870ms)

quint test --main ics20Test ../examples/cosmos/ics20/ics20.qnt
old: (61450ms)
new: (101720ms)

quint test --main=properChannelsTests ../examples/cosmos/ics20/denomTrace.qnt
old: (4635ms)
new: (6770ms)

quint test --max-samples 100 specs/quint/tests/line28/line28Test.qnt
old: (20624ms)
new: (25692ms)
```
</details>


bugarela:
Hi @romac @MahtabNorouzi! I realize this is way too big to review. Unfortunately, the nature of this work and my current resources made it so this is the only feasible option. @josef-widder has been using a version of this since last week as a beta tester, and I'll also be using this heavily over this week. I've also been testing this with our set of examples and I'm confident with its current state.

So I'd like to ask that you give this a look, however it fits in your time and priorities right now. Let me know if you think this is worth merging/releasing, or if there are any other requirements I should meet before putting it out.

I'm also happy to answer any questions or go through stuff in detail either in here, slack or zoom.

bugarela:
@konnov if you want to take a look at this, I'm sure you'll have valuable feedback. I think this is an evolution of what you wrote, and I would not be able to come up with it without the base you provided through your implementation.

konnov:
> @konnov if you want to take a look at this, I'm sure you'll have valuable feedback. I think this is an evolution of what you wrote, and I would not be able to come up with it without the base you provided through your implementation.

It's a massive change. I think it would make more sense to break it down into much smaller PRs. Reviewing such a big change in the simulator is a full-time job for a week. If you need help with that, Informal management knows my hourly rates.

bugarela:
> It's a massive change. I think it would make more sense to break it down into much smaller PRs. Reviewing such a big change in the simulator is a full-time job for a week. If you need help with that, Informal management knows my hourly rates.

Of course, yes. I've always be very in favor of small PRs and feel bad that this is so big. However, in this particular case, I don't think the effort taken (probably more than a week) to break this down would be compensated in significantly more informed reviews, considering the people available to review it at this point.

On top of that, I mentioned you here to make you aware of this change. I've tested the new version in many of our internal and public Quint projects and it's all good. I know you are also writing Quint specs, and I'd rather have you informed than surprised by such meaningful change in the simulator.

And sorry if I implied that I expected you to work for free! I'm just trying to be considerate of your participation in this as best as I can.

konnov:
> > It's a massive change. I think it would make more sense to break it down into much smaller PRs. Reviewing such a big change in the simulator is a full-time job for a week. If you need help with that, Informal management knows my hourly rates.
> 
> Of course, yes. I've always be very in favor of small PRs and feel bad that this is so big. However, in this particular case, I don't think the effort taken (probably more than a week) to break this down would be compensated in significantly more informed reviews, considering the people available to review it at this point.
> 
> On top of that, I mentioned you here to make you aware of this change. I've tested the new version in many of our internal and public Quint projects and it's all good. I know you are also writing Quint specs, and I'd rather have you informed than surprised by such meaningful change in the simulator.
> 
> And sorry if I implied that I expected you to work for free! I'm just trying to be considerate of your participation in this as best as I can.

Hey @bugarela! No worries! I am just saying that I could have a look at smaller PRs. Understanding a massive change is quite hard without having context. Normally the context exchange happens during regular meetings. Since we don't have those anymore, reviewing such a big change would be similar to a security audit to me, of course, the focus would not be on the security bugs.

If you have no other choice to refactor this big PR into smaller chunks, I would recommend you to introduce a flag to run both the old simulator and the new one in a lockstep, so you would be able to quickly detect when they start to deviate.

quint/src/names/collector.ts
```diff
@@ -143,8 +148,6 @@ export class NameCollector implements IRVisitor {
         return
       }
 
-      // Update the definition to point to the expression being overriden
-      constDef.id = ex.id
```
bugarela, line 147: This was a kind of a hack so the compiler would evaluate the constant as the expression it was overridden to. We now do it on the level of registers on the evaluator.

quint/src/static/callgraph.ts
```diff
@@ -190,9 +190,6 @@ export class CallGraphVisitor implements IRVisitor {
     const importedModule = this.context.modulesByName.get(decl.protoName)
     if (importedModule) {
       this.graphAddAll(decl.id, Set([importedModule.id]))
-      decl.overrides.forEach(([_param, expr]) => {
-        this.graphAddAll(expr.id, Set([decl.id]))
-      })
```
bugarela, line 195: This is from a reverted commit, which added a hack for something we don't need anymore

MahtabNorouzi (approved):

PR #1497, opened 2024-09-04
examples: add Lamport's concurrency teaching algorithm

lucab:
This adds a simple concurrent algorithm described by Lamport in https://lamport.azurewebsites.net/pubs/teaching-concurrency.pdf.
It provides an arbitrary-N module and a concrete-N instance.
It shows how to use temporal formulas and check for (eventual and correct) termination in the verifier.

### DIFF
--- a/examples/classic/distributed/TeachingConcurrency/teachingConcurrency.qnt
+++ b/examples/classic/distributed/TeachingConcurrency/teachingConcurrency.qnt
@@ -38,14 +38,11 @@ module teachingConcurrencyN {
     // Stepping action: randomly pick a pending process (if
     // any) and execute it. Otherwise stutter forever, as
     // every process has stopped.
-    action step = any {
-        all {
-            not(termination),
-            processUpdate,
-        },
-        all {
-            termination,
-            allStopped,
+    action step = {
+        if (termination) {
+            allStopped
+        } else {    
+            processUpdate
         }
     }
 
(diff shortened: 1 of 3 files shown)


examples/classic/distributed/TeachingConcurrency/teachingConcurrency.qnt
```diff
@@ -0,0 +1,101 @@
...
+    action step = any {
+        all {
+            not(termination),
+            processUpdate,
+        },
+        all {
+            termination,
+            allStopped,
+        }
+    }
```
bugarela, line 50: This specific pattern is exactly an if-then-else. I think it reads much easier as:
```suggestion
    action step = {
        if (termination) {
            allStopped
        } else {    
            processUpdate
        }
    }
```

bugarela (approved):
Thank you for adding this! Looks great, I only have one suggestion.

examples/classic/distributed/TeachingConcurrency/teachingConcurrency.qnt
lucab: Ah thanks, I didn't realize that; I've directly committed your suggestion.

Speaking of idiomatic patterns, is there a better away to express the termination state, without an explicit stuttering/nop step? I tried avoiding the `allStopped` action, but Apalache (correctly) complains about a deadlock. I didn't see a way to give a hint to the verifier that the execution can be halted at that point (so I don't really care about reassigning all state vars).

PR #1511, opened 2024-09-19
Bump Apalache to 0.45.6

bugarela:
Hello :octocat: 

We just released a version of Apalache that includes an `apalache-mc.bat` file, enabling Quint to start that file in order to run Apalache on Windows.

This PR bumps the Apalache version so we can use that, and then adds CHANGELOG entries related to Windows so we can cut a release


konnov:
I would recommend you to bump to 0.46.1 right away. It has two nice fine-tuning features for `quint verify`: https://github.com/apalache-mc/apalache/blob/main/CHANGES.md#0461---2024-09-24

MahtabNorouzi (approved):

PR #1519, opened 2024-09-30
updated files to support new tuple representation

MahtabNorouzi:
Updated the tuple structure across the codebase from { kind: 'app', opcode: 'Tup', args: [...] } to { kind: 'tuple', elements: [...] }. This refactor improves clarity and consistency in tuple handling and removes support for the previous Tup syntax. All affected files and references have been modified accordingly.


quint/src/ir/namespacer.ts
```diff
@@ -63,6 +63,27 @@ class Namespacer implements IRTransformer {
...
+      return element
+    })
+
+    // Return a new tuple with namespaced elements
+    return {
+      ...tup,
+      elements: namespacedElements,
+    }
+  }
+
```
bugarela, line 86: We actually don't need this! The IR transformer itself should already guarantee that we visited/transformed all of the elements before reaching the tuple. This works as it is right now because `namespacedName` is idempotent (it can be applied more than once and the result will be the same as if it was applied once, that is, `f(x) = f(f(x))`), but it will also work as if we remove this, and we should :smile: 
```suggestion
```

quint/src/runtime/impl/runtimeValue.ts
```diff
@@ -154,7 +154,9 @@ export const rv = {
    */
   mkMap: (elems: Iterable<[RuntimeValue, RuntimeValue]>): RuntimeValue => {
     // convert the keys to the normal form, as they are hashed
-    const arr: [RuntimeValue, RuntimeValue][] = Array.from(elems).map(([k, v]) => [k.normalForm(), v])
+    const arr: [RuntimeValue, RuntimeValue][] = Array.from(elems).map(([k, v]) => {
+      return [k.normalForm(), v] // Ensure normalForm works with tuples
```
bugarela, line 158: I think this was a leftover?
```suggestion
      return [k.normalForm(), v]
```

quint/src/runtime/impl/runtimeValue.ts
```diff
@@ -712,8 +714,14 @@ abstract class RuntimeValueBase implements RuntimeValue {
     if (this instanceof RuntimeValueStr && other instanceof RuntimeValueStr) {
       return this.value === other.value
     }
-    if (this instanceof RuntimeValueTupleOrList && other instanceof RuntimeValueTupleOrList) {
-      return this.kind === other.kind && this.list.equals(other.list)
+    if (this instanceof RuntimeValueList && other instanceof RuntimeValueTuple) {
+      return false
+    }
```
bugarela, line 719: You can safely remove this since this scenario (one list and one tuple) won't match any of the other patterns, and the default return is false (when no pattern is matched - line 764)
```suggestion
```

quint/src/runtime/impl/runtimeValue.ts
```diff
@@ -865,19 +873,53 @@ class RuntimeValueStr extends RuntimeValueBase implements ValueObject {
...
+  toQuintEx(gen: IdGenerator): QuintEx {
+    // Collect elements for QuintEx representation
+    const elems: QuintEx[] = []
+    for (const e of this.elements) {
+      elems.push(e.toQuintEx(gen))
+    }
+    return {
+      id: gen.nextId(),
+      kind: 'tuple',
+      elements: elems, // Adjusted to reflect the new tuple structure
```
bugarela, line 909:
```suggestion
      elements: elems,
```

quint/src/types/constraintGenerator.ts
```diff
@@ -166,6 +167,33 @@ export class ConstraintGeneratorVisitor implements IRVisitor {
     this.addToResults(e.id, right(toScheme({ kind: e.kind })))
   }
 
+  exitTuple(e: QuintTup) {
```
bugarela, line 170: As we did with the effects inferrer, I think we can also generalize something here so we don't have to repeat this code between `exitTuple` and `exitApp`. This will also repeat with the other structures (lists, maps, records, etc), so we can try to generalize when adding the next one.

quint/testFixture/SuperSpec.qnt
```diff
@@ -190,7 +191,8 @@ module withConsts {
     my_rec.a
 
   val tup_elem =
-    val my_tup = Tup("a", 3)
+    // val my_tup = Tup("a", 3)
```
bugarela, line 194:
```suggestion
```

PR #1520, opened 2024-10-01
Terminate Apalache server on Linux properly

konnov:
When running massive experiments on a Linux machine, I have noticed that the instances of Apalache server are spawned via `quint verify` but not properly terminated. This is especially visible when running multiple model checking instances in parallel. In the end, the machine runs out of swap space.

The main reason for that is spawning `apalache-mc` via shell from NodeJS, which does not terminate the grandchild process (java). This PR simply changes the spawning option `shell` to `false`. I have checked it on a Linux machine. Now the server is terminating.


bugarela (approved):
Awesome! I was having some weird issues myself, this was probably the cause.

konnov:
@bugarela Amazing! It looks like you have to merge it though.

PR #1521, opened 2024-10-02
Add the option `--apalache-version` to `quint verify`

konnov:
Since every release of Apalache requires a PR in Quint, this PR introduces the option `--apalache-version`. This option allows the Quint users to pull different versions of Apalache without waiting for a new release of Quint. For instance, the new version of Apalache v0.47.0 downgrades the version of z3. 

### DIFF
--- a/quint/src/cli.ts
+++ b/quint/src/cli.ts
@@ -124,7 +124,7 @@ const compileCmd = {
         default: verbosity.defaultLevel,
       })
       .option('apalache-version', {
-        desc: 'The version of Apalache to use, if it has to be downloaded',
+        desc: 'The version of Apalache to use, if no running server is found (using this option may result in incompatibility)',
         type: 'string',
         default: DEFAULT_APALACHE_VERSION_TAG,
       })
--- a/quint/src/cli.ts
+++ b/quint/src/cli.ts
@@ -328,7 +328,7 @@ const verifyCmd = {
         default: verbosity.defaultLevel,
       })
       .option('apalache-version', {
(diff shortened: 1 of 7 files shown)


konnov:
The failing integration tests does not seem to do with the PR. Is it a flaky test?

bugarela:
> The failing integration tests does not seem to do with the PR. Is it a flaky test?

It is. Idk why, but sometimes the file disappears before we remove it or something. I'll run it again, that usually fixes it. I still need to figure out what is happening and find a proper fix.

quint/src/cli.ts
```diff
@@ -123,6 +123,11 @@ const compileCmd = {
         type: 'number',
         default: verbosity.defaultLevel,
       })
+      .option('apalache-version', {
+        desc: 'The version of Apalache to use, if it has to be downloaded',
```
bugarela, line 127: iiuc, this would also work if I have multiple versions already downloaded and would like to use one of them, right?
Also, it would be good to add a disclaimer that not all versions of Apalache are compatible to all versions of Quint.
```suggestion
        desc: 'The version of Apalache to use, if no running server is found (using this option may result in incompatibility)',
```

quint/src/cli.ts
```diff
@@ -322,6 +327,11 @@ const verifyCmd = {
...
```
bugarela, line 331:
```suggestion
        desc: 'The version of Apalache to use, if no running server is found (using this option may result in incompatibility)',
```

bugarela (approved):
Looks useful, specially for us Quint/Apalache devs. I just added a suggestion on the option description.

PR #1528, opened 2024-10-08
VSCode Release v0.14.7

bugarela:

### DIFF
--- a/vscode/quint-vscode/CHANGELOG.md
+++ b/vscode/quint-vscode/CHANGELOG.md
@@ -14,6 +14,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 ### Security
 
+## v0.14.7 -- 2024-10-08
+
+### Added
+### Changed
+### Deprecated
+### Removed
+### Fixed
+### Security
+
 ## v0.14.6 -- 2024-09-25
 
 ### Added
--- a/vscode/quint-vscode/package-lock.json
+++ b/vscode/quint-vscode/package-lock.json
@@ -1,12 +1,12 @@
 {
 	"name": "quint-vscode",
-	"version": "0.14.6",
+	"version": "0.14.7",
 	"lockfileVersion": 2,
 	"requires": true,
 	"packages": {
 		"": {
 			"name": "quint-vscode",
-			"version": "0.14.6",
+			"version": "0.14.7",
 			"hasInstallScript": true,
 			"dependencies": {
 				"vscode-languageclient": "^7.0.0"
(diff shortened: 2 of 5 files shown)


lucab:
First of all, thanks for the release!
I was looking into picking up the new version from the marketplace, but the latest version I can see there is `0.14.6`: https://marketplace.visualstudio.com/items?itemName=informal.quint-vscode
Do you maybe still need to manually publish the latest release there?

bugarela:
> First of all, thanks for the release! I was looking into picking up the new version from the marketplace, but the latest version I can see there is `0.14.6`: https://marketplace.visualstudio.com/items?itemName=informal.quint-vscode Do you maybe still need to manually publish the latest release there?

Yes! I was traveling and for some reason the command to push the release was not working on my laptop and I didn't find time to fix it. I'm now back home and will push this properly.

bugarela:
Done: https://marketplace.visualstudio.com/items?itemName=informal.quint-vscode

lucab:
Thanks!
I just noticed that this one is apparently lagging behind too: https://www.npmjs.com/package/@informalsystems/quint-language-server

bugarela:
Right :facepalm:. Having this not automated is usually fine, but I guess it becomes a problem in non-ideal situations like this one :sweat_smile:. Just pushed it, should be updated in npm in a few mins.

Thanks a lot for keeping an eye on this - you avoided potentially loads of confusion in the future.

lucab:
I'm having some hard time understanding what's going on, but apparently the `informal.quint-vscode-0.14.7` extension is somehow carrying an older version of `@informalsystems/quint` in its `node_modules` directory:
```
$ jq '.version' ${HOME}/.vscode/extensions/informal.quint-vscode-0.14.7/server/node_modules/@informalsystems/quint/package.json 

"0.22.1"
```

I initially thought there was something wrong in my local setup, but then I verified that this is actually coming from the `.vsix` bundle in the marketplace.
You can check it yourself by clicking on `Download Extension` in https://marketplace.visualstudio.com/items?itemName=informal.quint-vscode and then manually opening the zip archive.

bugarela:
That's very weird! I believe there's a high chance that this is due to me preparing the release and then actually releasing from two different computers (I prepared it in my laptop, and pushed it (from the proper commit) from my home computer). Probably our script is not good enough to handle this scenario, and there are some missing commands that needed to be run (i.e. `npm install` from the release commit before pushing).

There's no way to change a release, but I'll cut a new version for VSCode tomorrow and open an issue to improve this setup. Again, thanks for checking and sorry for not testing this properly - it has been working fine until I did it a bit out of the ordinary way.

PR #1529, opened 2024-10-10
feat: include compiled binaries in releases

### LINKED ISSUE #1461: deno binary builds in releases
I have been using Quint with Deno for a while.

```console
$ deno install -A -n quint npm:@informalsystems/quint
✅ Successfully installed quint
/home/user/.deno/bin/quint
$ quint --version
0.21.0
$ # or just
$ deno run -A npm:@informalsystems/quint --version
0.21.0
```

But I wished if Quint releases could include binaries directly via `deno compile`.

But I couldn't suggest as `quint verify` was breaking because of a http2 bug. Today, I resolved it on denoland/deno#24576 which makes Quint Deno-compatible. You can now try it on quint examples.

```console
$ cd quint/examples/cosmos/ics20
$ deno run -A npm:@informalsystems/quint verify --max-steps=1 --main=bankTests --invariant=BalanceNonNegative,NonExistantAccountsOrCoinsAreZero bank.qnt
```

With this, would you consider including Deno compiled Quint binaries in future releases?

```console
$ deno compile -A -o quint npm:@informalsystems/quint
$ ./quint --version
0.21.0
```

Pros:
- No dependency to `node` or anything else.
- Standalone executables. Easily distributable.

Cons:
- Includes Deno runtime. Bigger package: 22MB (without counting the node or npm installation) vs 101MB.
- Deno is still ironing out bugs and APIs. But, Deno 2 is coming.

Refs:
- [`deno compile`, standalone executables](https://docs.deno.com/runtime/manual/tools/compiler/).
- [`pkgx` (previously `tea`) uses `deno compile` for release binaries](https://github.com/pkgxdev/pkgx/blob/ba219daa68404f4dc6206ffce43f37d5ca25386f/.github/workflows/cd.yml#L85).

rnbguy:
Closes #1461 


rnbguy:
I wonder why this workflow doesn't have [a write permission](https://github.com/softprops/action-gh-release/tree/v1/?tab=readme-ov-file#permissions) and yet `softprops/action-gh-release` can make a release. Are we using a PAT ? If yes, we should use [automatic token authentication](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication) instead.

bugarela:
This is so cool! Thanks for putting it together.

Do you have any suggestions on how to merge/test this? Is there some way to manually trigger he binary release for the latest release, or should I try it only on the next release (and then I suppose it will magically work as this workflow is triggered when I push a new tag)?

rnbguy:
It appears like the latest deno `>=2.1.0` broke something.

denoland/deno#27172

Need to wait until it's resolved or pin the deno version to `2.0.6`

rnbguy:
I fixed this in denoland/deno#27904 🙂

Quint works with the latest Deno 2.1.10.

rnbguy:
You can try dry-running the `.github/upload-binaries.sh` locally 😄

bugarela:
Ok, I was a bit confused and thought I had merged this :see_no_evil:. I just cut a new release, and now I ran the script locally to try it and it works: https://github.com/informalsystems/quint/releases/tag/v0.23.1

I tried downloading the linux binary, but have a (probably expected) NixOS errors. NixOS doesn't like whey you download binaries in general, so I don't think it's a general problem. @rnbguy Could you test it on your machine? I can test the Windows one later today as well.

```
$ chmod +x quint-linux-amd64
$ ./quint-linux-amd64 --version
Could not start dynamically linked executable: ./quint-linux-amd64
NixOS cannot run dynamically linked executables intended for generic
linux environments out of the box. For more information, see:
https://nix.dev/permalink/stub-ld
```

bugarela:
Windows also doesn't work:
```
$ .\quint-pc-amd64.exe --version

error: Uncaught (in promise) Error: Expected to resolve main module, got Import instead.
    at loadESMFromCJS (node:module:775:21)
    at Module._compile (node:module:720:12)
    at loadMaybeCjs (node:module:768:10)
    at Object.Module._extensions..js (node:module:759:12)
    at Module.load (node:module:660:32)
    at Function.Module._load (node:module:532:12)
    at Module.require (node:module:679:19)
    at require (node:module:810:16)
    at Object.<anonymous> (file:///C:/Users/xxxx/AppData/Local/Temp/deno-compile-quint-pc-amd64.exe/.deno_compile_node_modules/localhost/@informalsystems/quint/0.23.1/dist/src/cli.js:16:33)
    at Object.<anonymous> (file:///C:/Users/xxxx/AppData/Local/Temp/deno-compile-quint-pc-amd64.exe/.deno_compile_node_modules/localhost/@informalsystems/quint/0.23.1/dist/src/cli.js:361:4)
```

romac:
I get the same error on macOS:

```
❯ ./quint-macos-arm64 --version
error: Uncaught (in promise) Error: Expected to resolve main module, got Import instead.
    at loadESMFromCJS (node:module:775:21)
    at Module._compile (node:module:720:12)
    at loadMaybeCjs (node:module:768:10)
    at Object.Module._extensions..js (node:module:759:12)
    at Module.load (node:module:660:32)
    at Function.Module._load (node:module:532:12)
    at Module.require (node:module:679:19)
    at require (node:module:810:16)
    at Object.<anonymous> (file:///var/folders/6x/fbkm3r6s2td_6fx3f22w_46c0000gn/T/deno-compile-quint-macos-arm64/.deno_compile_node_modules/localhost/@informalsystems/quint/0.23.1/dist/src/cli.js:16:33)
    at Object.<anonymous> (file:///var/folders/6x/fbkm3r6s2td_6fx3f22w_46c0000gn/T/deno-compile-quint-macos-arm64/.deno_compile_node_modules/localhost/@informalsystems/quint/0.23.1/dist/src/cli.js:361:4)
```

rnbguy:
interesting. can you tell me the version of deno you're using?

```
deno run -A --node-modules-dir=auto --allow-scripts "npm:@informalsystems/quint@0.23.1" --version
```

this works fine on this workflow file https://github.com/rnbguy/quint-os/blob/main/.github/workflows/quint.yml

```
          # x64
          - ubuntu-22.04
          # arm64
          - ubuntu-22.04-arm
          # intel
          - macos-13
          # arm64
          - macos-14
          # x64
          - windows-2022
```

bugarela:
> interesting. can you tell me the version of deno you're using?

I used deno 2.1.9 to publish the releases. I saw `v2` on the CI setup and assumed this would be enough. Do you think this may be the cause of the issue?

rnbguy:
Ah there was a bug in deno@2.1.9. I fixed it in 2.1.10. Can you all try with >=2.1.10 ?

bugarela:
I updated the artifacts with the versions generated with deno 2.2.2. Can you try now? NixOS has the same issue as before, as expected. I can try on windows in a bit.

bugarela:
Works on Windows :tada:

rnbguy:
`quint-linux-amd64` works for me on linux 😄

I also added a job that runs multiple os runners with different archs to make sure deno can compile quint.

Ideally, we should run some tests using these binaries, so that we are sure everything is working fine.

rnbguy:
currently, we compile the binary from npm release. when denoland/deno#19621 is resolved, we can compile directly from a local copy.

.github/workflows/release.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         timeout-minutes: 30
         run: |
           REF_NAME=${{ github.ref_name }}
-          while ! deno run -A "npm:@informalsystems/quint@${REF_NAME#v}" --version; do
+          while ! deno run --allow-all --node-modules-dir=auto --allow-scripts "npm:@informalsystems/quint@${REF_NAME#v}" --version; do
             sleep 30
           done
```
bugarela: Would this get stuck in the loop if there is an error on a OS? Maybe we can loop hitting `npm` directly to wait for the release and then only try to run deno once

rnbguy: good point! you can use the following command to get the latest version from npm registry.

```
curl -s https://registry.npmjs.org/@informalsystems/quint | jq -r '."dist-tags".latest'
```

rnbguy: or directly query the version

```
curl -s https://registry.npmjs.org/@informalsystems/quint | jq -er '.versions."0.23.1"'
# non-zero ret-code if version not present
```

rnbguy: abf41b2

bugarela (approved):
Thanks again! I think we are fine to merge now. If there are any issues on the next release, we can open another PR to fix them.

rnbguy:
Leaving a comment on how to compile `quint` by deno directly on the repo. It still requires `npm` -- but we can come back to this later.

```
npm install
npm run compile
deno install --allow-scripts

# to run
deno run \
    --allow-all \
    --cached-only \
    --unstable-bare-node-builtins \
    --unstable-detect-cjs \
    --unstable-sloppy-imports \
    --no-check \
    dist/src/cli.js --version

# to compile
deno compile -o quint \
    --allow-all \
    --cached-only \
    --unstable-bare-node-builtins \
    --unstable-detect-cjs \
    --unstable-sloppy-imports \
    --no-check \
    dist/src/cli.js
./quint --version
```

It also makes sense to add the unstable flags inside a `deno.json`

```
// deno.json
{
  "unstable": [
    "bare-node-builtins",
    "detect-cjs",
    "node-globals",
    "sloppy-imports",
    "unsafe-proto"
  ]
}
```

PR #1547, opened 2024-11-07
Quint example for k-set consensus algorithm

kirdatatjana:
Quint Specification for Algorithm 18: K-set Consensus Algorithm in the Presence of Crash Failures

This specification is derived from the book Distributed Computing: Fundamentals, Simulations, and Advanced Topics (Second Edition) by Hagit Attiya and Jennifer Welch, specifically from Chapter 5, page 120.


josef-widder (approved):
This has been reviewed before on branches etc. All looks great.  Thanks @kirdatatjana and @bugarela  for putting effort into it.

PR #1549, opened 2024-11-11
Example: Two-layered Cache

mt40:
Continued from this thread: https://github.com/informalsystems/quint/discussions/1548#discussion-7450016

> This is a design of an imagined Two-layered Cache with the following behaviors:
> 
> - 2 layer, N keys, 1 server/cache process, 1 client process
> - read from Layer 1, if not exists then Layer 2
> - if key is found in Layer 2, write back to Layer 1
> - expire time passed = key not exist
> 
> Correctness:
> - strong consistency: any read returns the latest written value
> - validity: any read returns a non-expired value

Model checking output:
```
Launching Apalache server
[ok] No violation found (187959ms).
You may increase --max-steps.
Use --verbosity to produce more (or less) output.
Shutting down Apalache server
```


mt40:
hi @bugarela I simplified my spec a bit, inspired by the simplicity of https://github.com/mt40/azure-cosmos-tla/blob/master/general-model/cosmos_client.tla

Here are what's changed:
- Remove key-value pairs and use a monotonic increasing `num` instead. This means I'm only considering only 1 key. Since keys are independent anyway, if 1 key works, then it's trivial to see that N keys would also work.
- Removed server_process, client read/write directly from the cache layers
- Remove expired_time logic. I think that's too detail and too long for an example/learning purpose.
- Adding multiple client processes

The result is 100 LOCs stripped.

examples/tools/twolayeredcache
```diff
@@ -0,0 +1,185 @@
...
+    var history: List[HistoryEntry]
+
+    //**********************************************************
+    // FUNCTIONAL LAYER
+    // Values and functions that are state-independent
+    //**********************************************************
+
+    action writeLayer(layer: CacheLayer, v: int): bool = all {
+        layer' = layer.union(Set(v))
+    }
```
bugarela, line 40: Passing the state variable "as reference" here and assigning to a parameter is not valid Quint. I noticed Apalache is working with that, I guess because it's made for TLA+ and TLA+ allows arbitrary usage of the prime operator (`'`). In Quint, you can only use it directly with state variables, so you'd actually have to use something like this:

```suggestion
    action writeL1(v: int): bool = all {
        l1' = l1.union(Set(v))
    }

    action writeL2(v: int): bool = all {
        l2' = l2.union(Set(v))
    }
```

Using the Quint simulator (`quint run`) with the current definition results in a Runtime Error.

bugarela (commented):
Hi! Thanks again for creating this. I've made some comments, but also tweaked it a bit locally to incorporate some of these suggestions and also use sum types for the `HistoryEntry`, which I believe makes it easier to use and to read.

I added both the resulting file (after my changes) and the diff to this gist: https://gist.github.com/bugarela/5d5586abd7f6c56f71a7f05e3a92bbeb

I have 2 more high-level comments tho:
1. I don't understand how the cache layers can get out of sync in this model. Everytime we write to `l1` we also write to `l2` and vice-versa, so they will always be the same. I've tried debugging and it doesn't seem like `action handleL2Found` is ever called.
2. I saw in the other discussion that you might be running Apalache with only 2 steps. I tried running the simulator with 10 steps and it found an issue:

```
quint run two_layered_cache.qnt --invariant=safety  --max-steps=10

An example execution:

[State 0] { history: [], l1: Set(), l2: Set(), num: 0 }

[State 1] { history: [], l1: Set(), l2: Set(), num: 0 }

[State 2] { history: [Write(0)], l1: Set(0), l2: Set(0), num: 1 }

[State 3] { history: [Write(0)], l1: Set(0), l2: Set(0), num: 1 }

[State 4] { history: [Write(0)], l1: Set(0), l2: Set(0), num: 1 }

[State 5] { history: [Write(0)], l1: Set(0), l2: Set(0), num: 1 }

[State 6] { history: [Write(0), Write(1)], l1: Set(0, 1), l2: Set(0, 1), num: 2 }

[State 7] { history: [Write(0), Write(1), Write(2)], l1: Set(0, 1, 2), l2: Set(0, 1, 2), num: 3 }

[State 8] { history: [Write(0), Write(1), Write(2), Write(3)], l1: Set(0, 1, 2, 3), l2: Set(0, 1, 2, 3), num: 4 }

[State 9]
{ history: [Write(0), Write(1), Write(2), Write(3), Read(2)], l1: Set(0, 1, 2, 3), l2: Set(0, 1, 2, 3), num: 5 }

[violation] Found an issue (14ms).
Use --seed=0x16e3c92d0963bf to reproduce.
error: Invariant violated
```

mt40:
hi @bugarela I'm quite busy with job seeking lately, please don't close this PR, I will continue it afterward :) Thanks a lot for the detail review. I learned some new good things.

mt40:
hi @bugarela hope everything is well. I'm back with this PR. What's changed:
- Added all your suggestions above
- Further simplify the spec by allow only 1 value in each layer. This makes sense since my idea is to consider only 1 key since it is easy to generalize the correctness to multi-key design.
- Add a new temporal property called "agreement"

Please help take a look. Thank you.

---

I can't help but notice a lot of progress both in TLA+ and Quint. Nearly 200 commits just after a few months. Wow! Congratulations and nice work.

bugarela (commented):
Looks great! I pushed some commits with some fixes and wrote a comment about another one that requires a decision.

One important thing is that you had `agreement` as a value and was checking it as a temporal property (without it being `temporal` or having a temporal operator). Checking a simple predicate as a temporal property means checking that predicate only on the first state of the execution, which is not what we want here. I added `eventually` to that predicate and made it a proper temporal formula, which is satisfied. Is that what we need? Or should it be the case that the caches need to agree all of the time? If that's the case, than we don't need a temporal property, and this should be an invariant - and it does not hold.

bugarela (approved):
Awesome! Thanks for all the fixes, let's merge it :smile:

mt40:
Cheers! Nice working with you @bugarela

prestonph:
Hi @bugarela saw your youtube interview with MACROCOSM, it's awesome. Congrats.

bugarela:
> Hi @bugarela saw your youtube interview with MACROCOSM, it's awesome. Congrats.

Thank you! In case you are interested, I'll be doing some Quint live coding today with Chad Fowler here: https://www.youtube.com/live/ww046k3L2vY (in ~4 hours)

PR #1562, opened 2024-12-13
Counting witnesses

bugarela:
Hello :octocat: 

On recent projects, we found ourselves wanting to estimate how many of our simulations were "interesting", as in hitting interesting scenarios with specific interactions. To that purpose, I implemented a simple counter for witnesses, which I'm including here.

I see this as a start of a feature. We should have ways to dump witnesses to JSON as well, but this requires a bit more design work as I'm unsure how the interface should look like for that, and how it should interact with traces for violations.

Also, at first, I was also printing the state count (not only the trace count), but that information is harder to interpret and wasn't useful in the instances I looked at it.

This is how the coloring looks like (adding it here as it doesn't show up in the integration test):

image

### DIFF
--- a/examples/games/tictactoe/tictactoe.qnt
+++ b/examples/games/tictactoe/tictactoe.qnt
@@ -127,7 +127,7 @@ module tictactoe {
 
   action MoveX = all {
     nextTurn == X,
-    not(won(O)),
+    not(gameOver),
     if (boardEmpty) StartInCorner else
     if (canWin) Win else
     if (canBlock) Block else
@@ -139,7 +139,7 @@ module tictactoe {
 
   action MoveO = all {
     nextTurn == O,
-    not(won(X)),
+    not(gameOver),
     MoveToEmpty(O),
     nextTurn' = X,
   }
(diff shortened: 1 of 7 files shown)


ivan-gavran:
I would like to see one more thing reported with witnesses, and that is how many traces were inspected in total. 
If the invariant of a run is not violated, then we control it with `max-samples`. But if it is violated, we don't know how many traces were run, and this is a useful info.

(An alternative solution to the same problem may be to display the info on traces run before the violation, but I still prefer to have explicit info on witnessed in `x` cases out of total `y` cases.)

ivan-gavran (changes requested):
Overall looks good to me. 
I'd suggest adding [the feature](https://github.com/informalsystems/quint/pull/1562#issuecomment-2557065765) that shows total number of explored traces.

bugarela:
Good point, I haven't thought of that. I should also add a test case for printing witnesses in violations.

Not sure if I'll have time to pick this up before the holidays and I'll be on vacation for almost a month, so if anyone wants to work on this in the meantime, feel free.

bugarela:
@ivan-gavran WDYT?
image

ivan-gavran (approved):

Issue #1552, opened 2024-11-14
Simulator wastes a lot of resources computing states that won't be used


bugarela:
When there is an `any`, Quint simulator will do the following:
1. Execute all actions and save the result and the state after that action
2. Filter out actions for which the result is false (i.e. pre-conditions not satisfied)
3. From the remaining actions, randomly pick one
4. Take the state from the picked action to be the next state in the simulation and proceed

This means that we can compute a bunch of stuff in 1 that will never be used.

I need to think if this won't affect the probability distribution too much, but an alternative approach would be:
1. pick an action
2. execute it. If the result is false, try the next one
3. if we get back to the picked action, the result of `any` is false

This would also mean we would make a `rand()` call even on deterministic `any`s (that is, with disjoint pre-conditions). But I think is would be a much smaller problem than the performance issue.

jff-work:
Hey,

I looked into this issue, implemented the other solution, ran some tests and reasoned about the probabilities:
https://github.com/jff-work/quint-any-test

In case you like the solution there's a pull request:
https://github.com/informalsystems/quint/pull/1582

Happy Holidays,
Julian

bugarela:
Closed by https://github.com/informalsystems/quint/pull/1582

bugarela:
Closed by https://github.com/informalsystems/quint/pull/1582

PR #1582, opened 2024-12-29
Changed Any Operator to be less Wasteful

jff-work:
As suggested in issue:
https://github.com/informalsystems/quint/issues/1552

The Any operator in Quint was evaluating a lot of statements in cases that could be skipped. I implemented the proposed solution and wrote some tests as well as reasoned about the probabilities of different statements being returned by Any, which remain the same for this implementation.
https://github.com/jff-work/quint-any-test

I'd probably run some more complicated tests before merging this.


bugarela:
Hi! Thanks a lot for addressing this and documenting everything so well. I'll do some testing today - the failing tests don't mean there's anything wrong, but this does (as expected) change the way `rand()` is used, and therefore runs with a fixed `seed` will have different results under this new implementation. I just need to update the seeds in the failing tests.

Sorry for the delayed review, I was on vacation for a while.

bugarela:
I just tried this change on the (private) project where I noticed the slowdown caused by any and, for this instance, this change is a 2x speed up :tada:

bugarela (approved):
Looks great! I've updated the integration tests that relied on the seed and tested this on a couple of real projects, all seems great and the performance improved so much!

PR #1601, opened 2025-03-07
Adjust params to enable finding violation on K set

bugarela:
Hello :octocat: 

This fixes the comments and init definition to make it feasible to find violations in simulation.


josef-widder (approved):
It is very nice that we can now find the violates with simulations. Thanks for having taken another look.

Issue #1424, opened 2024-04-09
Integration with TLC


bugarela:
I've been thinking about and trying to use Quint with TLC for a while now, so here's an issue to track these thoughts/experiments.

First of all, we need to translate Quint to TLA+. We almost have that, with two pending issues:
- https://github.com/informalsystems/apalache/issues/2875
- https://github.com/informalsystems/apalache/issues/2863

This is done by
```
quint compile --target tlaplus file.qnt
```

Since this command will wrap the init, step and inv definitions, the `.cfg` file should always be something like
```
INVARIANTS
q_inv

INIT
q_init

NEXT
q_step
```

We can produce that automatically in Quint.

Then, we can invoke TLC pointing the lib path to Apalache libs, since the generated TLA+ module will depend on them. I found that the easiest way to do this is by using the TLC inside of `apalache.jar`, which is already automatically dowloaded by Quint in the `compile` command (which depends on Apalache).
```
java -Xmx4096m -cp ~/.quint/apalache-dist-0.44.10/apalache/lib/apalache.jar tlc2.TLC file.tla
```

However, the counterexample printed by TLC is hard to read, since it refers to things in the generated TLA+ spec, which don't map directly to the original Quint spec. For example, see this state from a tic tac toe spec:
```
State 6: <MoveX line 166, col 3 to line 176, col 23 of module tictactoe>
/\ board = << << [tag |-> "Occupied", value |-> [tag |-> "X", value |-> "U_OF_UNIT"]],
      [tag |-> "Occupied", value |-> [tag |-> "O", value |-> "U_OF_UNIT"]],
      [tag |-> "Occupied", value |-> [tag |-> "O", value |-> "U_OF_UNIT"]] >>,
   << [tag |-> "Empty", value |-> "U_OF_UNIT"],
      [tag |-> "Occupied", value |-> [tag |-> "X", value |-> "U_OF_UNIT"]],
      [tag |-> "Empty", value |-> "U_OF_UNIT"] >>,
   << [tag |-> "Empty", value |-> "U_OF_UNIT"],
      [tag |-> "Empty", value |-> "U_OF_UNIT"],
      [tag |-> "Occupied", value |-> [tag |-> "X", value |-> "U_OF_UNIT"]] >> >>
/\ nextTurn = [tag |-> "O", value |-> "U_OF_UNIT"]
```

While a similar trace state in quint is simply:
```
[State 5]
{
  board:
    Map(
[...]

PR #1613, opened 2025-03-14
Convert init into predicates when compiling to TLA+

bugarela:
Hello :octocat: 

Closes https://github.com/apalache-mc/apalache/issues/2863. Towards https://github.com/informalsystems/quint/issues/1424

This fixes an old issue where we failed to properly transpile Quint into TLA+ as Quint init definitions are actions with assignments, while TLA+ expects predicates with equalities for the initial state.

The strategy is:
1. Collect all definition names that are called from `q::init`, recursively
2. Iterate again, now replacing assignments whenever they are found inside of those definitions
3. In this second iteration, if this definitions are ever found also outside of the `q::init` context (being referred to by names outside of the tracked scope), we check if they have assignments (mode is `action`) and, if so, report an error.
  - If we update this type of action, we would break the other usages inside non-init actions.
  - This should be quite an edge case, and it would be complicated to handle, so I prefer to report an informative error and request a manual fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 ### Removed
 ### Fixed
+
+- Init definitions are now properly transpiled into TLA+ without assignments (#1613)
+
 ### Security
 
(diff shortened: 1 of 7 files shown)


romac (approved):
Gave it a go on a few handwritten specs, both the code and the results look good!

PR #1623, opened 2025-04-01
Integration with the Rust backend :tada: 

bugarela:
Hello :octocat: 

This adds integration with our Rust backend (to be released and open-sourced next week at the [Quint Launch Party](https://bit.ly/quint-launch)). This is a v0 and it's quite simple: spawn a synchronous process, writing the simulation parameters and parsed data to STDIN and then reading the outcome from STDOUT.

The backend can be called with `--backend=rust` on `quint run`. Typescript is still the default, for now.

This also includes some helpful functionality: the `--flatten` flag, which enables compiling to JSON without flattening, and the `main` field in the output, to propagate the information regarding which module should be treated as main. This are not necessary for the integration, but the Rust simulator uses it for its own testing/benchmarking functionality.

The integration is not tested yet as we won't have the binaries in the CI at this point. We should include tests soon.


docs/pages/docs/quint.md
```diff
@@ -85,12 +85,14 @@ Options:
   --temporal   the temporal properties to check, separated by commas    [string]
   --target     the compilation target. Supported values: tlaplus, json
                                                       [string] [default: "json"]
+  --flatten           Whether or not to flatten the modules into one
```
romac, line 88: Since the flag is on by default, we want might to show how disable it explicitly in its help message:

```suggestion
  --flatten=[true|false]           Whether or not to flatten the modules into one
```

I know it is also explained below, but it can't hurt to have it where the flag is defined as well.

bugarela: I couldn't make `yargs` print the `--help` text like that, but did this instead:

```
  --flatten           Whether or not to flatten the modules into one. Use
                      --flatten=false to disable       [boolean] [default: true]
```

Also, I realize we could use `choices` for stuff like `--backend` and `--target`, so I changed that and removed our manual checks. Error messages now look like this:
```
$ quint compile --target tla file.qnt
[...]
Invalid values:
  Argument: target, Given: "tla", Choices: "tlaplus", "json"
```

I didn't used `choices` for `flatten` cause it would print something silly like `[boolean] [choices: true, false]`

see-quick:
And I was going to ask why you guys don't have Quint written in Rust (as you have many projects in Rust) instead of Typescript. D. I wonder how this improves the simulator performance-wise, but anyway, this is really a great way forward! <3

bugarela:
> And I was going to ask why you guys don't have Quint written in Rust (as you have many projects in Rust) instead of Typescript. D. I wonder how this improves the simulator performance-wise, but anyway, this is really a great way forward! <3

We chose Typescript initially in hopes to have a lower barrier for contributors, but it really feels like Rust is the best choice now, specially for the simulator, where performance matters a lot. I wrote about it in this [blog post](https://informal.systems/blog/quint-deserves-rust).

We are releasing the Rust backend next week at the launch (which I see you already registered for!).

bugarela:
I don't want to change this too much as it's hard to test it at this point, but one thing for the near future: since we are getting a pinned release, we could just download the specific asset we need directly without first checking if the version exists and then if the asset exists as we do now. Like we do for Apalache:
https://github.com/informalsystems/quint/blob/ffc7a5c8a0bb87b13e181dca5619785e8e345c6e/quint/src/apalache.ts#L339

I got some GitHub API rate limit [issues](https://github.com/informalsystems/quint/pull/1626) when I tried downloading assets in the CI and **I think** this can be related - I remembered we had issues with Apalache and then changed to this, but I don't remember the details. Besides, I'm not sure whether we download apalache via Typescript on the CI at this moment (seems like we wanted to avoid it, but the examples dashboard might be doing it).

romac:
> since we are getting a pinned release, we could just download the specific asset we need directly without first checking if the version exists and then if the asset exists as we do now. 

Makes sense, happy to do the change here already if you want.

romac (approved):

quint/src/quintRustWrapper.ts
```diff
@@ -0,0 +1,384 @@
...
+import { LookupTable } from './names/base'
+import { replacer } from './jsonHelper'
+import { ofItf } from './itf'
+const spawn = require('cross-spawn')
+
+import path from 'path'
+import os from 'os'
+import chalk from 'chalk'
+
+const QUINT_EVALUATOR_VERSION = 'v0.0.0-pre.2'
```
bugarela, line 27:
```suggestion
const QUINT_EVALUATOR_VERSION = 'v0.1.0'
```

PR #1624, opened 2025-04-03
spell: add `min` alongside existing `max`

bengesoff:
I know it's supposed to go into the `rareSpells` first, but given that there's already a `max` one in `basicSpells`, I figured it would better for them to go together.


bugarela (approved):
Thank you! This definitely belong in `basicSpells` :smile:

PR #1627, opened 2025-04-05
:crab: Migrate the Rust evaluator from private repo

bugarela:
Hello :octocat: 

We started exploring a Rust evaluator for Quint in a private repository a couple of months ago, and now we reached a usable version of it, so we are moving things from there into this official repository :tada: 

Read more about the motivations for this work here: https://informal.systems/blog/quint-deserves-rust

This will be first deployed with a feature flag (`--backed=rust`) and should become the default once we have full feature parity and more battle-testing. If everything goes according to plan, we will deprecate the current simulator. Users should not be affected at all except for the speed bump :smile: 

## For Review

All commits were imported from the old repo and previously reviewed, except for:
- the last PR on docstrings, which I merged without review: https://github.com/informalsystems/quint-simulator/pull/48 (private)
- the last commits on this, fixing the CI and renaming `quint_simulator` to `quint_evaluator`. Review here: https://github.com/informalsystems/quint/compare/6db3bc0..3f24873


bugarela:
I'm not managing to make the Link workflow work, it seems to be ignoring my attempts to set the `working_directory` and therefore fail to find `Cargo.toml`.

Also, the Benchmark workflow is not running and I don't know why.

I want to update the existing typescript-related workflows to avoid triggering them on Rust-only changes. I should also separate vscode (language server) stuff from the regular Quint stuff.

On top of fixing the workflows, we'll need to update #1623 to fetch assets from this repo instead. I'm not sure about how to configure the release here. I removed the "on tag push` trigger of the workflow until we figure out a good way of doing it, and kept the manual trigger which will publish under tag `evaluator-vx.x.x`. I'd also like to use the name "quint evaluator" instead of "quint simulator" for the binaries as those will also be used for testing and the REPL in the future (not only simulation).

bugarela:
Found out why the benchmark workflow is not running
image

bugarela:
Ok, this should be all good now. I'll switch the required checks to the new names once we are ready to merge. 

I'll now work on updating https://github.com/informalsystems/quint/pull/1623. We will need to merge this into main to be able to cut a release and properly test that one, but we can do that on Monday probably.

bugarela:
For the releases, I'm not sure if we want the tags to look like `evaluator-v0.0.0` or `evaluator/v0.0.0`

romac:
> For the releases, I'm not sure if we want the tags to look like `evaluator-v0.0.0` or `evaluator/v0.0.0`

Fwiw the convention I see most often with multi-projects repo is to use a slash as a delimiter, but both work.

romac (approved):

Issue #1609, opened 2025-03-12
Apalache server doesn't start with `quint verify` sometimes


bugarela:
I don't know the cause.

We should make sure the server starts and, if not, at least print and informative error. Currently it just hangs.

Issue #1501, opened 2024-09-09
verify: surprising temporal result


lucab:
I was trying to verify a (state-machine plus math) model with some basic temporal props and I think I ran into a surprising false-negative result (likely a shortcoming in Apalache itself, I guess).

After deleting all the extra details, this is the minimal reproducer I have:

```
module buggyTemporals {
    var myState: int

    action init = myState' = 0

    action step = {
        myState' = myState + 1

        // This other logic instead gets a proper counterexample:
        // myState' = 1
    }

    temporal temporalProps = always(myState < 0)

    // Same result for:    
    // temporal temporalProps = eventually(myState < 0)
}
```

```
$ quint verify --temporal temporalProps buggyTemporals.qnt

[ok] No violation found (179ms).
You may increase --max-steps.
Use --verbosity to produce more (or less) output.
```

Just for completeness, increasing the number of steps does not help finding a counterexample.

I tried tweaking slightly the example above in this way:
```
    action step = {
        if (myState > 5) {
            myState' = myState
        } else {
            myState' = myState + 1
        }
    }

```

In this other case, Apalache detects that it eventually enters a loop where `temporalProps` is not valid. However, the counterexample shows that it was already not valid in `State 0`:
```
[State 0]
{
  __InLoop: false,
  [...]
  myState: 0
}

[...]

[State 6]
{
  __InLoop: false,
  [...]
  myState: 6
}

[State 7]
{
  __InLoop: true,
  [...]
  myState: 6
}

```

While I can understand that `eventually()` may have some troubles without a cap on the counter, I'm more concerned about the verification result of the `always()` property which looks like a verifier bug.

bugarela:
Thanks for reporting! I've actually come across some examples where temporal property verification is flaky in apalache. For example, it reports [ok] for this:
```
module test {
  var x: int

  action init = x' = 0
  action step = x' = x + 1

  temporal prop = always(next(x) < x)
}
```

I think the plan here is:
1. Check whether this problems also occur with TLA+ (using Apalache direcly)
  - If not, then there should be a problem in the translation from Quint to TLA+ (unlikely), and we can fix the translation
2. If yes, then
  - Open an issue in Apalache
  - #1424 so people can use TLC for temporal properties instead.

[...]

PR #1628, opened 2025-04-07
Apalache UX improvements

bugarela:
Hello :octocat: 

## Temporal property counterexamples

When checking temporal properties with Apalache, we print all variables returned by it, which includes some that are hard to decode:
```
{
  __InLoop: true,
  __q::temporalProps_init: false,
  __saved___temporal_t_1: false,
  __saved___temporal_t_2: false,
  __saved_offers: Set(),
  __saved_foo: Map(),
  __temporal_t_1: false,
  __temporal_t_2: false,
  __temporal_t_2_unroll: false,
  __temporal_t_2_unroll_prev: false,
  bar: Set(),
  foo: Map()
}
```

Since Quint doesn't have a UI to explain temporal traces (which often include loops i.e. loop starts here, back to state X), I guess we were motivated to keep this variables at first. However, I think they hurt more than they help, and might not be 100% accurate - for instance, this example is not supposed to be a loop, but `__InLoop` is true.

Once we have integration with TLC (#1424), we should strongly recommend or enforce that TLC is used for temporal properties instead of Apalache (see https://github.com/informalsystems/quint/issues/1501 for more context)

My solution was to remove all variables starting with `__` when parsing ITF. It might be a good idea to treat variables starting with `__` as internal and not show them in traces in general, not only for Apalache. That's for later.

## Hanging issue

Fixes https://github.com/informalsystems/quint/issues/1609

Myself and some other people have issues with `quint verify`, where it hangs if there is no server running already (instead of starting the server as expected). This is very annoying and makes the tool seem unreliable. I debugged a bit and found which gRPC call was hanging, and setting a deadline (timeout) for it fixed the problem for me. A more general timeout solution didn't work, as this gRPC call cannot be cancelled from the Node side.


romac (approved):
Nice!

PR #1630, opened 2025-04-07
Prevent init errors from being reported for runs

bugarela:
Hello :octocat: 

We recently added the conversion of `init` actions into TLA+ predicates when transpiling. This would result in an error in the rare scenario where `init` is used in an action context (other than the one being converted - for example, inside `step`).

I now realize this "rare scenario" happens often in runs. However, runs are discarded by Apalache anyway and there's no translation of runs into TLA+. So what I'm doing now is just dicarding the runs before we go into this init conversion step. This happens only on transpilation to TLA+.

### DIFF
--- a/quint/testFixture/ApalacheCompilation.qnt
+++ b/quint/testFixture/ApalacheCompilation.qnt
@@ -28,7 +28,6 @@ module ApalacheCompilation {
   // Tests whether we will sanitize identifiers
   def foo::bar(__123) = __123
 
-  // TODO: Tests that we remove primes from assignments in the init function
   action init = {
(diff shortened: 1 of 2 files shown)


quint/testFixture/ApalacheCompilation.qnt
```diff
@@ -47,4 +46,6 @@ module ApalacheCompilation {
 
   // Tests that we can specify an alternative invariant via CLI args
   def altInv = x >= 0
+
+  run test = init.then(step)
```
bugarela, line 50: This works as a regression test for this change

romac (approved):

PR #1643, opened 2025-04-10
Use paths-filter for workflows

bugarela:
Hello :octocat: 

This is an attempt to fix the current problem on our CI where all checks are required, but not all of them are run in all PRs due to the new filters.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,11 +1,12 @@
 # Quint GitHub CI Workflows
 
-## [./main.yml](./main.yml)
+This folder contains the Continuous Integration (CI) workflows for the Quint project. Most of them run on PRs and pushes to the main branch, except for the release workflows described below, and for the benchmark workflow, whose output is a PR comment and therefore only runs on PRs.
 
-- Triggered on pull requests into `main`.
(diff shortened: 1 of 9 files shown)


bugarela:
Two notes:
1. Not sure if this will work on the matrix-os workflows. I had to change the requirements for those from "Unit tests (ubuntu_latest)" to "Unit Tests" in order for it to be happy with the skipped workflow name. But I'm not sure if this will detect the flows that actually run.
2. Not sure if I should add the workflow itself to the filter, as I had done before with the triggers. My reasoning is that we want to run the workflow when we change something about it, to make sure that it works.

romac (approved):
Looks good! Yes these string-based filtering of required checks is really annoying, but I think we should indeed mark the workflows as required and the path filter step will make sure to mark the workflow as successful if the target paths haven't changed.

bugarela:
Looking at the other PRs, the matrix was in deed a problem - I'm trying something different with aggregators

bugarela:
Works now: https://github.com/informalsystems/quint/pull/1647
The only problem is the enormous list of checks :sweat:. But I guess it's fine.

PR #1651, opened 2025-05-14
Improve interface of `check_with_tlc.sh`

bugarela:
Hello :octocat: 

This improves the TLC script interface a bit, namely:
1. Taking the file as a positional argument instead of under a flag
2. Taking the invariant as an expression, as in `quint run`
3. Using `--workers=auto` as default
4. Taking init and step optionally, as well as `main`, which then produces a `tla` and `cfg` file with that main's name so TLC won't complain the module doesn't match.


see-quick:
> @see-quick would you like to take a look, as you made most of the work in this script so far? Can't tag you as a reviewer since you are not a maintainer, so just making a comment here 😬

Yeah, sure, I will take a look at this tomorrow. Thanks.

romac:
Could this be built into Quint as a command, similar to have we have `verify` for Apalache?

see-quick (approved):
LGTM 👍

bugarela:
> Could this be built into Quint as a command, similar to have we have `verify` for Apalache?

Yes, the plan is that this becomes `quint verify --backend=tlc`. This script is the MVP for that :grimacing:

Issue #1496, opened 2024-09-02
Show test results as soon as we finish running a single test


bugarela:
Right now we compile all results to show at the end, but it would be better for usability if we show whether a single test has succeeded/failed before starting to process another test.

jackcooper20:
Hi @bugarela! Can you kindly check out my attempt at solving this issue? Please let me know if I need to improve something.

PR #1657, opened 2025-05-15
Show every test result as soon as they finish running

jackcooper20:
**Description**
Implements immediate feedback for individual tests as they complete, instead of waiting until the entire suite has finished. When verbosity is enabled, each test’s pass/fail status (and sample count) is printed instantly.

**What Changed**

* Wrapped the test loop so that after every call to `evaluator.test()` we:

  * Push the `TestResult` into `results`
  * Immediately `out()` a green “ok” message for passes or a red numbered “failed” message for failures
  * Increment the failure counter for numbering failed cases
* Retained timing around the whole run (`startMs`/`elapsedMs`) for final summary or diagnostics
* No change to test collection logic or tracing—only the display timing

Addresses issue #1496


bugarela (approved):
Thank you, this is a great improvement! It is still not 100% perfect as the actual errors (explaining what broke) are shown only at the end. However, these errors need a lot of improvement on their own, so we can address that later. This is already much better than what we have currently.

PR #1660, opened 2025-05-19
feat: Support single-argument q::debug calls via macro-like transformation

### LINKED ISSUE #1567: Calling `q::debug` with a single argument should work
Currently, if I want to add debug statements to an expression like:
```bluespec
foo(bar(1 + 2))
```

I need to do something like:
```bluespec
q::debug("foo", foo(q::debug("bar", bar(q::debug("plus", 1 + 2))))))
```

I'd like to also be able to write:
```bluespec
q::debug(foo(q::debug(bar(q::debug(1 + 2))))))
```

And have Quint print the expressions:
```bluespec
> foo(bar(1 + 2)): 45
> bar(1 + 2): 10
> 1 + 2: 3
```

Moreover (not this issue), we can also consider having something like:
```bluespec
q::deepDebug(foo(bar(1 + 2)))
```

halcyonet:
Fixes #1567  

This PR adds support for calling `q::debug` with a single argument by implementing a macro-like transformation in the parser. When `q::debug` is called with one argument, the parser automatically transforms it to include the expression text as the first argument.

Changes:
- Added transformation in `ToIrListener.ts` for single-argument calls
- Updated documentation in `builtin.qnt`
- Added test case in `evaluation_tests.rs`

### DIFF
--- a/evaluator/tests/evaluation_tests.rs
+++ b/evaluator/tests/evaluation_tests.rs
@@ -1203,3 +1203,14 @@ fn run_q_debug() -> Result<(), Box<dyn std::error::Error>> {
          run run1 = (n' = 1).then(n' = q::debug(\"n plus one\", n + 1))",
     )
 }
+
+#[test]
+fn run_q_debug_single_arg() -> Result<(), Box<dyn std::error::Error>> {
+    assert_var_after_run(
+        "n",
+        "2",
+        "run1",
+        "var n: int\n
+         run run1 = (n' = 1).then(n' = q::debug(n + 1))",
+    )
+}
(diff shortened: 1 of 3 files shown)


halcyonet:
Hi @bugarela . I have send a PR #1660 . Could you kindly review and let me know?

bugarela (approved):
Awesome! Could you please also add an entry to `CHANGELOG.md` so people now this feature is available now?

halcyonet:
@bugarela Done!!! Could you kindly check and merge if everything looks correct.

halcyonet:
@bugarela I have made the changes. Please check and let me know.

bugarela:
Interesting, the error in CI is because of a limitation of the GitHub API in forks:
> Note: In public repositories this action does not work in pull_request workflows when triggered by forks. Any attempt will be met with the error, Resource not accessible by integration. This is due to token restrictions put in place by GitHub Actions. 

https://github.com/peter-evans/create-or-update-comment

I'll bypass and merge

PR #1669, opened 2025-05-26
`quint run`: report information on trace length and speed

bugarela:
Hello :octocat: 

This adds extra report info to the `quint run` command, including speed (which is now available in the progress bar, but you can't see it after the command finishes, and I found that annoying) and trace length.

The trace length is an attempt to make it more explicit when the simulator fails to expand a trace further. I'm currently working on improvements on the simulation in this regard, and this makes it more visible. Ideally, in most cases, min = max = avg, but that's not the reality right now since the simulator "gives up" on traces to easily. See this example for the tendermint spec:

```sh
quint run ../examples/cosmos/tendermint/TendermintModels.qnt --invariant=n5_f2::AgreementOrAmnesia --init n5_f2::Init --step n5_f2::Next --max-steps=50
```


romac (approved):
This is great!

PR #1682, opened 2025-06-03
Add diagnostic for incorrect map type syntax

### LINKED ISSUE #1676: Suggest the correct syntax `a -> b` when people write `Map[a, b]`
Since set types are `Set[elem]` and lists are `List[elem]`, people (and AI tools) often try to write map types as `Map[key, elem]`, but that is not correct. I think a simple solution would be to parse `Map[a, b]` and then flag an error with an LSP quick fix (see example below)

https://github.com/informalsystems/quint/blob/13d07c2f8bab34cbded0bc68547bc28ba02fe678/vscode/quint-vscode/server/src/server.ts#L255-L274

cordencodes:
Fixes #1676
…ntax

This PR adds a diagnostic warning that detects when users write map types using the `Map[key, elem]` syntax and suggests using the arrow syntax `key -> elem` instead.

- Added a new diagnostic check in the language server that detects `Map[key, elem]` syntax
- Implemented a quick fix action that automatically converts the syntax to the correct arrow form
- Added warning messages that explain the correct syntax to use


vscode/quint-vscode/server/src/server.ts
```diff
@@ -520,6 +527,50 @@ export class QuintLanguageServer {
...
+      const startPos = document.positionAt(match.index)
+      const endPos = document.positionAt(match.index + match[0].length)
+      
+      const diagnostic: Diagnostic = {
+        severity: DiagnosticSeverity.Warning,
+        range: {
+          start: startPos,
+          end: endPos
+        },
+        message: `Use '${keyType} -> ${valueType}' instead of 'Map[${keyType}, ${valueType}]' for map types in Quint.`,
```
bugarela: This is ultimately what we want, but we also this suggestion to appear somehow on the CLI. We should identify this error on the parsing side and report it with a `QuintError` containing the fix info as a `QuintErrorFix` inside the `data` field so it gets picked up by https://github.com/informalsystems/quint/blob/13d07c2f8bab34cbded0bc68547bc28ba02fe678/vscode/quint-vscode/server/src/server.ts#L222 and then produces this diagnostic here automatically.

This ensures users who are only in the CLI have the same experience, as it works, for example, in Rust.

test_files/map_syntax.qnt
```diff
@@ -0,0 +1,12 @@
+module test {
```
bugarela: By having the fix suggested on the CLI as well, we can add an integration test for this file. We should also move it to `testFixtures` (instead of creating a new folder called `test_files`)

cordencodes:
@bugarela Please review when you get a chance.

bugarela:
I think the reason it doesn't work is this line
https://github.com/cordencodes/quint/blob/631ebdce8b061f2695b022dfab1ec6229db7db43/quint/src/generated/Quint.g4#L290

`Map` is a token, while it shouldn't be. Fixing this problem and running `quint run antlr` to regenerate the parsing files before running `npm run compile` again makes it work:
```
$ quint run ../local/test.qnt
/home/gabriela/projects/quint/local/test.qnt:3:19 - error: [QNT000] Use 'int -> str' instead of 'Map[int, str]' for map types
3:   type UserData = Map[int, str]
                     ^^^^^^^^^^^^^

/home/gabriela/projects/quint/local/test.qnt:6:16 - error: [QNT000] Use 'str -> int' instead of 'Map[str, int]' for map types
6:   val userMap: Map[str, int] = Map()
                  ^^^^^^^^^^^^^

error: parsing failed
```

I'll commit some changes

PR #1689, opened 2025-06-11
Refactor code to replace lodash.isequal with node:util's  with  isDeepStrictEqual 

### LINKED ISSUE #1606: Fix lodash isEqual deprecation
Using node > 20 to install Quint, we se the following warning:

```
npm warn deprecated lodash.isequal@4.5.0: This package is deprecated. Use require('node:util').isDeepStrictEqual instead
```

aleskrin:
- Closes #1606 


aleskrin:
hey @bugarela , could you please take a look at this pr when you're free?

src/apalache.ts
bugarela, line 1: ?

package-lock.json
bugarela, line 1: ?

aleskrin: @bugarela ,

removing `lodash.isequal` from `package-lock.json` is the final instruction that stops npm from installing the deprecated package. When any developer runs `npm install` with our **updated** `package-lock.json`, npm sees that `lodash.isequal` is no longer a required dependency and skips it entirely. No installation means no deprecation warning, which resolves the issue.

src/apalache.ts
aleskrin: @bugarela 

The deprecation of `lodash.isequal` is just one sign that we were relying on older patterns. The updates in `apalache.ts` fix other subtle issues that come up in modern Node.js environments.For example, recent Node versions use Web Streams for the fetch API, but our code expected a Node.js Stream. Wrapping the response body with `Readable.fromWeb(res.body)` fixes this mismatch and prevents download failures when fetching Apalache.

I also added an explicit cast to `Uint8Array` for byte data. This helps avoid potential type errors, since different library versions can handle data formats a bit differently. I think, together these changes make Quint more robust and ready for up-to-date dev and CI setups.

package-lock.json
bugarela: You created a new file.

src/apalache.ts
bugarela: You created a new file.

aleskrin:
@bugarela, Thanks for pointing that out. package-lock.json was generated because I mistakenly ran   npm i outside quint in the root directory. I generally ran codeRabbit locally for code review which actually created the file  apalache.ts . Its my bad, I should have been more careful about this.

aleskrin:
Hi @bugarela,

I’ve cleaned up the files you flagged and  pushed the latest commit.
Please let me know if there are any further adjustments needed.
Thank you for your thorough review.

quint/src/apalache.ts
```diff
@@ -292,7 +292,10 @@ async function loadProtoDefViaReflection(
 
       // Decode reflection response to FileDescriptorProto
       let fileDescriptorProtos = protoDefResponse.file_descriptor_response.file_descriptor_proto.map(
-        bytes => protobufDescriptor.FileDescriptorProto.decode(bytes) as protobufDescriptor.IFileDescriptorProto
+        bytes =>
+          protobufDescriptor.FileDescriptorProto.decode(
+            bytes as unknown as Uint8Array
+          ) as protobufDescriptor.IFileDescriptorProto
```
romac, line 298: Why are these changes needed?

quint/src/apalache.ts
```diff
@@ -340,7 +343,11 @@ function downloadAndUnpackApalache(apalacheVersion: string): Promise<ApalacheRes
   return fetch(url)
     .then(
       // unpack response body
-      res => pipeline(res.body!, tar.extract({ cwd: apalacheDistDir(apalacheVersion), strict: true })),
+      res =>
+        pipeline(
+          Readable.fromWeb(res.body! as any),
+          tar.extract({ cwd: apalacheDistDir(apalacheVersion), strict: true })
+        ),
```
romac, line 350: Same here?

quint/src/apalache.ts
aleskrin: @romac   Oho! I think , that would not be necessary. I manually checked some type casting over there. Fixing it on next commit.

quint/src/apalache.ts
aleskrin: Same here. Fixing it on next commit.

bugarela:
This has become heavily outdated with my recently merged refactors. Plus, the amount of removed code concerns me, and I'd prefer to make the mechanical "find and replace" to avoid any unwanted changes. Feel free to re-submit the PR with only the mechanical changes and no extra hidden modifications.

PR #1696, opened 2025-06-17
Towards a more permisse grammar

### LINKED ISSUE #1688: Unclear error message when record keys are keywords
Hi there. I'm really enjoying playing around with quint so far. Thanks for making it!

I wasn't aware that `from` was a quint keyword, and so used it as a field name in a record like below. 

```quint
module test_typechecker {
  type Bad = {
    from: int
  }
}
```
When type checking, quint gives a very unhelpful error message. 

```
ajs@meems:~/quint-simple> quint typecheck keyword-fail.qnt
/home/ajs/quint-simple/keyword-fail.qnt:4:5 - error: [QNT000] mismatched input 'from' expecting {'}', '|', LOW_ID, CAP_ID}
4:     from: int
       ^^^^

/home/ajs/quint-simple/keyword-fail.qnt:7:1 - error: [QNT000] extraneous input '}' expecting {<EOF>, 'module', DOCCOMMENT}
7: }
   ^

error: parsing failed
```

It would be great if quint could report that `from` is a keyword instead. Documenting` keywords in the language manual would also be helpful. Thanks!

bugarela:
Hello :octocat: 

This are the first few changes towards a more permissive grammar. This was a point of discussion in the past, and after discussing this with other people from outside the initial circle and thinking more about it, I decided to go with a more permissive grammar, where we parse wrong formats sometimes in order to
1. recover from errors (and report more errors at the same time to reduce the number of iterations it takes to correct all errors)
2. report better error messages

In this PR, I'm making two initial changes (that are related in a subtle way):
1. When someone writes type application with parenthesis instead of square brackets (i.e. `List(int)` instead of `List[int]`, we'll report a better error.
2. Some previously reserved keywords can now be used as identifiers, and those who cant will be properly reported. This closes #1688. I chose `from` and `as` since I found myself trying to use those as identifiers before, and also `List` and `Set` because I think the error message `error: [QNT101] Built-in name 'List' is redefined in module X` makes more sense than `List is a reserved keyword`. 


andrewjstone:
Thank you so much @bugarela. My sample module now typechecks. Also, having `from` available as an identifier was an unexpected gift! You went above and beyond.

PR #1706, opened 2025-07-14
[website] MBT page

bugarela:
- [ ] I have read and I understand the [Note on AI-assisted contributions](/CONTRIBUTING.md#note-on-ai-assisted-contributions)

### DIFF
--- a/docs/pages/docs/model-based-testing.mdx
+++ b/docs/pages/docs/model-based-testing.mdx
@@ -38,13 +38,176 @@ Let's use a toy bank application written in Rust to go through the process. Code
 
 <Steps>
 ### Generate traces from the model
+
+We can create a folder called `traces` to hold all of our trace files an then generate as many as we want using `--n-traces`:
+```sh
+quint run bank.qnt --mbt --n-traces=10000 --out-itf=traces/out.itf.json
+```
+
+Explaining the command:
+- `quint run` is the subcommand for the Quint simulator
+- `bank.qnt` is the file holding the model for which we want to generate traces from
+- `--mbt` says we want extra info in the traces that are very useful for model based testing
+  - `mbt::actionTaken` which will hold which action was used to reach the given state
+  - `mbt::nondetPicks` which will hold what value was chosen for each relevant non-deterministic choice leading to that state
+- `--n-traces` determines how many traces the simulator will hold in memory. It will hold the `N` longest traces in case there is no violation, or the shortest ones if they represent violations to given properties (not the case in this example).
+  - Note: if you want more than 10k traces, you need to also increase `--max-samples`, as 10k is the default for that.
+  - Troubleshooting: If you run out of node memory while running this, you can increase it with `NODE_OPTIONS="--max-old-space-size=8192"`. Shouldn't be the case for 10k on modern computers.
+- `--out-itf` specifies that we want to dump all those traces to JSON, and where to put them. 
+
+This will create files from `traces/out0.itf.json` to `traces/out9999.itf.json` with a bunch of executions that our model allows.
+
+
 ### Parse the traces in the test driver
+
+Now we need to get all that data into our implementation language structures, so we can use them in tests. In this example, this is Rust, which makes things super easy as we can use the [`itf-rs`](https://github.com/informalsystems/itf-rs) library, which takes care of all the boring parts for us. It should not be hard to set it up in other languages as well - people have done it in Go, C++ and Java, for instance.
+
+In Rust, all we have to do is define the types we are parsing. Some type definitions can be re-used from the implementation (under `src`) as they match the Quint types. Usually this is not the case, but this is a toy example:
+```rust
+#[derive(Clone, Debug, Deserialize)]
+pub struct Investment {
+    pub owner: String,
+    pub amount: BigInt,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct BankState {
+    pub balances: HashMap<String, BigInt>,
+    pub investments: HashMap<BigInt, Investment>,
+    pub next_id: BigInt,
+}
+```
+
+For the rest, we also only define their types, and add some `serde` annotations on the `Option` types and on the `mbt::*` metadata:
+```rust
+    use itf::de::{self, As};
+
+    #[derive(Clone, Debug, Deserialize)]
+    pub struct NondetPicks {
+        #[serde(with = "As::<de::Option::<_>>")]
+        pub depositor: Option<String>,
+
+        #[serde(with = "As::<de::Option::<_>>")]
+        pub withdrawer: Option<String>,
+
+        #[serde(with = "As::<de::Option::<_>>")]
+        pub sender: Option<String>,
+
+        #[serde(with = "As::<de::Option::<_>>")]
+        pub receiver: Option<String>,
+
+        #[serde(with = "As::<de::Option::<_>>")]
+        pub amount: Option<BigInt>,
+
+        #[serde(with = "As::<de::Option::<_>>")]
+        pub buyer: Option<String>,
+
+        #[serde(with = "As::<de::Option::<_>>")]
+        pub seller: Option<String>,
+
+        #[serde(with = "As::<de::Option::<_>>")]
+        pub id: Option<BigInt>,
+    }
+
+    #[derive(Clone, Debug, Deserialize)]
+    pub struct State {
+        pub bank_state: BankState,
+
+        #[serde(with = "As::<de::Option::<_>>")]
+        pub error: Option<String>,
+
+        #[serde(rename = "mbt::actionTaken")]
+        pub action_taken: String,
+        #[serde(rename = "mbt::nondetPicks")]
+        pub nondet_picks: NondetPicks,
+    }
+```
+
+Inspecting one of the JSON files should be helpful in understanding which fields are there. [You can use the ITF Trace Viewer]() VSCode extension to view a formatted version of them in VSCode, or use some JSON tool like `jq`.
+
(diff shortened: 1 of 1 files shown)


docs/pages/docs/model-based-testing.mdx
```diff
@@ -1,9 +1,213 @@
...
 # Model-based Testing
 
-<Callout type="warning" emoji="🚧️">
-  This page is under construction.
-</Callout>
+One of the most asked questions when it comes to writing specifications is "what about the code?". Or, in a more expressive problem statement:
+> How to bridge the gap between the specification and the implementation?
 
-For now, refer to Informal System's page on Model-Based Techniques: [mbt.informal.systems](https://mbt.informal.systems/)
+That is, even if we verify that some model is correct, there could still be problems in the code that is ran in production, which is what actually matters. This specially likely to happen if the specification is burried deep into some project sub-sub-sub-folder and no one dares to open it, let alone update it as the code evolves.
```
josef-widder:
```suggestion
That is, even if we verify that some model is correct, there could still be problems in the code runs in production. This is especially likely to happen if the specification is buried deep in some project sub-sub-sub-folder and no one dares to open it, let alone update it as the code evolves.
```

docs/pages/docs/model-based-testing.mdx
```diff
@@ -1,9 +1,213 @@
...
+
+The perfect solution would be: verify your code! But that is far from easy, and can most likely only be accomplished with a fat budget (or some really smart volunteers). Other people might say: write the spec in several layers of abstraction until you are really close to the implementation, proving the refinement between each of them - but that is also hard, and doesn't even address the problem of keeping code and spec in sync. Finally, there's code generation (or extracting code from the formal spec), which is great for some sequential algorithms, but get's weird in many ways when you start having multiple processes - and that discussion doesn't fit in this page.
```
josef-widder:
```suggestion
The perfect solution would be: verify your code! But that is far from easy, and can most likely only be accomplished with a fat budget (or some really smart volunteers). But then, changing the code after it has been verified, can mean that you have to do the exercise again.
Other people might say: write the spec in several layers of abstraction until you are really close to the implementation, proving the refinement between each of them - but that is also hard, and doesn't even address the problem of keeping code and spec in sync. Finally, there's code generation (or extracting code from the formal spec), which is great for some sequential algorithms, but get's weird in many ways when you start having multiple processes - and that discussion doesn't fit in this page.
```

docs/pages/docs/model-based-testing.mdx
```diff
@@ -1,9 +1,213 @@
...
+
+Let's focus on more practical solutions:
```
josef-widder:
```suggestion
We have found that the following two pragmatic approaches strike a balance between investing work and increasing confidence in the code:
```

docs/pages/docs/model-based-testing.mdx
```diff
@@ -1,9 +1,213 @@
...
+Both of these won't give you a proof that your code is correct, but they will make you more confident that the behavior between the model (specification) and the code (implementation) matches, even throughout changes to the code base (or the spec, which tends to change way less).
+
+## MBT in one paragraph
+
+Produce a bunch of traces (executions) from your model, which should be valid traces in your system. For example: Bob deposits $100 and then transfers $60 to Alice, and then Alice withdraws $20. A trace is just a JSON file with all the data you need. You then write a test driver to call the respective functions (`deposit("Bob", 100)`, `transfer("Bob", "Alice", 60)`, `withdraw("Alice", 20)`) in your implementation, and optionally check things between each call or at the very end (such as asserting a property or comparing the state). That's it, you use the model to generate test data (inputs and expectations) for you in a very efficient way.
+
+| Data  | Expectations |
+| :---  | :---         |
+| Model | Model        |
+
```
josef-widder:
```suggestion

Observe the potential here: In many software projects, a test is a function written in your programming language, that captures one specific scenario (Bob is sending to Alice, who has 20 tokens at the end). In our experience, it is very unlikely that someone is going to write more than 5 tests of this type. In contrast, in MBT you write one test driver, and the scenarios are generated by a tool. You can easily test your software against thousands of test scenarios. So instead of saying "my code behaves like the specification in the case Bob sends 20 tokens to Alice" you will be able to say "I have a list of x (put a huge number here) scenarios, where the code behaves like the specification.

```

docs/pages/docs/model-based-testing.mdx
```diff
@@ -1,9 +1,213 @@
...
+
+## Trace validation in one paragraph
+
+Trace validation can be seen as the other way around (from MBT). Instead of using the model to generate test data, you use the data from your production (or staging etc) environments, but the expectations of what should happen with that data will come from the model. You do this by adding some sort of logging to your code that spits out information about the trace (either the transitions, telling what happened, like `deposit("Bob", 100)`; and/or the states, with sequential snapshots of the relevant balances, like `"Bob" -> 40, "Alice -> 60"`). You then capture the logged data and ask Quint: is this sequence of transitions/states possible in my model? And if Quint tells you "no", it means that your code allows something that the model doesn't.
```
josef-widder:
```suggestion
Trace validation can be seen as the other way around (from MBT). Instead of using the model to generate test data, you use the data from your production (or staging etc) environments, but the expectations of what should happen with that data will come from the model. You do this by adding some sort of logging to your code that spits out information about the trace (either the transitions, telling what happened, like `deposit("Bob", 100)`; and/or the states, with sequential snapshots of the relevant balances, like `"Bob" -> 40, "Alice -> 60"`). You then capture the logged data and ask Quint: is this sequence of transitions/states possible in my model? And if Quint tells you "no", it means that your code allows something that the model doesn't, in other words, there is a bug in your code.
```

docs/pages/docs/model-based-testing.mdx
```diff
@@ -1,9 +1,213 @@
-import { Callout } from 'nextra/components'
+import { Callout, Steps } from 'nextra/components'
 
 # Model-based Testing
 
-<Callout type="warning" emoji="🚧️">
-  This page is under construction.
-</Callout>
+One of the most asked questions when it comes to writing specifications is "what about the code?". Or, in a more expressive problem statement:
```
josef-widder:
```suggestion
Quint is a specification language that shines in building complex distributed systems. In such projects there are two questions one needs to address:
1. Am I building the right system?
2. Am I building the system the right way?

When writing a specification in Quint, and using the tools "quint run --invariant..." and "quint verify --invariant." we can check that the design is right, that is, that the interactions of the components ensures that system-wide invariants are satisfied. At this point, big progress has been made: we can be sure that if we implement each of the components in a way that is consistent with the specification, then the system-wide invariants will also hold. We have reduced system-level correctness (potentially involving distributed and concurrent components) to making sure that each component in isolation is implemented according to the specification.

So, naturally, one of the most asked questions when it comes to writing specifications is "what about the code?". Or, in a more expressive problem statement:
```

docs/pages/docs/model-based-testing.mdx
```diff
@@ -1,9 +1,213 @@
...
+> How to bridge the gap between the specification and the implementation?
```
josef-widder:
```suggestion
> How to bridge the gap between the specification and the implementation?

... which is just another way of asking Question 2 from above.
```

docs/pages/docs/model-based-testing.mdx
```diff
@@ -1,9 +1,213 @@
...
+## MBT in one paragraph
+
+Produce a bunch of traces (executions) from your model, which should be valid traces in your system. For example: Bob deposits $100 and then transfers $60 to Alice, and then Alice withdraws $20. A trace is just a JSON file with all the data you need. You then write a test driver to call the respective functions (`deposit("Bob", 100)`, `transfer("Bob", "Alice", 60)`, `withdraw("Alice", 20)`) in your implementation, and optionally check things between each call or at the very end (such as asserting a property or comparing the state). That's it, you use the model to generate test data (inputs and expectations) for you in a very efficient way.
+
+| Data  | Expectations |
+| :---  | :---         |
+| Model | Model        |
+
+
+## Trace validation in one paragraph
```
josef-widder: Is there more coming? We don't have a detailed section on this. We should at least mention that there is none, if there is none.

docs/pages/docs/model-based-testing.mdx
```diff
@@ -1,9 +1,213 @@
...
+    // ...
+    let res = deposit(&mut bank_state, depositor, amount);
+    compare_error(state.value.error.clone(), res)
+}
+```
+</Steps>
+
+### Running the complete example
+
+The repository includes a shell script file that will both generate the traces for you and run `cargo test`. The example includes a mismatch between model and implementation, where the implementation has a bug that allows a user to sell the same investment more than once, potentially draining funds from the bank. The model-based test we wrote identifies this issue. We can be much more confident that the model is correct as we can formally verify that this sort of draining never happens - and model-based testing transfers some of that confidence to the code.
```
josef-widder:
```suggestion
The repository includes a shell script file that will both generate the traces for you and run `cargo test`. The example includes a mismatch between model and implementation, where the implementation has a bug that allows a user to sell the same investment more than once, potentially draining funds from the bank. The model-based test we wrote identifies this issue. 

Assuming that the specification is thoroughly analyzed and or verified, this in fact constitutes a bug in the implementation and not a protocol bug.
```

josef-widder (commented):
I left some comments. Are we adding a more detailed section on trace validation as well?

docs/pages/docs/model-based-testing.mdx
bugarela: I added this

image

PR #1713, opened 2025-07-23
Inductive invariants tutorial

bugarela:
- [ ] I have read and I understand the [Note on AI-assisted contributions](/CONTRIBUTING.md#note-on-ai-assisted-contributions)

### DIFF
--- a/docs/content/posts/inductive_invariants.mdx
+++ b/docs/content/posts/inductive_invariants.mdx
@@ -181,23 +181,40 @@ val IndInv = and {
 ...and with the following command asks quint two questions (1) is IndInv and inductive invariant and (2) if so, does it imply that safety holds. 
 
 ```sh
-quint verify reliablebc.qnt --inductive-invariant IndInv --invariant invariant
+quint verify reliablebc.qnt --inductive-invariant IndInv --invariant safety
 ```
 
 Here is what Quint has to say:
 
 ```ansi
-[90m> [1/2] Checking whether the inductive invariant 'IndInv' holds in the initial state(s) defined by 'init'...[39m
-[90m> [2/2] Checking whether 'step' preserves the inductive invariant 'IndInv'...[39m
+[90m> [1/3] Checking whether the inductive invariant 'IndInv' holds in the initial state(s) defined by 'init'...[39m
+[90m> [2/3] Checking whether 'step' preserves the inductive invariant 'IndInv'...[39m
 [90mAn example execution:[39m
 [90m[39m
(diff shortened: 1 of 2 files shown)


docs/content/posts/inductive_invariants.mdx
```diff
@@ -0,0 +1,511 @@
...

+    [32mSet[39m(

+      { payload[90m:[39m [32m"goodbye"[39m, receiver[90m:[39m [33m2[39m, sender[90m:[39m [33m1[39m },

+      { payload[90m:[39m [32m"goodbye"[39m, receiver[90m:[39m [33m2[39m, sender[90m:[39m [33m2[39m }

+    ),

+  system[90m:[39m [32mMap[39m([33m1[39m[90m ->[39m { s[90m:[39m [32mInit[39m, send_to[90m:[39m [33m2[39m }, [33m2[39m[90m ->[39m { s[90m:[39m [32mSend[39m([32m"goodbye"[39m), send_to[90m:[39m [33m0[39m })
```
josef-widder: Here the highlight is wrong. It should be on `{ payload: "goodbye", receiver: 2, sender: 1 }` instead of `Send("goodbye")`

docs/content/posts/inductive_invariants.mdx
```diff
@@ -0,0 +1,511 @@
...
+[31m[violation][39m Found an issue [90m(688ms).[39m

+[31m  ❌ IndInv[39m

+error: found a counterexample

+
+```
+
+It says that it is violated. But look at `State 0`: Process `1` is in `Init`, but it has already sent a `"goodbye"` message. The verifier has just tried all states described by `TypeOK`, and it doesn't know that if you are in `Init` you haven't sent a message yet. We are in the scenario we discussed above that argued how intuition helps and harms us. Just observe that the verifier doesn't have any intuition; it produces states that satisfy the constraints. If the constraints are not good, it will just produce stupid states. So let's teach the tool this fact, by adding this to our candidate: It says that if there is a message in the network sent by P, then P cannot be in `Init` (anymore). It actually has to be in either `Send(m)` or `Delivered(m)` for that message's payload `m`.
```
josef-widder: > We are in the scenario

Perhaps before this sentence we could add an info box, to explain the difference between an inductive invariant and a normal invariant. Something like:

We are still motivated to show that `safety` is an invariant, that is, that any state reachable from the initial state satisfies `safety`. However, what the tool tells us here is that `safety` **alone** is not enough to characterize correct states. In other words, while hopefully turning out to be an invariant towards the end of this post, we already know now, for certain, that `safety` alone is not inductive. So we need to find a more precise description of a correct state.

docs/content/posts/inductive_invariants.mdx
```diff
@@ -0,0 +1,511 @@
+---
+title: How to Write Inductive Invariants
+date: 2025-07-30
+excerpt: TODO
+authors:
+    - name: Josef Widder
+    - name: Gabriela Moreira
```
josef-widder, line 7:
```suggestion
    - name: Gabriela Moreira
    - name: Josef Widder
```
I suggest we do it in alphabetical order.

PR #1718, opened 2025-07-24
fix(cli) : validate that --n-traces is not greater than --max-samples

### LINKED ISSUE #1712: CLI should return an error when `n-traces` is bigger than `max-samples`
```
  --max-samples  the maximum number of runs to attempt before giving up
                                                       [number] [default: 10000]
  --n-traces     how many traces to generate (only affects output to out-itf)
                                                           [number] [default: 1]
```

Currently, if `n-traces` is bigger than `max-samples`, Quint will quietly produce `max-samples` traces instead of the desired `n-traces`. We should validate this arguments and produce an error in this case instead.

icebelly29:
Fixes #1712

Previously, the CLI would silently limit --n-traces to --max-samples without alerting the user. This change adds validation in the 'run' command so the CLI now throws a clear error when --n-traces exceeds --max-samples.
validated manually with a .qnt spec and confirmed expected behavior for both valid and invalid input cases.

- [x] I have read and I understand the [Note on AI-assisted contributions](/CONTRIBUTING.md#note-on-ai-assisted-contributions)

### Notes
- No automated test added yet, behavior verified via local CLI runs using a test Quint file
Screenshot 2025-07-25 010934
Screenshot 2025-07-25 011111


quint/src/cli.ts
```diff
@@ -399,9 +399,15 @@ const docsCmd = {
   handler: (args: any) => load(args).then(chainCmd(docs)).then(outputResult),
 }
 
-const validate = (_argv: any) => {
-  return true
-}
+// validate that --n-traces is not greater than --max-samples
+const validate = (argv: any) => {
+  if (argv['n-traces'] !== undefined && argv['max-samples'] !== undefined) {
```
bugarela, line 404: We also need to consider this validation when `max-samples` is not given and falls back to the default (10k). Can you check that this covers that? Ideally, we should add a test to `io-cli-tests.md`

icebelly29: The validation does trigger even when ```--max-samples``` is omitted.  I manually tested this (```--n-traces 20000``` with no ```--max-samples```) and added an integration test to `io-cli-tests.md` to cover this case.
Screenshot 2025-07-28 231523

quint/package-lock.json
bugarela, line 1: This file and the one below cannot be removed

quint/testFixture/NTracesValidation.qnt
bugarela, line 1: We can use any other file for this test, I don't see a reason to creating a new file just for this. We could use https://github.com/informalsystems/quint/blob/main/examples/language-features/counters.qnt for example.

quint/package-lock.json
icebelly29: I have added them back

quint/testFixture/NTracesValidation.qnt
icebelly29: Okay yeah fair point, ill remove that file and use the example file instead

bugarela (approved):
Thank you!

PR #1695, opened 2025-06-16
Handle Apalache process termination when Java is missing

teresper-eth:
Fixes #1502 

- [ ] Tests added for any new code
- [ ] Documentation added for any new functionality
- [x] Entries added to the respective `CHANGELOG.md` for any new functionality
- [ ] Feature table on [`README.md`](../README.md#roadmap) updated for any listed functionality

- Add proper error handling for Apalache process exit events
- Improve error messages to guide users about Java requirement
- Update README to make Java dependency more prominent
- Ensure clean process termination when Java is not found

PR #1729, opened 2025-08-12
Fix console hang if Apalache exits early

### LINKED ISSUE #1502: verify: command freezes if no Java SDK is found
I think the spec in case is not relevant here, but I can provide its location if needed.

```
$ quint verify part_stream.qnt                                                                                                                                                       [cason/341-streaming-quint] 
The operation couldn’t be completed. Unable to locate a Java Runtime.
Please visit http://www.java.com for information on installing Java.

```

From this point, `CTRL` + `C` does not work and there is no way to leave the program. Actually, `CTRL` + `Z` does work but something is left running there, and I am not sure what it is, since `$ killall quint` says there is no quint running.

Fortunately, I found out in this [page](https://quint-lang.org/docs/getting-started#verify-the-result) how to download a Java SDK , and after installing it the command worked as expected. It would be helpful, though, that the "error" message above pointed out directly to this information.

The priority here is of course low, but opening this issue just in case.

thpani:
Fixes #1502

I've tried the proposed PR in #1695 and it does not work for me on macOS.

The fix here is twofold:
- add a signal listener on the Apalache subprocess and exit Quint when that subprocess exits (it's no longer necessary to even Ctrl-C Quint if Apalache exits early)
- rewrite `exitHandler()` to remove the listener on Apalache before exiting (it could race) and omit the try/catch (`ESRCH` can no longer happen; we would've exited below if Apalache died)

### DIFF
--- a/.github/workflows/vscode.yml
+++ b/.github/workflows/vscode.yml
@@ -57,7 +57,7 @@ jobs:
             ${{ runner.os }}-nix-
             ${{ runner.os }}-
       - name: Install Nix
-        uses: cachix/install-nix-action@v22
+        uses: cachix/install-nix-action@v31
         with:
           extra_nix_config: |
             access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}


bugarela:
Hey, thank you for picking this up! The CI is failing on Linux, doesn't seem like something too obvious. I'm a bit busy for the rest of today, but can help you debug later this week (as I'm on Linux).

thpani:
Thanks! Maybe it's just a flake?
From the logs it looks like Quint cannot connect to Apalache server, tries to bring up its own, and then fails when binding the port (which would mean the server was running all along).
atm, I don't have a good idea what would cause this, other than a weird timing thing.

I've fixed the failing format CI in the meantime.

bugarela:
I can reproduce the issues locally and it's intermittent (some times I get more failures than others). ChatGPT says:

> Short version: you’ve got a race around a fixed port (8822) plus slightly different socket + process semantics on Linux. On Linux, closing a TCP listener often leaves the port in TIME_WAIT and it can’t be rebound immediately; macOS (BSD) is more permissive with SO_REUSEADDR. That explains the intermittent:
> - Failed to bind ... 8822: Address already in use (a previous Apalache is still bound or the port is in TIME_WAIT).
> - 1 CANCELLED: Call cancelled from @grpc/grpc-js (your client call gets torn down because the channel closes / process exits while you’re still trying to connect, or the server isn’t actually up yet).

What it suggest is to give a different port for each of the test commands (or some more complicated fixes).

I can't reproduce the issue when using `quint verify` normally (without a bunch of quick commands like in the integration test), so I guess we can consider this a test-only issue and fix ports for the tests to avoid it.

thpani:
> What it suggest is to give a different port for each of the test commands (or some more complicated fixes).

From what I understand there is a single apalache server brought up by the integration test before txm runs:
https://github.com/informalsystems/quint/blob/54440b0efcebff8248836b1c0be7128a4b6b1476/.github/workflows/integration.yml#L76-L79

If you run it locally on Linux with a separate apalache server, does something bring it down after the first test?

konnov:
The tests also run on my Ubuntu machine

bugarela:
> If you run it locally on Linux with a separate apalache server, does something bring it down after the first test?

Nope, this works fine. Tests are only breaking if I don't have a server running.

> The tests also run on my Ubuntu machine

Maybe it's something specific to the Nix flake then, as that's what I'm using locally as well.

bugarela:
Just saw this: https://github.com/apalache-mc/apalache/pull/3130

Will do the same here

thpani:
> I've added a sleep between starting the server in the CI and running the tests.

Good catch! If the server isn't up yet, it would cause Quint to launch its own, and since the TXM tests run in parallel, it would try to bind the same port multiple times.

PR #1741, opened 2025-08-22
Duplicate field on records should return parse error

### LINKED ISSUE #1677: Report error on duplicated fields on records

el-yawd:
Fixes #1677

- [ ] Documentation added for any new functionality N/A
- [ ] Entries added to the respective `CHANGELOG.md` for any new functionality N/A

image

I couldn't find how to not get this debug message (`[DEBUG] generating undefined expr to fill hole in: vala={foo:1,foo:2}`), it seems to be extremely trivial but since I'm not familiar with the codebase it's kinda hard to find, any help on this would be very appreciated.

### DIFF
--- a/quint/src/parsing/ToIrListener.ts
+++ b/quint/src/parsing/ToIrListener.ts
@@ -38,7 +38,13 @@ import { TerminalNode } from 'antlr4ts/tree/TerminalNode'
 import { QuintTypeDef } from '../ir/quintIr'
 import { zip } from '../util'
 import { QuintError } from '../quintError'
-import { lowercaseTypeError, mapSyntaxError, tooManySpreadsError, undeclaredTypeParamsError } from './parseErrors'
+import {
+  duplicatedRecordField,
+  lowercaseTypeError,
+  mapSyntaxError,
+  tooManySpreadsError,
+  undeclaredTypeParamsError,
+} from './parseErrors'
 import { Loc } from '../ErrorMessage'
 import assert, { fail } from 'assert'
 import { typeToString } from '../ir/IRprinting'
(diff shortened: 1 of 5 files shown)


bugarela:
> I couldn't find how to not get this debug message ([DEBUG] generating undefined expr to fill hole in: vala={foo:1,foo:2}), it seems to be extremely trivial but since I'm not familiar with the codebase it's kinda hard to find, any help on this would be very appreciated.

~~This is expected, sorry for the confusion. One of the tests has an error that causes this message to be logged, and we still don't have a log level setup to omit debug messages, so it shows up every time.~~

Actually, I just saw this is happening in the REPL when you type the expression, not when running the tests. so it is actually a problem. I'll take a look

quint/src/parsing/ToIrListener.ts
```diff
@@ -788,6 +794,18 @@ export class ToIrListener implements QuintListener {
...
+    // Check for duplicate field names
+    const fieldNames = pairs
+      .map(pair => pair[0])
+      .filter((field): field is QuintStr => field.kind === 'str')
+      .map(field => field.value)
+
+    if (new Set(fieldNames).size !== fieldNames.length) {
+      const id = this.getId(ctx)
+      this.errors.push(duplicatedRecordField(id))
+      return
```
bugarela: I think removing the return here might fix the [DEBUG] message. The idea is that we don't prevent the expression from being built in this case, we just continue and then report the error in the end. This helps us potentially find other errors and report all of them in the same pass, which is better for the user.
```suggestion
```

el-yawd: bah, that makes sense, rookie mistake. Thanks! I've just fixed :)

el-yawd:
@bugarela Sure! There is only one caveat, I took the spread example from [here](https://github.com/informalsystems/quint/pull/1681#discussion_r2126402848), but if I do:

```text
>>> val a = {foo:1, bar: 2}

>>> {...a, foo: 1, baz: 3}
static analysis error: error: [QNT000] Incompatible tails for rows with disjoint fields: {} and { | tail__t1 }
Trying to unify { foo: int, bar: int } and { baz: int | tail__t1 }

{...a, foo: 1, baz: 3}
^^^^^^^^^^^^^^^^^^^^^
```
It triggers this error instead of duplicated, is this right?

bugarela:
> @bugarela Sure! There is only one caveat, I took the spread example from [here](https://github.com/informalsystems/quint/pull/1681#discussion_r2126402848), but if I do:
> 
> ```
> >>> val a = {foo:1, bar: 2}
> 
> >>> {...a, foo: 1, baz: 3}
> static analysis error: error: [QNT000] Incompatible tails for rows with disjoint fields: {} and { | tail__t1 }
> Trying to unify { foo: int, bar: int } and { baz: int | tail__t1 }
> 
> {...a, foo: 1, baz: 3}
> ^^^^^^^^^^^^^^^^^^^^^
> ```
> 
> It triggers this error instead of duplicated, is this right?

Yes! Right is a strong word, as the error message for this is awful, but that's on me :sweat_smile:. In this example, there's no duplication, only adding a `baz` field which doesn't exist. I think the only way of triggering duplication on spreads is like with the example you added to the tests, so we are all good :smile:

bugarela (approved):
Amazing, thank you!

Issue #1757, opened 2025-09-05
Passing the same argument twice (--max-steps) causes weird behavior


bugarela:
This could be any spec, just an arbitrary example:
```
$ quint run tendermint.qnt --main no_agreement --max-steps=10 --max-steps=15
```

This results in a example execution of a single state, so seems like the argument is interpreted as zero.

Expected behavior would be an informative error message saying you are passing the argument more than once.

I just tested this with `--max-steps`, but it should be the case with at least some other arguments. Maybe there's a general solution in `yargs`.

PR #1767, opened 2025-09-15
Validate that non-array args can't be passed twice

erickpintor:
Yargs will unexpectedly convert these to arrays, which leads to unspecified behavior in the codebase. There are two possible fixes:

1. Disallow converting duplicated args to arrays;
2. Validate that non-array options can't be passed twice.

Option (1) requires close to no code but, it makes yargs ignore duplicated arguments and silently keep only the last one. Option (2) requires additional code but, it allows for more explicit error messages.

Ultimately option (2) was selected since it'll tell users about possible mistakes instead of implicitly deciding which arguments to take and which ones to ignore.

See https://github.com/yargs/yargs/issues/1318.

Closes https://github.com/informalsystems/quint/issues/1757.


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Fixed several issues with the integration of Apalache (#1754)
 - Fixed unauthorized errors when downloading the Rust backend (#1763)
+- Fixed undefined behavior when non-array options are passed twice (#1757)
```
bugarela:
```suggestion
- Fixed undefined behavior when non-array options are passed twice in the CLI (#1757)
```

quint/io-cli-tests.md
```diff
@@ -1546,6 +1546,19 @@ quint run ./examples/language-features/counters.qnt --n-traces 20000
...
+
+<!-- !test in non-array options set more than once -->
+```sh
+quint run ./examples/language-features/counters.qnt --max-steps 10 --max-steps 15
+```
+
+<!-- !test exit 1 -->
+<!-- !test err non-array options set more than once -->
+```text
+--max-steps can not be specified more than once
```
bugarela, line 1559: Perfect :sparkles:

quint/src/cli.ts
```diff
@@ -399,8 +399,15 @@ const docsCmd = {
   handler: (args: any) => load(args).then(chainCmd(docs)).then(outputResult),
 }
 
-// validate that --n-traces is not greater than --max-samples
-const validate = (argv: any) => {
+// Perform parser input validation.
+const validate = (argv: any, opts: any) => {
+  // Validate that only array options can be specified more than once.
+  for (const key in argv) {
+    if (key !== '_' && !opts.array.includes(key) && Array.isArray(argv[key])) {
```
bugarela, line 406: Is `'_'` some sort of special key for yargs? I don't think I've seen anything about this before.

Also, do I understand correctly that `opts.array` is API provided by yargs?

erickpintor: Yes. The `_` represents the non-option arguments, for example: `program foo bar --baz 42`, `_` will be `['foo', 'bar']`. The `opts` is provided by yargs as a hash of parser configs. From there we can get the keys for all options configured as arrays with `opts.array`. Here's a snippet of the `opts` argument provided by yargs:

```
{
  ...
  array: [ 'invariants', 'witnesses', 'hide' ],
  boolean: [ 'help', 'version', 'mbt' ],
  string: [
    'out',     'main',
    'out-itf', 'init',
    'step',    'invariant',
    'seed',    'backend'
  ],
  ...
}
```

PR #1775, opened 2025-09-18
Move recent CHANGELOG entries to UNRELEASED

erickpintor:
I only now realized that I've been placing entries in the CHANGELOG file in the wrong section of it :man_facepalming:


bugarela:
Nope, you are actually good, it's just that I've cut a release yesterday (sorry for not letting you know). Cutting a release moves the items from UNRELEASED to their respective release (0.28.0 in this case)

erickpintor:
Ah. Okay. I missed that release. No worries.

Issue #577, opened 2023-01-27
REPL copy-paste breaks on empty prompts


thpani:
Copy-pasting REPL into REPL fails on empty `>>>` prompts.

To reproduce: copy-paste the following, valid REPL session into REPL:

```
>>> 1+5
6
>>> 
>>> 42
42
```

(Note the second, empty `>>>` created by skipping that line immediately with `ENTER`.)

Fails with

```
syntax error: <input>:1:1 - error: mismatched input '>' expecting {'module', '{', 'const', 'var', 'assume', 'type', 'import', 'val', 'def', 'pure', 'action', 'run', 'temporal', '[', 'all', 'any', 'if', 'nondet', '_', STRING, BOOL, INT, 'and', 'or', 'iff', 'implies', 'Set', 'List', 'Map', '-', '(', IDENTIFIER}
1: >>>
```

PR #1776, opened 2025-09-18
Fix copy and past for multi-line expression

erickpintor:
The repl supports trimming the prompt and the continue prompt prefixes when copying and pasting previous expressions form the repl. However, the '...' prefix of multi-line expressions confuses the interpreter into thinking that those lines are special commands instead.

This patches fixes the repl's line handler so that lines starting with '...' continue prompt prefix are interpreted as quint expressions, thus allowing for copying and pasting multi-line commands from the repl itself.

### DIFF
--- a/quint/src/repl.ts
+++ b/quint/src/repl.ts
@@ -327,7 +327,9 @@ export function quintRepl(
     const g = (s: string): string => {
       return chalk.gray(s)
     }
-    if (!line.startsWith('.')) {
+    // The continue prompt is handled by `nextLine` as the input may be acopy
+    // and paste from the reply itself.
+    if (!line.startsWith('.') || line.startsWith(settings.continuePrompt)) {
       // an input to evaluate
       nextLine(line)
     } else {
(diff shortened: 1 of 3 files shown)


quint/src/repl.ts
```diff
@@ -327,7 +327,9 @@ export function quintRepl(
     const g = (s: string): string => {
       return chalk.gray(s)
     }
-    if (!line.startsWith('.')) {
+    // NB. the continue promppt is handled by `nextLine` as the input may be a
+    // copy and past from the reply itself.
```
bugarela: I had to google what `NB` was :sweat_smile: this is definitely not common use in this code base, so maybe it's better not to introduce it by now.

The other suggestions are on typos.
```suggestion
    // The continue prompt is handled by `nextLine` as the input may be a
    // copy and paste from the reply itself.
```

bugarela (approved):
Nice one! There's also https://github.com/informalsystems/quint/issues/577 which is related, if you want to use the context you just got to fix it as well :smile:

quint/src/repl.ts
erickpintor: Sorry! That's muscle memory from a past life :D

Issue #1768, opened 2025-09-16
CLI doesn't allow passing commands to the REPL


bugarela:
# Context

Our current way of calling the REPL with some pre-defined command is to use `echo` and pipe (see https://github.com/informalsystems/quint/blob/main/quint/io-cli-tests.md#use-of-repl-test-and-run). This is not the most convenient way, and is specially a problem when our LLM friends try to run the REPL, we can't trust generic commands starting with `echo` and, therefore, need to approve each command individually (instead of approving all commands starting with `quint repl`, for example).

We already have a way to start the REPL with a given list of expressions, that can currently be given at the end of the file as comments like in https://github.com/informalsystems/quint/blob/5be46f25c8147809705fa849ef485dbd64aab8aa/examples/tutorials/repl/kettle.qnt#L96

Running `quint -r kettle.qnt` executes all those commands. Also, it works without the module because it has a `@mainModule kettle` annotation at the top.

# Task details

Evaluating the given expressions is currently handled here: https://github.com/informalsystems/quint/blob/5be46f25c8147809705fa849ef485dbd64aab8aa/quint/src/repl.ts#L293-L299

I don't have a specific interface in mind. Perhaps this can be an yargs array so we can pass things like:
```sh
quint -r module.qnt --eval foo --eval bar
```
or
```sh
quint -r module.qnt --eval foo bar
```

But other ideas could work as well.

PR #1777, opened 2025-09-18
Introduce support for repl eval expressions

erickpintor:
Currently, the CLI does not support opening the repl with a expression to evaluate. Users managed to work around that with linux pipes where they first `echo` the command they want to run, then pipe the result to the quint process. Although this approach works, it's less ergonomic and causes LLMs to ask for confirmation before running commands in the repl due to mistrust in the `echo` command.

This patch introduces support to calling quint repl with a command to be evaluated upon entering the repl.

Note that the [issue](https://github.com/informalsystems/quint/issues/1768) suggests `--eval "expr"`. I ended up going with simply another command line as it feels more ergonomic in most shells. Let me know you feel like the interface should be different.

Closes https://github.com/informalsystems/quint/issues/1768.


bugarela:
The reason why I suggested `--eval expr` in the issue is that, this way, we can pass multiple commands (with multiple `--eval` options or `--eval foo bar baz`). With `echo`, this is possible by using line breaks (`echo "1 + 1\n2 + 2"`).

It is important that we support multiple commands because the REPL is stateful and we often need to set up the environment (i.e. `init` + a few actions) before evaluating some expression.

That being said, I do like how elegant `quint "1 + 1"` looks, and how we could use it like `quint -r tendermint.qnt::too_many_byz disagreementTest`. Maybe we can keep it in addition to `--eval`. But, for this case, I'd expect the REPL to quit after evaluating. Not sure.

erickpintor:
Most modern shells have some form of multi-line string support these days. For example:

**Using here-document:**

```
$ quint <<EOF
1+1
2+2
EOF

Quint REPL 0.28.0
>>> 1+1
2
>>> 2+2
4
```

Supported by bash and zsh.

**Using multi-line strings:**

```
$ quint '1+1
2+2'

Quint REPL 0.28.0
>>> 1+1
2
>>> 2+2
4
```

Supported by bash and fish.

The here-document was already supported, BTW, since it works by replacing the process `stdin`.

That being said, perhaps we could allow for an multiple expressions to be passed in? Something like:

```
$ quint -r spec.qnt \
  "init" \
  "step" \
  "inv"
```

This approach might be helpful for scripting where you can compose with other functions in the script that produce expressions:

```
$ quint -r spec.qnt \
  $(derive_init) \
  $(derive_step) \
  "inv"
```

I don't discard adding `--eval` but, it feels more verbose to me. Since the primary goal of the arguments in this use-case is to run expressions to setup the repl, passing them in directly feels more straightforward. 

What do you think?

erickpintor:
@bugarela I've implemented the multiple positional arguments approach. Lemme know what you think.

bugarela (approved):
Looks great! Thanks for adding support for multiple commands :)

PR #1780, opened 2025-09-23
Introduce parallel simulations

erickpintor:
This patch receives a `nthreads` input from the quint CLI. If present and more than 1, it runs simulations in parallel using the specified number of threads.

### DIFF
--- a/evaluator/Cargo.toml
+++ b/evaluator/Cargo.toml
@@ -6,6 +6,10 @@ edition = "2021"
 [profile.release]
 codegen-units = 1
 lto = "thin"
+panic = 'abort'
+
+[profile.dev]
+panic = 'abort'
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
--- a/evaluator/src/main.rs
+++ b/evaluator/src/main.rs
@@ -7,6 +7,7 @@
 use std::fs::{self, File};
 use std::io::{self, Read, Write};
 use std::path::PathBuf;
+use std::sync::Arc;
 use std::time::Instant;
 
 use argh::FromArgs;
@@ -82,9 +83,11 @@ struct SimulateInput {
     nruns: usize,
     nsteps: usize,
     ntraces: usize,
+    #[serde(default)]
+    nthreads: usize,
 }
 
-#[derive(Serialize)]
+#[derive(Eq, PartialEq, Serialize)]
 #[serde(rename_all = "lowercase")]
 enum SimulationStatus {
     #[serde(rename = "ok")]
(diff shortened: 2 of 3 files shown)


evaluator/src/main.rs
```diff
@@ -186,27 +189,105 @@ fn simulate_from_stdin() -> eyre::Result<()> {
...
+    let reporter_thread = progress::spawn_reporter_thread(nruns);
+
+    for i in 0..nthreads {
+        // FIXME: it should be possible to share the `ParsedQuint` across
+        // threads since it's immutable after the parsing phase. However, its
+        // internals use `hipstr::LocalHipStr` and other data structures that
+        // are thread-local. We'll clone the whole `ParsedQuint` here for now.
+        let parsed = parsed.clone();
+
+        // In the case of an odd split, the first thread executes the remain
```
bugarela: :nail_care:  I think "odd" is only the right word if `nthreads` is 2 :sweat_smile: 
```suggestion
        // In the case of a non-exact split, the first thread executes the remain
```

evaluator/src/main.rs
```diff
@@ -186,27 +189,105 @@ fn simulate_from_stdin() -> eyre::Result<()> {
...
+    loop {
+        let mut outcome = out_rx.recv().expect("closed channel");
+        samples += outcome.samples;
+        nthreads -= 1;
+        if nthreads == 0 || outcome.status != SimulationStatus::Success {
+            // Report back the total number of samples executed by all threads
+            // instead of just this one. Note that trace statistics are
+            // preserved, meaning that the final report will correspond to the
+            // trace statistics of a single thread. Presuming the workload is
+            // homogeneous (no thread is performing work that is significantly
```
bugarela, line 276: I think it would be useful to aggregate the statistics, specially because of the `max` and `min` trace lengths, as believing that the reported max is the actual max after 10k samples is quite different than believing that for 500 samples, so we are losing some data quality here.

Not important tho, we can leave this for later. We will also need to aggregate data in a similar way when we implement the `--witneeses` option in the rust simulator.

evaluator/Cargo.toml
```diff
@@ -6,6 +6,10 @@ edition = "2021"
 [profile.release]
 codegen-units = 1
 lto = "thin"
+panic = 'abort'
+
+[profile.dev]
+panic = 'abort'
```
bugarela, line 12: What is the motivation for this in the dev env?

erickpintor: It makes it so that any thread that panics kills the whole process instead of just logging the panic. I've found it useful as it catches bugs quicker instead of having the main thread hanging waiting on threads that won't finish.

evaluator/src/main.rs
erickpintor: That's fair. We can address this when implementing witnesses.

PR #1781, opened 2025-09-23
Use parallel simulations with rust backend

### LINKED ISSUE #1637: [Rust] Parallelize simulations
One simulation is independent of the other. It should be straightforward to run them in parallel. The goal would be to use multiple cores so users can check more samples in the same amount of time. We should be careful to not run all parallel instances with the same seed - there's no point checking the same sample more than once.

erickpintor:
This patch makes the CLI call the rust backend with the available number of CPUs by default, thus runnig simulations in parallel.

Closes #1637.

### DIFF
--- a/quint/rust-backend-tests.md
+++ b/quint/rust-backend-tests.md
@@ -13,10 +13,14 @@ Asserts that on an empty `QUINT_HOME` directory, it downloads the latest Rust
 evaluator from GitHub releases. Note that a successful run indicates a
 successful download and execution of the spec with the set backend.
 
+This test passes the `--n-threads` flag to the rust backend to check that the
+CLI accepts it correctly.
+
 <!-- !test in download and run -->
 ```
-QUINT_HOME=$(mktemp -d) quint run \
+QUINT_HOME="$(mktemp -d)" quint run \
   --backend=rust \
+  --n-threads=4 \
   ./testFixture/simulator/gettingStarted.qnt \
   | grep -o "Downloading Rust evaluator from https://api.github.com/repos/informalsystems/quint/releases/assets/"
 ```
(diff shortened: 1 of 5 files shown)


quint/rust-backend-tests.md
```diff
@@ -13,10 +13,14 @@ Asserts that on an empty `QUINT_HOME` directory, it downloads the latest Rust
 evaluator from GitHub releases. Note that a successful run indicates a
 successful download and execution of the spec with the set backend.
 
+This test passes the `-n-threads` flag to the rust backend to verify that the
```
bugarela: I like to avoid using `verify` unless we are referring to `quint verify` or software verification in some other sense - just to avoid overloading of the word and confusion, cause when I read `verify` I immediately think of `quint verify`

```suggestion
This test passes the `--n-threads` flag to the rust backend to check that the
```

quint/src/cli.ts
```diff
@@ -257,6 +258,11 @@ const runCmd = {
         type: 'number',
         default: 20,
       })
+      .option('n-threads', {
+        desc: 'the number of threads to use when running simulations with the `rust` backend',
```
bugarela, line 262: should we add a check or at least a warning for the case someone provides `--n-threads` but not `--backend=rust`?

erickpintor: I've tried but I ran into a problem: by setting the default to the number of cores, I can't distinguish if you're setting the value yourself of the tool is setting it for you. I've played with the idea of having the default set to `null` and derive the number of threads later but, I've found the CLI help more informative this way.

```
quint run <input>

Simulate a Quint specification and (optionally) check invariants

Options:
  ...
  --n-threads    the number of threads to use when running simulations with the
                 `rust` backend                           [number] [default: 24]
  ...
```

Lemme know if you want me to try something else.

bugarela: Makes sense, I thought this could be a problem but hoped there was some way around it. Let's go with this then!

bugarela (approved):

PR #1782, opened 2025-09-24
Move rust integration tests to a pre-release step

erickpintor:
Rust integration tests are flaky when running on a per PR basis due to the GH API rate limit that prevents us from downloading the evaluator multiple times in parallel. This patch moves the integration test to a pre-release step so it runs only as a sanity check before releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,8 +21,20 @@ jobs:
           registry-url: "https://registry.npmjs.org"
       - name: Compile
         run: npm ci
-      - name: Test
+      - name: Run unit tests
         run: npm run test
+      - name: Run integration tests
+        run: |
+          # Link quint to the system path
+          npm link
+          # Fetch the latest apalache release
+          make -C .. apalache
+          # Start the apalache server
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -23,6 +23,8 @@ jobs:
         run: npm ci
       - name: Test
         run: npm run test
+      - name: Rust backend integration tests
+        run: npm run compile && npm link && npm run rust-integration
```
bugarela: Do we also want to add all other integration tests here? We have these in `package.json`:
```json
    "integration": "txm cli-tests.md && txm io-cli-tests.md",
    "apalache-integration": "txm apalache-tests.md",
    "apalache-dist": "txm apalache-dist-tests.md",
```

I think it would be useful to add a `"all-integration"` task or something, include all of them and then here have
```suggestion
      - name: Integration tests
        run: npm run compile && npm link && npm run all-integration
```

erickpintor: This is a good idea. I've updated the PR to run all integration test as a pre-release check.

PR #1783, opened 2025-09-24
Showing diff between states when making transitions in the REPL

bugarela:
Hello :octocat: 

This changes the REPL so we pretty print a diff from old state to new state whenever it makes a transition:

image

This can be disabled by setting `--verbosity` to something lower than 2.

The pretty printer is heavily based on the existing pretty printer we already have, but the code is significantly different as we focus on the diff. We collapse (`...`) parts of the diff that are not relevant. This pretty printer was written with heavy help from AI.

## Integration test updates

I had to update some tests to disable this diff printing (with `--verbosity 1`) as the information printed can be non-deterministic, and we don't set the seeds in most tests. I took this opportunity to also update them to use the recently added CLI option of passing the commands as arguments instead of using echo. I kept some echo ones for completeness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Added support for passing commands to the REPL (#1768)
+- Added support for running simulations in multiple threads with the rust backend (#1637)
 - The REPL now prints a diff between states when it makes a transition (#1783)
 
 ### Changed
--- a/docs/app/layout.tsx
+++ b/docs/app/layout.tsx
@@ -5,7 +5,6 @@ import { getPageMap } from 'nextra/page-map'
 import { InformalSystemsLogo } from '../components/home/InformalSystemsLogo'
 import '../style.css'
 import { LogoSwitcher } from '../components/LogoSwitcher'
-import { AnnouncementBanner } from '../components/AnnouncementBanner'
 
 export const metadata = {
   title: 'Quint',
@@ -81,7 +80,6 @@ export default async function RootLayout({ children }) {
         <Layout
           navbar={navbar}
           footer={footer}
-          banner={<AnnouncementBanner />}
           pageMap={await getPageMap()}
           docsRepositoryBase="https://github.com/informalsystems/quint/blob/main/docs"
           sidebar={{
--- a/evaluator/src/ir.rs
+++ b/evaluator/src/ir.rs
@@ -2,13 +2,16 @@
 //! input.
 
 use fxhash::FxBuildHasher;
-use hipstr::LocalHipStr;
+use hipstr::HipStr;
 use indexmap::IndexMap;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 pub type QuintId = u64;
-pub type QuintName = LocalHipStr<'static>;
+// NOTE: be aware of a bug in HipStr where a LocalHipStr is allowed to cross
+// thread boundaries by implementing Send, however, it causes a double free on
+// large heap-allocated strings.
+pub type QuintName = HipStr<'static>;
 
 #[derive(Debug, Clone, Error, PartialEq, Serialize)]
 #[error("[{code}] {message}")]
--- a/evaluator/src/main.rs
+++ b/evaluator/src/main.rs
@@ -234,8 +234,8 @@ fn simulate_in_parallel(
     for i in 0..nthreads {
         // FIXME: it should be possible to share the `ParsedQuint` across
         // threads since it's immutable after the parsing phase. However, its
-        // internals use `hipstr::LocalHipStr` and other data structures that
-        // are thread-local. We'll clone the whole `ParsedQuint` here for now.
+        // internals use data structures that are thread-local. We'll clone the
+        // whole `ParsedQuint` here for now.
         let parsed = parsed.clone();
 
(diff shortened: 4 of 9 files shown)


quint/src/repl.ts
```diff
@@ -616,7 +616,10 @@ function tryEval(out: writer, state: ReplState, newInput: string): boolean {
       if (ex.kind === 'bool' && ex.value) {
         // A Boolean expression may be an action or a run.
         // Save the state, if there were any updates to variables.
-        const missing = state.evaluator.shiftAndCheck()
+        const [shifted, missing] = state.evaluator.shiftAndCheck()
+        if (shifted && verbosity.hasDiffs(state.verbosity)) {
+          console.log(state.evaluator.trace.renderDiff(terminalWidth(), { collapseThreshold: 2 }))
```
bugarela, line 621: We normally use `out` instead of `console.log` here, but `out` makes this be captured by the unit test output check, and that would add unnecessary complexity to the tests. I'd rather test this only with integration tests.

The bad part is that using `console.log` makes us print stuff in the middle of testing, so `npm test` gets a bit polluted.

quint/src/runtime/impl/runtimeValueDiff.ts
```diff
@@ -0,0 +1,448 @@
...
+    return false
+  }
+
+  // Try different value types to check their size
+  const checkers = [
+    () => value.toOrderedMap().size, // Records
+    () => value.toMap().size, // Maps
+    () => value.toList().size, // Lists/tuples
+    () => (value.isSetLike ? value.toSet().size : 0), // Sets
+  ]
```
erickpintor, line 42: Q: Do we have a tag in `RuntimeValue` that we can switch? The sequence of try/catches here feels like could be error pone as the code base evolves. Also, I'm not sure how node does it but, exceptions are expensive/slow to throw and catch in other programming languages.

quint/src/runtime/impl/runtimeValueDiff.ts
```diff
@@ -0,0 +1,448 @@
...
+
+/**
+ * Check if a runtime value is a primitive (bool, int, string)
+ */
+function tryRenderPrimitive(value: RuntimeValue, color: (s: string) => string): Doc | undefined {
+  const primitiveCheckers = [
+    () => tint(String(value.toBool()), color),
+    () => tint(value.toInt().toString(), color),
+    () => group([tint('"', color), tint(value.toStr(), color), tint('"', color)]),
+  ]
```
erickpintor, line 67: Same question here about type tags vs try/catches.

quint/src/runtime/impl/runtimeValueDiff.ts
```diff
@@ -0,0 +1,448 @@
...
+
+  for (const checker of primitiveCheckers) {
+    try {
+      return checker()
+    } catch {
+      // This primitive type check failed, continue to next
+      continue
+    }
+  }
+  return undefined
```
erickpintor, line 77: Q: does typescript prefer `undefined` over `null`? What's the standard in the code base?

quint/src/repl.ts
erickpintor: That does bother me a bit but, I think we can leave with it for a while and figure out later what we do with this console.log (perhaps omit during tests with an env var? idk).

erickpintor (approved):
This looks good to me. I have a few questions about the way we switch between runtime values. I think it'd be good to stay away from try/catches as they tend to be error prone in my experience. Nonetheless, the output looks super good and given that this isn't in a the hot path of the application where performance would be a concern, I think it's fine if we merge it and address technical concerns later.

Great work!

quint/src/runtime/impl/runtimeValueDiff.ts
bugarela: we don't have a tag. The other approach would be to have checks like `value instanceOf RuntimeValueRecord`, for which we'd have to start exposing (exporting) the `RuntimeValueRecord` classes, which we don't currently.

I think the ideal solution would be to change `RuntimeValue` representation into a disjunctive union like all of our other types, but this was already done in the Rust side, and hopefully this code* will be retired soon, so I think it's not worth to refactor that part too much.

*by "this code" I mean the evaluator code around `RuntimeValue`. We probably want to keep some of `RuntimeValue` as we use that in the conversion from ITF traces (from Rust and Apalache), for this diff code, etc. It's a useful representation. But it will be way easier to refactor it once we don't have all the evaluator code around it.

quint/src/runtime/impl/runtimeValueDiff.ts
bugarela: Yes, we never use `null`, only `undefined`. I haven't done research in what are the standards, but this is what I see around. Probably due to the question mark syntax (`foo?: int`) using `undefined` as we talked yesterday.

quint/src/runtime/impl/runtimeValueDiff.ts
bugarela: Same answer applies. Is either this, `instanceOf` (+ exporting inner classes from `RuntimeValue`) or changing `RuntimeValue` significantly.

bugarela:
The `console.log` comment is not answerable for some reason :thinking:. I'll reply here: Maybe we have have a different version of `out` that is ignored in tests, but I can't think of a decent name for it :grimacing:. Or an optional arg in `out`. 

Although I think we do have a couple more `console.log`s leaking in tests, so maybe a more general solution fits best.

PR #1790, opened 2025-09-26
Remove REPL version line from test expectation

bugarela:
Hello :octocat: 

I've forgot to remove the REPL banner from the test output before and now the test fails on new versions of Quint :facepalm: 

Here's the fix


erickpintor (approved):
🐑 it!

PR #1802, opened 2025-10-30
Fix issue where Rust simulator didn't store the violation trace

bugarela:
Hello :octocat: 

Josef found an issue at our retreat and now Yassine hit the same issue: the counterexample reported in the rust simulator was not actually a counterexample. We sort the traces we find in a way where violations always come first, and the "best" trace (sorted in first place) is the one we show to the user. However, when saving the trace to this storage that handles sorting, I was marking violation traces with `violation: false`, which led them to be sorted incorrectly and not end up first on the list :grimacing:

This fixes that and adds a test to cover this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 ### Changed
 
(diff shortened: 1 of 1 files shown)


erickpintor (approved):
🐑 it!

PR #1809, opened 2025-11-14
refactor: replace lodash.isequal with node:util.isDeepStrictEqual, #1606

yazaldefilimone:

### DIFF
--- a/quint/src/effects/simplification.ts
+++ b/quint/src/effects/simplification.ts
@@ -12,8 +12,9 @@
  * @module
  */
 
-import { isEqual, unreachable } from '../util'
+import { unreachable } from '../util'
 import { ConcreteEffect, Effect, Entity, StateVariable } from './base'
+import { isDeepStrictEqual } from 'node:util'
 
 /**
  * Simplifies a concrete effect by:
--- a/quint/src/effects/substitutions.ts
+++ b/quint/src/effects/substitutions.ts
@@ -17,7 +17,7 @@ import { ErrorTree, buildErrorTree } from '../errorTree'
 import { Effect, Entity } from './base'
 import { effectToString, substitutionsToString } from './printing'
 import { simplify } from './simplification'
-import { isEqual } from 'lodash'
+import { isDeepStrictEqual } from 'node:util'
 
 /*
  * Substitutions can be applied to both effects and entities, replacing
--- a/quint/src/runtime/impl/nondet.ts
+++ b/quint/src/runtime/impl/nondet.ts
@@ -15,7 +15,7 @@
 import { Either, right } from '@sweet-monads/either'
 import { EvalFunction } from './builder'
 import { RuntimeValue, rv } from './runtimeValue'
-import { isEqual } from 'lodash'
+import { isDeepStrictEqual } from 'node:util'
 import { QuintError } from '../../quintError'
 import { CachedValue } from './Context'
 import { Maybe } from '@sweet-monads/maybe'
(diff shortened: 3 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -436,7 +436,7 @@ lose the advantage of exhaustiveness checking. Here's an example:
 Assume we have a type `T`
 
 ```typescript
-type T = 'a' | 'b' | 'c'
+type T = "a" | "b" | "c";
```
bugarela: Those work fine with simple quotes and are preferable afaik
```suggestion
type T = 'a' | 'b' | 'c';
```

CONTRIBUTING.md
```diff
@@ -569,12 +569,12 @@ type that can be used to represent a value that can either be a success or a
 failure.
 
 ```typescript
-import { left, right } from '@sweet-monads/either'
+import { left, right } from "@sweet-monads/either";
```
bugarela:
```suggestion
import { left, right } from '@sweet-monads/either'
```

CONTRIBUTING.md
```diff
@@ -569,12 +569,12 @@ type that can be used to represent a value that can either be a success or a
...
 
 if (all_good(expr)) {
-  return right(value)
+  return right(value);
```
bugarela:
```suggestion
  return right(value)
```

CONTRIBUTING.md
```diff
@@ -569,12 +569,12 @@ type that can be used to represent a value that can either be a success or a
...
 } else {
-  return left({ code: 'QNT500', message: `${expr} is not good!`, reference: expr.id })
+  return left({ code: "QNT500", message: `${expr} is not good!`, reference: expr.id });
```
bugarela:
```suggestion
  return left({ code: 'QNT500', message: `${expr} is not good!`, reference: expr.id })
```

quint/package-lock.json
bugarela, line 1: We should not delete this file, it is necessary for building Quint with nix

quint/src/util.ts
```diff
@@ -112,3 +113,11 @@ function findInsertionIndex<A>(array: A[], pred: (a: A) => boolean): number {
 
   return low
 }
+
+/**
+ * This function replaces `lodash.isEqual` using the native Node.js implementation.
+ *
+ */
+export function isEqual<L, R>(left: L, right: R): boolean {
+  return isDeepStrictEqual(left, right)
```
bugarela: Not sure if we need this indirection here through `util` - we could just replace the imports to import `isDeepStrictEqual` and then use it directly instead of `isEqual` everywhere. @erickpintor do you have some advice here?

erickpintor: I agree. It'd be simpler to just rely on `node:util` for equality everywhere.

bugarela: @yazaldefilimone can we delete this now?

quint/package-lock.json
bugarela: @yazaldefilimone can you add this file back please? We can't delete it.

bugarela (commented):
Hi! Sorry, I don't get notified by commits so you need to mention me or re-request a review when you want me to take a look again. We just have two more small things and then are good to merge

quint/src/effects/simplification.ts
```diff
@@ -12,8 +12,7 @@
  * @module
  */
 
-import isEqual from 'lodash.isequal'
-import { unreachable } from '../util'
+import { isEqual, unreachable } from '../util'
```
bugarela: This is making us fail to compile
```suggestion
import { unreachable } from '../util'
```

bugarela: And we also need to add an import for `isDeepStrictEqual`

yazaldefilimone: @bugarela sorry... fix now.

PR #1812, opened 2025-11-15
fix: do not require init/step when compiling to JSON, #1584

yazaldefilimone:


quint/src/cliCommands.ts
```diff
@@ -517,6 +517,33 @@ export async function compile(typechecked: TypecheckedStage): Promise<CLIProcedu
...
   }
 
+  const hasInit = typechecked.resolver.collector.getDefinition(args.init) !== undefined
+  const hasStep = typechecked.resolver.collector.getDefinition(args.step) !== undefined
+  const flattenRequested = args.flatten == true
+
+  const targetIsTla = args.target === 'tlaplus'
+
+  // check which core definitions are missing
+  let flattenRequires: string | null = null
```
bugarela: Can we call this `missingFlattenDefs`? I find the `requires` confusing in this context

quint/src/cliCommands.ts
```diff
@@ -517,6 +517,33 @@ export async function compile(typechecked: TypecheckedStage): Promise<CLIProcedu
...
+  if (!hasInit && !hasStep) flattenRequires = 'init and step'
+  else if (!hasInit) flattenRequires = 'init'
+  else if (!hasStep) flattenRequires = 'step'
+
+  let shouldFlatten = flattenRequested && !flattenRequires
+
+  // warn user if flattening was requested but dependencies are missing (except for TLA target)
```
bugarela:
```suggestion
  // For the JSON target, we can compile even if init/step are missing.
  // In this case, we just warn the user (if --flatten=true) and compile anyway.
```

quint/src/cliCommands.ts
```diff
@@ -517,6 +517,33 @@ export async function compile(typechecked: TypecheckedStage): Promise<CLIProcedu
     return cliErr(`module ${mainName} does not exist`, { ...typechecked, errors: [], sourceCode: new Map() })
   }
 
+  const hasInit = typechecked.resolver.collector.getDefinition(args.init) !== undefined
+  const hasStep = typechecked.resolver.collector.getDefinition(args.step) !== undefined
+  const flattenRequested = args.flatten == true
```
bugarela: IIRC, we need to account for the default value here (if `args.flatten` is `undefined`, `flattenRequested` should be true). I think `yargs` should be taking care of this, but I hit the problem in the past (hence my comment), so just to be safe:
```suggestion
  // CANNOT be `args.flatten`, we need to make sure it's a boolean value
  const flattenRequested = !(args.flatten === false)
```

quint/src/cliCommands.ts
```diff
@@ -517,6 +517,33 @@ export async function compile(typechecked: TypecheckedStage): Promise<CLIProcedu
...
+  if (flattenRequested && flattenRequires && !targetIsTla) {
+    console.warn(
+      chalk.yellow(`Warning: flattening requires ${flattenRequires}, which are not defined. Disabling flattening.`)
+    )
+  }
+
+  if (!shouldFlatten && targetIsTla) {
+    console.warn(chalk.yellow('Warning: flattening is required for TLA+ output, ignoring --flatten=false option.'))
+  } else if (!shouldFlatten) {
+    return right({ ...typechecked, mainModule: main, main: mainName, stage: 'compiling' })
```
bugarela, line 547:
```suggestion
    // Early return with the original (unflattened) module and its fields
    return right({ ...typechecked, mainModule: main, main: mainName, stage: 'compiling' })
```

quint/src/cliCommands.ts
yazaldefilimone: @bugarela done

PR #1823, opened 2025-12-17
fix: write ITF traces when using the Rust backend with --out-itf

beu5a:
The Rust backend was not writing ITF trace files when --out-itf was specified, even though the trace data was available in the simulation outcome. This adds the missing call to the onTrace callback after the Rust simulation completes.

### DIFF
--- a/quint/src/quintRustWrapper.ts
+++ b/quint/src/quintRustWrapper.ts
@@ -11,7 +11,7 @@
  *
  * @module
  */
-import { QuintEx, QuintModule, QuintApp, QuintStr } from './ir/quintIr'
+import { QuintApp, QuintEx, QuintModule, QuintStr } from './ir/quintIr'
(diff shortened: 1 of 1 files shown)


quint/src/cliCommands.ts
```diff
@@ -426,6 +426,19 @@ export async function runSimulator(prev: TypecheckedStage): Promise<CLIProcedure
       prev.args.nTraces ?? 1,
       prev.args.nThreads
     )
+
+    if (options.onTrace) {
```
bugarela: Looks alright! A minor thing: as I suggested in the issue (I think you might not have seen my comment there), I think this would be better placed inside the `quintRustWrapper` file inside of `simulate` method, since the TS side happens inside `Evaluator`'s `simulate`, and that would make things more symmetrical.

PR #1828, opened 2025-12-19
feat: support --seed flag in Rust simulator

beu5a:
- Return the seed from traces so users can reproduce specific runs
- Include --backend flag in the seed output message for clarity
- Accept seed as input to reproduce previous runs deterministically

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## UNRELEASED
 
 ### Added
+
+- Added support for passing `--seed` to the Rust backend for reproducible simulations (#1827)
+
 ### Changed
 ### Deprecated
 ### Removed
 ### Fixed
+
+- Fixed ITF trace output when using the Rust backend with `--out-itf` (#1823)
+- Fixed a crash in the Rust backend when `--nthreads` exceeds `--max-samples` (#1825)
+
 ### Security
 
 ## v0.29.1 -- 2025-11-11


beu5a:
@bugarela this was tested locally, and I just added some basic e2e tests to verify that the seed is actually reproducible.

bugarela:
@beu5a I was surprised to see the tests passing with this new e2e testing you added, since as I said, our setup for integration tests in Rust is not properly done yet. So I went looking and I believe we need to change something before merging this, or this is going to break the release workflow.

From what I gathered, `rust-integration` tests are only ran in the release CI workflow (under `all-integration`, but not in the `integration` workflow that runs on each PR.

I believe that, if we try to run `all-integration` now on the CI, the tests will fail. Because we don't have a setup that builds Rust from source in the integration pipeline, so we are actually running the tests with the latest published rust evaluator version, which won't have the functionality needed for testing. So options are

1. We comment/skip this test until we fix the setup
2. We add `all-integration` to the `integration.yml` workflow (so we can check it in PRs) and try to make it work (by building the evaluator from source, and copying it to the right folder in ~/.quint)

We shouldn't completely stop running the `rust-integration` test as that has at least a basic test that checks if the CLI can download the binary. (I'm a bit confused by that test having the `--n-threads` flag which we introduced in the latest release. But we did have problems in that release that I don't remember details of now).

quint/package.json
```diff
@@ -83,6 +83,7 @@
     "apalache-integration": "txm apalache-tests.md",
     "apalache-dist": "txm apalache-dist-tests.md",
     "rust-integration": "txm rust-backend-tests.md",
+    "rust-local-integration": "txm rust-backend-local-tests.md",
     "all-integration": "npm run integration && npm run apalache-integration && npm run rust-integration",
```
bugarela: We need to change this `all-integration`, which is executed on release, to use `rust-local-integration` instead of `rust-integration`.

bugarela: Actually, we can keep the `rust-integration` and just add this new one

PR #1832, opened 2026-01-05
Improve error message from `then` failures

bugarela:
Hello :octocat: 

It is currently hard to identify what caused a `then` to fail when there are multiple `then` calls chained. This happens for two reasons:
1. The error output is truncated
2. We highlight the entire `then` expression, including the first argument which can be an expression of all previous `then` chains. For instance, this is the error for a case where `b` returns false (with the truncation problem already fixed):
```
error: [QNT513] Cannot continue in A.then(B), A evaluates to 'false'
      9:   run bFailsTest = init(true)
                            ^^^^^^^^^^
      10:     .then(a)
          ^^^^^^^^^^^^
      11:     .then(b)
          ^^^^^^^^^^^^
      12:     .then(c)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

here, there was an issue in the then from line 12 because `b` was false, and therefore the whole sub-expression from line 9 to line 11 is false. This is hard to interpret if you are not actively thinking about the semantics of `then`.

This PR fixes those 2 issues, and the same error now looks like this:
```
error: [QNT513] Cannot continue in `then` because the highlighted expression evaluated to false
      11:     .then(b)
                    ^
```

### DIFF
--- a/quint/src/runtime/impl/builder.ts
+++ b/quint/src/runtime/impl/builder.ts
@@ -456,13 +456,11 @@ function buildExprCore(builder: Builder, expr: QuintEx): EvalFunction {
         return ctx =>
           op(ctx, args).mapLeft(err => {
             // Improve reference of `then`-related errors
-            if (expr.opcode === 'then' && err.reference === undefined) {
-              if (err.code === 'QNT513') {
-                if (expr.args[0].kind === 'app' && expr.args[0].opcode === 'then') {
-                  return { ...err, reference: expr.args[0].args[1].id }
-                }
-                return { ...err, reference: expr.args[0].id }
+            if (expr.opcode === 'then' && err.code === 'QNT513' && err.reference === undefined) {
+              if (expr.args[0].kind === 'app' && expr.args[0].opcode === 'then') {
+                return { ...err, reference: expr.args[0].args[1].id }
               }
+              return { ...err, reference: expr.args[0].id }
             }
 
             return err


quint/io-cli-tests.md
```diff
@@ -1111,6 +1113,41 @@ exit $exit_code
   1) mintTwiceThenSendError:
       HOME/coin.qnt:176:5 - error: [QNT511] Test mintTwiceThenSendError returned false
       176:     run mintTwiceThenSendError = {
+               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
bugarela, line 1116: This is a consequence of removing the truncation. I'll address this in a follow-up PR, as in this case, I think the extra lines are not useful and the noise hurts the UX.

quint/src/runtime/impl/builder.ts
```diff
@@ -453,7 +453,20 @@ function buildExprCore(builder: Builder, expr: QuintEx): EvalFunction {
       // If the operator is a lazy operator, we can't evaluate the arguments before evaluating application
       if (lazyOps.includes(expr.opcode)) {
         const op = lazyBuiltinLambda(expr.opcode)
-        return ctx => op(ctx, args)
+        return ctx =>
+          op(ctx, args).mapLeft(err => {
+            // Improve reference of `then`-related errors
+            if (expr.opcode === 'then' && err.reference === undefined) {
+              if (err.code === 'QNT513') {
```
erickpintor:
```suggestion
            if (expr.opcode === 'then' && err.code === 'QNT513' && err.reference === undefined) {
```

erickpintor: Minor: looks like this 2 ifs could be combined.

erickpintor (approved):
Just one minor comment. LGTM, otherwise.

PR #1833, opened 2026-01-05
Errors about entire definitions shown on their name only

bugarela:
Hello :octocat: 

When we have some error that refers to an entire definition, we highlight (in the CLI or in the editor) the entire def, which is annoying. This change makes it so, on those cases, we only highlight the definition's name instead.


erickpintor (approved):
🐑 it!

PR #1836, opened 2026-01-05
Prevent `expect` from adding a state to the trace

bugarela:
Hello :octocat: 

This fixed a problem where `expect` would add a stuttering state to the trace, which is annoying when doing MBT. I didn't change the existing procedure of actually shifting the state before evaluating the condition, which simplifies things a lot. I instead added a `drop` to remove the artificially added state from the trace.

It also deletes a test-related file that was added by accident in a previous PR.


erickpintor (approved):
Thank you! 🐑 it!

PR #1837, opened 2026-01-05
Simple destructuring for tuples and records

bugarela:
Hello :octocat: 

This adds simple (one level) destructuring for tuples and records:
```
val (foo, _, bar) = my_tuple
val { foo, bar } = my_record
```

We desugar those into something like:
```
val quintDestructTemp1 = my_tuple
val foo = quintDestructTemp1._1
val bar = quintDestructTemp1._3
```

so we don't have to worry about Apalache or any of the evaluators. 

This is similar to what we do for unpacking (i.e. `Set((1, 2), (3, 4)).map(((x, y)) => x + y)`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed ITF trace output when using the Rust backend with `--out-itf` (#1823)
 - Fixed a crash in the Rust backend when `--nthreads` exceeds `--max-samples` (#1825)
 - Fixed a problem where the REPL error reporting showed an incorrect location (#1835)
-- Fixed a problem where `expect` would add a stuttering state to the trace (#1836)
+- Fixed a problem where `expect` would add a stuttering state to the trace (#1846)
+- Fixed a problem where tests without and `expect` at the end would be missing
+  the last state in the ITF trace (#1846)
+- Fixed an edge case with nested operators that had operators as arguments could cause a runtime issue (#1843)
 
 ### Security
 
--- a/evaluator/src/evaluator.rs
+++ b/evaluator/src/evaluator.rs
@@ -573,8 +573,9 @@ impl<'a> Interpreter<'a> {
                     };
                     let compiled_expr = self.compile(expr);
                     CompiledExpr::new(move |env| {
+                        let saved = cached_value.replace(None);
                         let result = compiled_expr.execute(env);
-                        cached_value.replace(None);
+                        cached_value.replace(saved);
                         result
                     })
                 }
--- a/quint/cli-tests.md
+++ b/quint/cli-tests.md
@@ -410,3 +410,11 @@ Tests that `quint run` works with nested setOfMaps and oneOf.
 
 <!-- !test check 1736 -->
     quint run testFixture/bug1736setOfMaps.qnt
+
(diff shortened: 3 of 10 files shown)


erickpintor:
Feature looks great!

I wonder, thought, if we should fail if destructuring isn't complete. Users decomposing tuples and records to use their fields to compose new values might silently forget about new fields if the tools don't let them know. Rust is a good source of inspiration here: it will fail to compile if you miss a field but, there's a wildcard if partial destructuring is intended.

Any thoughts on that?

bugarela:
> I wonder, thought, if we should fail if destructuring isn't complete. Users decomposing tuples and records to use their fields to compose new values might silently forget about new fields if the tools don't let them know. Rust is a good source of inspiration here: it will fail to compile if you miss a field but, there's a wildcard if partial destructuring is intended.
> 
> Any thoughts on that?

I can't really think of a case where forgetting a field would be a problem in the context of Quint. Maybe if you forget to validate a field, but then we already have this problem with the current and most common field access (we won't complain if a field is never read in a function), so I'm biased to believe it will be fine.

Then, I feel like the friction of people having to learn that they need to type something like `, ..` is not worth the benefit (Quint needs to be easier to learn than a programming language, since it's more like a secondary tool to most people), and people will find it more annoying than useful.

I'm happy to discuss this further if you have a different perspective @erickpintor

erickpintor:
The case that I feel like it's most problematic is when re-composing states where previous values are supposed to be carried forward after some computation. For example, consider a hypothetical code:

```
type State = {
  some_field: int
}
...
pure def handle_something(s: State): State = {
  val { some_field } = s
  { some_field: some_field + 1 }
}
```

If you later add a new field to `State`, it can become quite annoying to locate all places where you might have destructed `State` into its inner parts and make sure they are also carried forward.

I understand that currently we don't enforce the same rules on field projection and the same behavior exists today if you don't use the spread operator (`...s`). At the same time, destructuring is a new feature so, I feel like we should make a call here between the trace-offs of language ergonomics versus language assistance specially considering that making the feature more strict later is a breaking change.

bugarela:
@erickpintor I don't really understand the advantage in your example, I might be missing something. What I understand is: the compiler will complain that now `handle_something` should return `State`, but a field is missing. So you either have to
1. Add `{ ...s, some_field: some_field + 1 }` and never worry about it again
2. Add your new field to the destructuring statement and use it on the record construction

Either way, seems like your chances of missing something important or level of annoyance would be the same between the two approaches, so I don't understand what you mean here.

Also, while typing this, I realized that the more permissive approach might be better for polymorphic records with row types like Quint has. In Quint, you can have an operator that accepts both:
- { some_field: int, other_field: bool }
- { some_field: int }
by having `f({ some_field: int | r }): { some_field: int | r }` in the type signature. In this case, the behavior of whether we should complain about missing fields in destructuring gets more complex both for us to implement and for users to understand the feedback. What if `r` has another shared field, and the user expects us to warn them about this field not being destructured but we don't due to the type signature? I think it gets too complicated.

erickpintor:
You're right. This ends up being more problematic in untyped languages, which I believe is where I've experienced most problems with destructuring before. I think it's fine to move on, specially considering the complexity of adding struct destructuring for polymorphic types.

erickpintor (approved):
🐑 it.

PR #1839, opened 2026-01-07
Revert "Prevent `expect` from adding a state to the trace"

bugarela:
Hello :octocat: 

Reverts informalsystems/quint#1836

Yesterday we realized the results of this were not as we expected, and then I noticed I made a mistake when reproducing the problem + in the unit test setup that "confirmed" my changed work.

Turns out I'm actually not able to reproduce the original problem, so we are reverting this until we have a reproduction of the issue to guide the solution better.


erickpintor (approved):
un- :sheep:  it :cry:

PR #1841, opened 2026-01-15
Extract Apalache verification logic

beu5a:
Hello :octocat:

This is a small refactoring PR to extract the Apalache verification logic (for the `quint verify` command) from `cliCommands.ts`. 

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,8 @@ jobs:
     defaults:
       run:
         working-directory: ./quint
+    permissions:
+      contents: write
     steps:
       - name: Checkout
         uses: actions/checkout@v4


quint/src/quintVerifier.ts
```diff
@@ -1,39 +0,0 @@
...
- * Verifies the configuration `config` by model checking it with the Apalache server
- *
- * @param serverEndpoint
- *   a server endpoint
- *
- * @param config
- *   an apalache configuration. See https://github.com/apalache-mc/apalache/blob/main/mod-infra/src/main/scala/at/forsyte/apalache/infra/passes/options.scala#L255
- *
- * @returns right(void) if verification succeeds, or left(err) explaining the failure
- */
```
bugarela, line 30: It seems like this doc comment was lost when moving things. Can we add it back?

quint/src/verify.ts
bugarela, line 1: I think the name of this file is a bit misleading. What we actually end up having is two wrappers: one for Apalache, and other for TLC. I think the existing `apalache.ts` file and the `tlc.ts` file you created on the other PR are the actual wrappers, while this is more the verification functionality (which is kinda similar between Apalache and TLC, so I get having both verify* functions on the same file). Maybe we can call this `verify.ts`

beu5a: I agree. I fixed it in the latest commit.

PR #1843, opened 2026-01-15
Fix nested calls with let

bugarela:
Hello :octocat: 

We found a really specific issue! Since we don't have recursion, we normally wouldn't have a case of something like
```
let foo = ...
let bar = ...
foo + bar
```

where `foo` and `bar` are exactly the same **definition** (same id in the IR and lookup table, etc). However, by passing an operator `f` that calls `p` to a parameterized operator `p` (like in the regression test in this PR), when `p` has a let definition, can cause this nested let problem where both definitions are actually the same in the IR, and therefore share a cache.

I'm not 100% positive that this fix is complete, and that no combination of `foo` and `bar` usage can break it. I tried a few things I thought could break it, and none of them broke it.

Also, the first solution I had was not saving the value from the cache, but rather simply resetting it to `None` before (and after) the evaluation. I think this could cause unnecessary re-evaluation of the let value, so I'm taking the most conservative approach of saving the value, but I could not find a scenario that caused the re-evaluation.

<details>
<summary>Benchmark results (against main)</summary>

```
     Running benches/evaluator.rs (target/release/deps/evaluator-071ace652ba1e11d)
Gnuplot not found, using plotters backend
evaluator/tictactoe     time:   [425.42 µs 426.51 µs 427.69 µs]
                        change: [-1.6153% -1.2956% -0.9661%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
evaluator/JMT           time:   [27.822 ms 27.906 ms 27.995 ms]
                        change: [-1.3429% -0.9388% -0.5348%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe

     Running benches/simulation.rs (target/release/deps/simulation-956888a9f3f681ab)
Gnuplot not found, using plotters backend
tictactoe/rust          time:   [780.16 ms 781.11 ms 782.12 ms]
                        change: [-1.3790% -0.9960% -0.6287%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) low severe
  1 (10.00%) low mild
tictactoe/typescript    time:   [2.6874 s 2.6954 s 2.7032 s]
                        change: [+0.6393% +0.9762% +1.3646%] (p = 0.00 < 0.05)
                        Change within noise threshold.

     Running benches/tuples.rs (target/release/deps/tuples-62b7f5d5a63d2dde)
Gnuplot not found, using plotters backend
tuples/rust             time:   [955.40 ms 961.46 ms 968.26 ms]
                        change: [-0.3842% +0.5629% +1.4973%] (p = 0.29 > 0.05)
                        No change in performance detected.
Benchmarking tuples/typescript: Warming up for 10.000 s
Warning: Unable to complete 10 samples in 10.0s. You may wish to increase target time to 60.7s.
tuples/typescript       time:   [6.0418 s 6.0667 s 6.0911 s]
                        change: [-0.4067% +0.4678% +1.2787%] (p = 0.32 > 0.05)
                        No change in performance detected.
```

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed ITF trace output when using the Rust backend with `--out-itf` (#1823)
 - Fixed a crash in the Rust backend when `--nthreads` exceeds `--max-samples` (#1825)
 - Fixed a problem where the REPL error reporting showed an incorrect location (#1835)
+- Fixed a problem where `expect` would add a stuttering state to the trace (#1846)
+- Fixed a problem where tests without and `expect` at the end would be missing
+  the last state in the ITF trace (#1846)
 - Fixed an edge case with nested operators that had operators as arguments could cause a runtime issue (#1843)
 
 ### Security
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -1020,31 +1020,24 @@ exit $exit_code
 ### Test does not skip assignments (#1133)
 
 See: https://github.com/informalsystems/quint/issues/1133
-
-FIXME: fix the traces found by the simulator once #1133 is resolved.
+Fixed in: https://github.com/informalsystems/quint/pull/1846
+This is now a regression test
 
 <!-- !test in test1133 -->
 ```
-output=$(quint test --match='(t1|t2)' --out-itf='out_{seq}_{test}.itf.json' \
-  ./testFixture/simulator/lastActionInRun.qnt)
-exit_code=$?
-echo "BEGIN"
-# This test should have 3 states (FIXME: it does not!)
-cat out_0_t1.itf.json | jq '.states' | grep "s" | wc -l | grep 3
+quint test --match='(t1|t2)' --out-itf='out_{seq}_{test}.itf.json' \
+  ./testFixture/simulator/lastActionInRun.qnt > /dev/null
+cat out_0_t1.itf.json | jq '.states | length'
 rm out_0_t1.itf.json
-# This test should have 4 states (FIXME: it does not!)
-cat out_1_t2.itf.json | jq '.states' | grep "s" | wc -l | grep 4
+cat out_1_t2.itf.json | jq '.states | length'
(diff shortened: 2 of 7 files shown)


evaluator/src/evaluator.rs
```diff
@@ -573,8 +573,13 @@ impl<'a> Interpreter<'a> {
                     };
                     let compiled_expr = self.compile(expr);
                     CompiledExpr::new(move |env| {
-                        let result = compiled_expr.execute(env);
+                        let saved = cached_value.borrow().clone();
```
bugarela: My understanding is that this is cloning `None` almost all the time (I guess, except for the edge case of the bug being fixed here). @erickpintor do you think this is ok, or should we maybe do something like "if it's not None, then we clone"?

Also, this is adding the cloning overhead to every single `let`. An alternative would be to just write `None` to the cache and risk having to re-evaluate the let expression in the rare scenario of this bug. This feels shady though, so I tend to prefer the more homogeneous solution.

erickpintor: I don't think cloning is a problem here, specially if cloning `None` is the most common case. The more expensive operations that we do is borrowing from the `RefCell` both by calling `borrow()` and `replace()`.

The thing that calls my attention is the fact that we're cloning a value because we don't have ownership over it, but then we replace the value with `None`, which returns ownership over the previously held value back. So, in fact, we don't need borrowing at all. We can simply hold ownership over to the previous value after replace.

```rust
CompiledExpr::new(move |env| {
    let saved = cached_value.replace(None);
    let result = compiled_expr.execute(env);
    cached_value.replace(saved);
    result
})
```

bugarela: Ah, I didn't realize `replace` returns the value! That's much better, thanks

bugarela: Updated!

erickpintor (approved):
🐑 it.

PR #1844, opened 2026-01-16
Add TLC as a backend to quint verify

beu5a:
Hello, 

  This PR adds TLC as an alternative verification backend alongside Apalache.

  ### Usage:
  `quint verify --backend tlc spec.qnt`

  - `--backend tlc` - use TLC instead of Apalache
  - `--tlc-config` - path to TLC runtime configuration (JSON)

  TLC runs locally using the Apalache JAR (no server required for TLC itself, but Apalache is still needed for Quint-to-TLA+ compilation).
  
  
### DIFF
--- a/quint/verification-tests.md
+++ b/quint/verification-tests.md
@@ -558,9 +558,7 @@ quint verify --backend tlc --invariant inv --max-steps=10 ./testFixture/apalache
 
 Test that TLC reports errors when given invalid configuration.
 
-<!-- !test in TLC config error -->
+<!-- !test check TLC config error -->
 ```
 quint verify --backend tlc --init=nonExistentInit ./testFixture/apalache/tlcConfigError.qnt 2>&1 | grep 'error:'
 ```
-
-<!-- !test exit 1 -->
--- a/quint/testFixture/apalache/tlcCounter.qnt
+++ b/quint/testFixture/apalache/tlcCounter.qnt
@@ -5,8 +5,10 @@ module tlcCounter {
 
   action init = counter' = 0
 
-  action step = counter' = counter + 1
+  action step = all {
+    counter < 3,
+    counter' = counter + 1
+    }
 
-  // Invariant that always holds for bounded model checking
   val inv = counter >= 0
 }


quint/src/quintMCWrapper.ts
```diff
@@ -5,18 +5,81 @@
...
+    return cliErr('Failed to convert init to predicate', {
+      ...verifying,
+      errors: mainModule.value.map(mkErrorMessage(prev.sourceMap)),
+    })
+  }
+
+  const verifyingFlat = { ...prev, modules: [mainModule.value] }
+  const parsedSpec = outputJson(verifyingFlat)
+
+  console.log(chalk.green('[TLC]') + ' Compiling to TLA+...')
```
bugarela: We should only log under a high enough verbosity (maybe 3 or more so we log by default, but we shouldn't log in verbosity 0 for instance)

quint/src/quintMCWrapper.ts
```diff
@@ -5,18 +5,81 @@
...
+  const tlaResult = await compileToTlaplus(args.serverEndpoint, args.apalacheVersion, parsedSpec, verbosityLevel)
+
+  if (tlaResult.isLeft()) {
+    return cliErr('error', { ...verifying, errors: tlaResult.value.errors })
+  }
+
+  const [, invariantsList] = getInvariants(args)
+  const tlcRuntimeConfig = loadTlcConfig(args.tlcConfig)
+
+  console.log(chalk.green('[TLC]') + ' Running TLC model checker...\n')
```
bugarela: same here about logging

quint/src/quintMCWrapper.ts
bugarela: People might thing there is a bug if they see "Apalache" in the logs when giving `--backed tlc`, so explicitly adding this to the log might help
```suggestion
  console.log(chalk.green('[TLC]') + ' Compiling to TLA+ (via Apalache)...')
```

bugarela (changes requested):
I tested it locally and it works great! However, we should add at least 2 integration tests: one for success, and other for a violation being found. We can add them to `quint/apalache-tests.md` since that file is already ran in a setup where Apalache is installed

quint/verification-tests.md
bugarela, line 1: Should we rename this to `verification-tests.md`? (And rename references to it on the scripts, docs, etc)

PR #1846, opened 2026-01-16
Fix `expect` trace behavior

bugarela:
Hello :octocat: 

After trying to fix this (https://github.com/informalsystems/quint/pull/1836) and breaking other things instead, this is now the right fix. This fixes two issues:
1. When a run that has an expect (like [this](https://github.com/informalsystems/emerald/blob/049c50f7523e36919b040b805685774de4cdba70/specs/emerald_tests.qnt#L56)) was called from another run (like [this](https://github.com/informalsystems/emerald/blob/049c50f7523e36919b040b805685774de4cdba70/specs/emerald_tests.qnt#L165)), we'd make and extra `shift()` and **duplicate** a state: the first `shift()` done inside `expect`, and the second `shift()` done by `then`.
2. When looking at an ITF trace from a test (run) that doesn't end with `expect`, we would be **missing** the final state.

PS: `shift()` is the method that moves from one state to a new one, registering a new state to the trace.

We have two problems in the code that handles this:
1. `expect` makes a `shift()` while it shouldn't (this is what I had fixed before and reverted)
2. We *don't* make `shift()` after evaluating a test.

So, the two problems were almost cancelling each other out in some cases, and this is why my fix for `1` made things works instead of better. Now, I'm fixing both problems. The first 3 commits here are copies of the ones in my previous fix.

This actually fixes https://github.com/informalsystems/quint/issues/1133 (which seems to have been closed by accident)

### DIFF
--- a/quint/io-cli-tests.md
+++ b/quint/io-cli-tests.md
@@ -1020,31 +1020,24 @@ exit $exit_code
 ### Test does not skip assignments (#1133)
 
 See: https://github.com/informalsystems/quint/issues/1133
-
-FIXME: fix the traces found by the simulator once #1133 is resolved.
+Fixed in: https://github.com/informalsystems/quint/pull/1846
+This is now a regression test
 
 <!-- !test in test1133 -->
 ```
-output=$(quint test --match='(t1|t2)' --out-itf='out_{seq}_{test}.itf.json' \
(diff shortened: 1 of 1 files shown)


erickpintor (approved):
🐑 it!

PR #1847, opened 2026-01-16
[rust] Move trace collection to inside Env

bugarela:
Hello :octocat: 

This PR moves trace collection from the simulator level into the `Env` struct, achieving parity with the TypeScript evaluator and unblocking QNT-47 (adding support for `quint test` in the Rust backend).

## The Problem

In the current Rust evaluator, traces are only collected at the simulation level (after each step in `simulator.rs`). However, for proper trace tracking during tests and runs, we need to collect trace states whenever the `then` operator advances the state.

In the TypeScript evaluator, trace collection happens inside the context's `shift()` method, which is called by the `then` operator. This ensures traces are properly collected regardless of where state transitions happen, including in complex expressions with nested `then` operators.

## The Solution

- Added a `trace` collector to the `Env` struct
- Updated `Env::shift()` to record states after shifting variables
- Modified `expect` to drop the state from the trace (avoiding duplicate states when rolling back)
- Updated simulator to use the trace from `Env` instead of manually building it
- Added regression test to verify `expect` doesn't duplicate states


quint/rust-backend-local-tests.md
```diff
@@ -115,3 +115,21 @@ rm /tmp/debug_test.qnt
...
+```
+quint run --backend=rust --out-itf=expect-run-rust.itf.json --max-steps=20 --seed=42 --verbosity=0 \
+  ./testFixture/expectNoStateDuplication.qnt
+cat expect-run-rust.itf.json | jq '.states | length'
+rm expect-run-rust.itf.json
+```
+
+<!-- !test out rust backend expect no duplicate states in run -->
+```
+42
```
bugarela, line 134: This would give `21` before the changes, since the `then` inside each `step` was not recording an extra state in the trace.

evaluator/src/evaluator.rs
```diff
@@ -82,15 +82,19 @@ pub struct Env {
 
     // The random number generator, used for nondeterministic choices. This is stateful.
     pub rand: Rand,
-    // TODO: trace recorder (for --verbosity) and trace collector (for proper
-    // trace tracking in runs)
+
+    // The trace collector, used to track state transitions during evaluation.
+    // This is collected whenever `then` advances the state by calling `shift()`.
+    pub trace: Rc<RefCell<Vec<Value>>>,
```
erickpintor: It seems like every place we use `trace` we have full ownership of the `Env` struct. This can be simplified to just: `trace: Vec<Value>` and save on allocation and ref-count overhead.

evaluator/src/simulator.rs
```diff
@@ -92,12 +93,12 @@ impl ParsedQuint {
...
-                interpreter.shift();
-
-                trace.push(interpreter.var_storage.borrow().as_record());
+                // Shift the state and record it in the trace
+                env.shift();
 
                 if !invariant.execute(&mut env)?.as_bool() {
-                    trace_lengths.push(trace.len());
+                    trace_lengths.push(env.trace.borrow().len());
+                    let trace = env.trace.borrow().clone();
```
erickpintor:
```suggestion
                    let trace = std::mem::take(&mut env.trace);
```

We could transfer ownership from the env to the binding here and avoid the clone cost since we expect this branch to return (no more uses of `env.trace`.

evaluator/src/simulator.rs
```diff
@@ -123,11 +124,12 @@ impl ParsedQuint {
...
                     // drop the run. Otherwise, we would have a lot of false
                     // positives, which look like deadlocks but they are not.
-                    trace_lengths.push(trace.len());
+                    trace_lengths.push(env.trace.borrow().len());
                     break;
                 }
             }
-            trace_lengths.push(trace.len());
+            trace_lengths.push(env.trace.borrow().len());
+            let trace = env.trace.borrow().clone();
```
erickpintor:
```suggestion
            let trace = std::mem::take(&mut env.trace);
```

We can also move ownership here since we expect the next iteration of the loop to have a clean vector anyway.

evaluator/src/simulator.rs
```diff
@@ -79,7 +79,8 @@ impl ParsedQuint {
         for sample_number in 1..=samples {
             reporter.next_sample();
             let seed = env.rand.get_state();
-            let mut trace = Vec::with_capacity(steps + 1);
+            // Clear the trace for this sample
+            env.trace.borrow_mut().clear();
```
erickpintor: We could remove this clean up since the very last use of `env.trace` moves it out of `env` and leave it empty already. See the uses of `std::mem:take(..)` below.

PR #1851, opened 2026-01-19
Force dowloading the Rust evaluator before release tests

bugarela:
Hello :octocat: 

We're hitting [problems on CI](https://github.com/informalsystems/quint/actions/runs/21139562605/job/60792590807) due to Rust evaluator not being available by the time we run tests, and then the tests are not prepared to handle downloading.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,7 +34,7 @@ jobs:
           # Wait for the server to start
           sleep 5
           # Download the rust evaluator before the tests
-          quint run testFixture/simulator/gettingStarted.qnt --backend rust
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -33,6 +33,8 @@ jobs:
           _build/apalache/bin/apalache-mc server &
           # Wait for the server to start
           sleep 5
+          # Download the rust evaluator before the tests
+          quint run testFixture/simulator/gettingStarted.qnt --backend rust
```
erickpintor: Make it a bit faster?

```suggestion
          quint run testFixture/simulator/gettingStarted.qnt --backend rust --max-samples=1 --max-steps=1
```

PR #1857, opened 2026-01-20
Fix #1327

### LINKED ISSUE #1327: Broad error location
Take this spec:
https://github.com/cosmos/interchain-security/blob/79e8044d0d1337be0f63de9cc4e63f642a6727b3/tests/mbt/model/ccv.qnt#L335

There is an error in the spec, see this:
Screenshot 2024-01-09 at 11 35 29

to me, it feels like
a) the error location is too broad - maybe my assumption is too naive, but it seems at least in this case it's possible to pinpoint a statement where the type mismatch occurs (it's that the return type doesn't fit, in my case, so only the return statement should probably be underlined)
b) the error text is too broad - if two types mismatch, it seems like it should be possible to omit all the parts in which they agree. in this case, the return type has the "consumerAddrsToPrune" field of a wrong type (and everything else matches), but that's not possible to figure out easily from the error message. imo specifically the part of the types that mismatches should be the very first thing in the message, and the full types can maybe be given later.

I'm not very familiar with PL theory and types, so I'm not sure how possible it is, but I think the error-squashing experience here is not great right now, so maybe there's something that can be improved.

arnaudgelas:

### DIFF
--- a/quint/src/types/constraintSolver.ts
+++ b/quint/src/types/constraintSolver.ts
@@ -186,15 +186,23 @@ export function unifyRows(table: LookupTable, r1: Row, r2: Row): Either<ErrorTre
       // This call will fit in the above case of row unification
       const tailSubs = unifyRows(table, { ...ra, fields: uniqueFields1 }, { ...rb, fields: uniqueFields2 })
 
-      // Sort shared fields by field name, and get their types
-      const fieldTypes: [QuintType, QuintType][] = sharedFieldNames.map(n => {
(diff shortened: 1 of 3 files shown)


bugarela:
Please rebase with main, this includes commits from you other PR. 
And can you include a screenshot of how the error looks like now so we can compare to the screenshot in https://github.com/informalsystems/quint/issues/1327?

arnaudgelas:
✅ rebased onto main
✅ before vs after in the commit message

PR #1860, opened 2026-01-20
 Implements `test-from-stdin` command for Rust

beu5a:
Hello, 

This PR Implements `test-from-stdin` command for the Rust backend, following the same pattern as `simulate-from-stdin`. This is a subtask from the broader feature of adding `quint test` support to the Rust backend. 
This introduces a new `tester.rs` module and a new `test-from-stdin` command in `main.rs`

      

### DIFF
--- a/evaluator/src/tester.rs
+++ b/evaluator/src/tester.rs
@@ -49,7 +49,7 @@ impl TestCase {
 
         let mut errors = Vec::new();
         let mut nsamples = 0;
-        let mut traces = Vec::new();
+        let mut trace = None;
 
         for _ in 0..max_samples {
             let prev_rng_state = env.rand.get_state();
@@ -59,7 +59,6 @@ impl TestCase {
 
             match compiled_test.execute(&mut env) {
                 Ok(result) => {
-                    // Shift the state and record it in the trace
                     env.shift();
 
                     if !result.as_bool() {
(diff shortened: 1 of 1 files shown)


evaluator/src/tester.rs
```diff
@@ -0,0 +1,102 @@
...
+    ) -> TestResult {
+        let test_name = &self.name;
+
+        let mut interpreter = Interpreter::new(&self.table);
+        let mut env = Env::with_rand_state(interpreter.var_storage.clone(), seed);
+        let compiled_test = interpreter.compile(&self.test);
+
+        let mut errors = Vec::new();
+        let mut nsamples = 0;
+        let mut seen_rng_states = HashSet::new();
```
bugarela: I don't think we need to keep a set here. In typescript, we simply compare the state with the seed:
https://github.com/informalsystems/quint/blob/46c9b9e89732b47fba1e5fafc0c5ccd845377154/quint/src/runtime/impl/evaluator.ts#L402

evaluator/src/tester.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        for _ in 0..max_samples {
```
bugarela, line 54: In the future, we should support parallelization of this like we do with simulation.

bugarela (approved):
LGTM, just a small simplification

bugarela:
This is still marked as Draft. Should it be "Ready for review"?

evaluator/fixtures/runs.qnt
erickpintor, line 1: Nit: aren't we using 2 spaces instead of 4 everywhere else?

evaluator/src/tester.rs
```diff
@@ -0,0 +1,102 @@
...
+                            status: TestStatus::Failed,
+                            errors: vec![error],
+                            seed,
+                            nsamples,
+                        };
+                    }
+                }
+                Err(e) => {
+                    errors.push(e);
+                    break;
```
erickpintor, line 99: If we only ever going to return one error, does it make sense to collect them in a vector? It seems like we can simplify `TestResult` to have an optional error and return here instead.

beu5a: The TS side expects an array of errors [(testing.ts)](https://github.com/informalsystems/quint/blob/71d0fd19b95a769232f193a39bb192d89686d248/quint/src/runtime/testing.ts#L47) so having a vector of errors on the rust side too felt consistent with the codebase.

erickpintor (approved):
🐑 it.

PR #1862, opened 2026-01-21
Add support for `quint test` via the rust backend

beu5a:
Hey,
this PR build on top of #1860 and fully implements the feature: `quint test --backend = rust`.
Currently, the tests via the rust backend don't support high verbosity outputs. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for `quint test --backend=rust` to run tests using the Rust backend (#1862)
 
 ### Changed
+
+- Moved LSP-server compilation and analysis to its own worker thread (1876)
+- Improved memory allocation profile of LSP-server log messages (#1877)
+
 ### Deprecated
 ### Removed
 ### Fixed
+
+- Fixed LSP-server stopping if `LOG_LEVEL` is not `NONE` (#1874)
+- Fixed LSP-server crashing upon console.{log, error} messages (#1875)
+
 ### Security
 
 ## v0.30.0 -- 2026-01-19
--- a/vscode/quint-vscode/server/src/complete.ts
+++ b/vscode/quint-vscode/server/src/complete.ts
@@ -26,7 +26,7 @@ import {
   walkExpression,
 } from '@informalsystems/quint'
 import { findMatchingResults } from './reporting'
-import { logger, overrideConsole } from './logger'
+import { isEnabled, logger, overrideConsole } from './logger'
 import JSONbig from 'json-bigint'
 
 overrideConsole()
@@ -51,15 +51,18 @@ export function completeIdentifier(
       const refId = table.get(declBigInt)?.id ?? declBigInt
       const type = analysisOutput.types.get(refId)?.type
 
-      logger.info(
-        `DeclId: ${declId} (BigInt: ${declBigInt.toString()}), RefId: ${refId.toString()}, Type: ${JSON.stringify(
-          type,
-          (_, val) => (typeof val === 'bigint' ? val.toString() : val)
-        )}`
-      )
+      if (isEnabled('INFO')) {
+        logger.info(
+          'DeclId: %s (BigInt: %s), RefId: %s, Type: %s',
+          declId,
+          declBigInt.toString(),
+          refId.toString(),
+          JSON.stringify(type, (_, val) => (typeof val === 'bigint' ? val.toString() : val))
+        )
+      }
       return type
     } catch (error) {
-      logger.error(`Error processing declId ${declId}: ${error}`)
+      logger.error('Error processing declId %s: %s', declId, error)
       return undefined
     }
   })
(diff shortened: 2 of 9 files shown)


evaluator/src/tester.rs
```diff
@@ -45,7 +45,27 @@ impl TestCase {
...
-        let compiled_test = interpreter.compile(&self.test);
+
+        // Look up the test definition and compile it (not just the expression)
+        // This ensures instance overrides are applied via compile_under_context
+        let test_def = match self.table.get(&self.test_def_id) {
+            Some(def) => def,
+            None => {
+                let error = QuintError::new(
+                    "QNT404",
+                    &format!("Test definition with id {} not found", self.test_def_id),
```
erickpintor: This error shows to users, right? Is the reported ID significant information to them? It might be confusing.

quint/src/cliCommands.ts
```diff
@@ -307,8 +309,48 @@ export async function runTests(prev: TypecheckedStage): Promise<CLIProcedure<Tes
...
+        const tableEntry = prev.table.get(def.id)
+        console.error(`\nDEBUG table entry for test id=${def.id} (${testName}):`)
+        console.error(`  kind: ${tableEntry?.kind}`)
+        console.error(`  name: ${tableEntry?.name}`)
+        console.error(`  importedFrom: ${tableEntry?.importedFrom?.kind}`)
+        console.error(`  namespaces: ${JSON.stringify(tableEntry?.namespaces)}`)
+        if (tableEntry?.importedFrom?.kind === 'instance') {
+          console.error(`  instance has ${tableEntry.importedFrom.overrides.length} overrides`)
+        }
+      }
```
erickpintor: Do we need to keep this debug logs? Looks like users can accidentally hit it.

evaluator/src/tester.rs
beu5a: I agree, I will change this to report the test name.

quint/src/cliCommands.ts
```diff
@@ -307,8 +309,35 @@ export async function runTests(prev: TypecheckedStage): Promise<CLIProcedure<Tes
...
-  const evaluator = new Evaluator(prev.table, newTraceRecorder(verbosityLevel, options.rng, 1), options.rng)
-  const results = testDefs.map((def, index) => evaluator.test(def, options.maxSamples, index, options.onTrace))
+  let results: TestResult[]
+  if (prev.args.backend === 'rust') {
+    const quintRustWrapper = new QuintRustWrapper(verbosityLevel)
+    // Add test definitions from definitionsByModule to the lookup table
+    // This ensures instance metadata (importedFrom, namespaces) is preserved
+    testDefs.forEach(def => {
+      prev.table.set(def.id, def)
+    })
```
bugarela: I don't understand why this is needed. `prev.table` should have the exact same entries here. Also, I don't think we lookup the def id (as we are not resolving a name in this situation).

quint/src/cliCommands.ts
```diff
@@ -307,8 +309,35 @@ export async function runTests(prev: TypecheckedStage): Promise<CLIProcedure<Tes
...
+    results = []
+    for (const def of testDefs) {
+      if (def.kind !== 'def') {
+        throw new Error(`Expected test definition, got ${def.kind}: ${def.name}`)
+      }
+      const testName = nameWithNamespaces(def.name, List(def.namespaces))
+
+      const result = await quintRustWrapper.test(
+        def.id,
```
bugarela: Hmm, maybe this is why we need the table manipulation from above: we are passing the id here instead of the entire def.

Maybe this is fine, but I guess it could simplify things if we passed the whole `LookupDefinition` object as we do with the typescript one.

Also, for the future, we might want to have a single rust process that gets initialized with a table and then runs all tests, instead of spinning up one process per test, sequentially, and passing the table every time. This should be on another iteration tho (not this PR).

Back to the paramaters: I think we could benefit from having a matching interface to the Evaluator's `test()` here.
- pass the whole definition
- `testName` is not computed here, but inside the wrapper's function (or even in the rust side)
- We need `index` somewhere, in order to produce the proper ITF trace files.
- Not sure yet why we are passing the RNG state here, but it is weird to me.

quint/src/quintRustWrapper.ts
```diff
@@ -170,6 +141,54 @@ export class QuintRustWrapper {
...
+      source: source,
+      witnesses: witnesses,
+      nruns: nruns,
+      nsteps: nsteps,
+      ntraces: ntraces,
+      nthreads: nthreads,
+      seed: seed,
+    }
+
+    const output = await this.runRustEvaluator(
```
bugarela, line 187: Beautiful ✨

quint/rust-backend-integration-tests/quint-test.md
```diff
@@ -423,11 +316,16 @@ exit $exit_code
                                 ^^^^^^^^^^^
     Use --seed=0x1 --match=initFailsTest to repeat.
   3) lastActionFailsTest:
-      HOME/thenErrorMessages.qnt:18:7 - error: [QNT511] Test lastActionFailsTest returned false
+      HOME/thenErrorMessages.qnt:18:29 - error: [QNT511] Test lastActionFailsTest returned false
```
bugarela: This seems like a regression to me. We should report the failure on the definition's ID so we get the highlighting to happen on the definition's name rather then it's whole body (which can be several lines)

quint/src/cliCommands.ts
bugarela: I guess it makes sense that we need to pass some seed here, but in the simulator we just do `prev.args.seed`. My concern here is that the state of this rng is not changing since we are not using it. I guess we don't need it to change, and it is fine to use the same seed for all tests, but then using the seed directly is clearer.

bugarela: Additional info as I read the linear card explaining your solution to the name resolution problem: you can still `compile_def` in the rust side if you pass the whole `LookupDefinition`.

beu5a: I agree, passing the whole `LookupDefinition ` is so much better !

PR #1863, opened 2026-01-21
Organize tests and duplicate for rust backend

bugarela:
Hello :octocat: 

This duplicates our test suite so we run all the same tests in Rust. This should replace the TS tests soon when we deprecate that backend.

I'm splitting them by command now so we can be more efficient, but it actually seems like a good idea permanently. We can later add a command that runs all of them, but for now we should run them individually and we are not going to add this to CI before the features are ready.

I also added some tests we only had in the rust side to the `io-cli-tests.md` (TS) for completeness.

### DIFF
--- a/evaluator/fixtures/runs.qnt
+++ b/evaluator/fixtures/runs.qnt
@@ -1,0 +1,32 @@
+module simple_test {
+  var x: int
+
+  action init = {
+    x' = 0
+  }
+
+  action step = {
+    x' = x + 1
+  }
+
+  // Test that returns true (passes)
+  run passingTest = {
+    1 + 1 == 2
+  }
+
+  // Test that returns false
+  run failingTest = {
+    1 + 1 == 3
+  }
+
+  // Test that uses assert and fails (fails with QNT508)
+  run failingAssertTest = {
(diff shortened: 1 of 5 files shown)


erickpintor (commented):
Nit: `rust-backend-integration-tests` is kinda long. Could we maybe migrate to something shorter like `integration-tests` or simply `it`?

Also, we should add a ticket to migrate CI to run this tests and remove the ones at `rust-backend{-local}-tests.md` once we're done implementing all functionality required.

bugarela:
@erickpintor I would like to make all CI-related changes in a single pass (once we are done, like you suggest). If I call the folder something too simple now, it will be confusing as most of our integration tests are out of this folder. And if I move the other integration tests to this folder too, I'll have to update CI-related things, docs that refer to these files, etc. I'd rather do it only once when we have the more definitive suite, but I could also do it now too. WDYT?

erickpintor:
I think it's fine to do a single pass at the CI later. Let's just document this on the ticket so we don't forget.

erickpintor (approved):
🐑 it.

PR #1871, opened 2026-01-23
Fix various security issues

arnaudgelas:
- [ ] I have read and I understand the [Note on AI-assisted contributions](https://github.com/informalsystems/quint/blob/main/CONTRIBUTING.md#note-on-ai-assisted-contributions)

### DIFF
--- a/quint/src/tlc.ts
+++ b/quint/src/tlc.ts
@@ -96,6 +96,39 @@ function tlcErr(explanation: string, isViolation: boolean): TlcError {
 }
 
 /**
+ * Validates a module name to prevent path traversal attacks.
+ * Module names should only contain alphanumeric characters and underscores.
+ *
+ * @param moduleName - The module name to validate
+ * @returns Either an error message or the validated module name
+ */
+function validateModuleName(moduleName: string): Either<string, string> {
+  // Only allow alphanumeric characters, underscores, and must start with letter/underscore
+  const safePattern = /^[a-zA-Z_][a-zA-Z0-9_]*$/
+  if (!safePattern.test(moduleName)) {
+    return left(
+      `Invalid module name: "${moduleName}". ` +
+        `Module names must contain only alphanumeric characters and underscores, ` +
+        `and must start with a letter or underscore.`
+    )
+  }
+  return right(moduleName)
+}
+
+/**
+ * Safely removes a temporary directory, ignoring errors if it doesn't exist.
+ *
+ * @param tmpDir - The directory path to remove
+ */
+function cleanupTmpDir(tmpDir: string): void {
+  try {
+    fs.rmSync(tmpDir, { recursive: true, force: true })
+  } catch {
+    // Ignore cleanup errors - directory may already be removed
+  }
+}
+
+/**
  * Validates a JVM memory argument (maxHeap or stackSize) against expected patterns.
  *
  * This function prevents JVM argument injection attacks by ensuring only valid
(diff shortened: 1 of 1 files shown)


erickpintor:
Hello. I spent some time reviewing this PR. Please keep in mind I’m not a security engineer. Someone with a security background might have a different opinion than I do.

Here’s my commit by commit review, conclusion at the bottom.

## Commit-by-commit review

### 1. Fix critical command injection vulnerability in Rust evaluator extraction

This commit seems legit. It’s feasible to think that by Quint querying the internet for artifacts, some malicious agent could hijack the request and respond with some embedded shell script.

### 2. Add JVM argument validation to prevent injection in TLC backend

The attack vector seems unreasonable since the user is the one responsible for providing the TLC config. Quint is not fetching the TLC config from an untrusted source.

### 3. Add ReDoS protection for –match regex pattern validation

Same as commit number 2. These are user-provided regular expressions. Quint is not fetching them from untrusted sources.

### 4. Add type-safe JSON parsing for external process communication

This patch is not okay, IMO. It makes JSON failures silent, which turn programming mistakes into logical errors in the application layer:

```javascript
function safeParseJson<T>(json: string, validator: (obj: unknown) => obj is T): T | undefined {
  try {
    const parsed: unknown = JSON.parse(json)
    return validator(parsed) ? parsed : undefined
  } catch {
    return undefined
  }
}
```

If we are trying to access a field that does not exist, we want the feedback in the form of a key not found error (and alike).

The bottom line here is that the patch treats Apalache and the Quint Evaluator as external services where, for all intents and purposes, they are not. These are in-process tools managed by Quint itself.

### 5. Add promise rejection handlers to all CLI command handlers

This doesn’t seem to be security related at a first look. If we don’t have a reason to believe that we have orphan promises, then this patch is a “best effort” kind of change, and it’s difficult to evaluate its value.

### 6. Replace non-null assertions with validated access patterns

Also seems not security related and seems similar to commit number 4. The bang operator (\!) signals an expectation: we trust that a value must exist at runtime at this point, otherwise it’s a programming mistake. Turning them into explicit errors increases complexity and doesn't add much user value, IMO, as they are not meant to break in the hands of users in the first place.

### 7. Fix type comparison bug in addedEntities effect checker

This seems like a legit bug. However, the lack of a unit test makes it hard to evaluate.

### 8. Improve temporary file security in TLC backend

Semi-legit, IMO. Checking module names and other parts of the spec for sensitive, security related issues is unnecessary. Quint does not load any specs from the internet and all the specs that it runs are user-provided.

The part about changing file permissions is nice to have. No one should be putting sensitive data in models but, it’d still be surprising if someone else in your intranet got access to your work in progress spec due to a shared temp folder.

## Conclusion

It’s quite difficult to approve a security related PR with so much in it. The commit descriptions are detailed but lack motivation: are these security bugs well known and documented? Or are any of them speculative?

For the commits that are security related, I’d expect a justification based on cross-referencing with documented vulnerabilities. The ones that don’t seem security related should be split into separate PRs so they don’t distract reviewers from the security focus.

Finally, the fact that no unit test has been changed as a result or added to exercise the new code makes the patch difficult to guarantee it’s working as intended.

The bottom line is: some commits seem legit to me. Others add code and complexity that makes it difficult to justify merging due to the lack of concrete evidence of their benefit (use-cases, unit/integration tests, etc).

erickpintor:
Hello again. I've opened 3 separate issues for the most impactful changes here:

* https://github.com/informalsystems/quint/issues/1933
* https://github.com/informalsystems/quint/issues/1934
* https://github.com/informalsystems/quint/issues/1935

At this point we decided to close this PR given the previous [feedback](https://github.com/informalsystems/quint/pull/1871#issuecomment-3993491197) provided and we'll prioritize fixing the issues accordingly. 

Of course, feel free to address them if you wish and thank you very much for taking the time bring these to our attention.

PR #1875, opened 2026-01-28
Always override console.{log,error} for LSP server

erickpintor:
Currently, the code only overrides stdout and stderr if logging is set. However, LSP clients that use std in and out for protocol communication with the LSP server break if non-conforming messages are printed out.

This patch makes it so the code always overrides both std out and err and let the global LOG_LEVEL constant define weather the message is logged or not.


bugarela:
TBH I'm surprised by this, since in Emacs, the `lsp-log` buffer has never shown server crashes to me (when I look at it). Not sure if I was looking at the wrong thing, or if maybe this is not happening in all instances of `--stdio` language server usages.

I'm fine with the changes tho.

PR #1876, opened 2026-01-28
Introduce async compilation pipeline to LSP server

erickpintor:
Quint compilation process is not incremental or async. Due to its single threaded model, the LSP server can freeze on some requests while compiling large modules.

This patch pushes compilation and code analysis to a separate worker thread so that the LSP server can continue to respond to LSP client requests while the code is being compiled and analyzed in the background.

The trade-off here is that on large modules we may potentially delay showing compilation errors or respond to request, like goto definition, based on a stale version of the module while the new one is being compiled.

### DIFF
--- a/vscode/quint-vscode/server/src/server.ts
+++ b/vscode/quint-vscode/server/src/server.ts
@@ -36,7 +36,7 @@ import {
 } from 'vscode-languageserver/node'
 import { DocumentUri, TextDocument } from 'vscode-languageserver-textdocument'
 import { URI } from 'vscode-uri'
-import { logger, overrideConsole } from './logger'
+import { isEnabled, logger, overrideConsole } from './logger'
 
 import {
   AnalysisOutput,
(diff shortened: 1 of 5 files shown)


vscode/quint-vscode/server/src/server.ts
```diff
@@ -473,52 +461,34 @@ export class QuintLanguageServer {
...
     this.analysisOutputByDocument.delete(uri)
+    this.pipeline.deleteDocument(uri)
   }
 
-  private triggerAnalysis(document: TextDocument) {
-    clearTimeout(this.analysisTimeout)
-    this.connection.console.info(`Triggering analysis`)
-    this.analyze(document)
-  }
+  private handleCompilationResult(result: CompilationResult): void {
```
bugarela: Not sure about the naming here. I feel like this should be `handleParsingResult`

bugarela (approved):
I'm impressed by how much manual work was required here. I'd expect LSP to have a better way of doing something that must be quite common. But I keep getting disappointed by LSP 😞.

I tried this locally on both Emacs+LSP and VSCode and it makes a huge difference in big specs. Thanks for taking care of this 🚀

PR #1879, opened 2026-01-30
Seed and trace on runtime error (rust only)

bugarela:
Hello :octocat: 

This start printing the trace (until the problem) and the seed for runtime erors, which should be of great help in debugging those.

Before:
image

After:
image

This also works for panics, but I can't think of a way to test it as we don't know any scenarios where the application panics from user input (and if we learn one, we will fix it). I tested locally by introducing an artificial panic.

It's probably easy to add this to the typescript backend too, but I didn't bother since we are deprecating it very soon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Moved LSP-server compilation and analysis to its own worker thread (1876)
 - Improved memory allocation profile of LSP-server log messages (#1877)
+- Stream JSON in and out or rust evaluator to reduce memory allocations (#1882)
 
 ### Deprecated
 ### Removed
--- a/evaluator/src/log.rs
+++ b/evaluator/src/log.rs
@@ -1,7 +1,7 @@
 //! Logging module for the simulation, used by the CLI.
 
-use std::fmt;
 use std::sync::atomic::AtomicBool;
+use std::{fmt, io};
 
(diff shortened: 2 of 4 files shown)


quint/io-cli-tests.md
```diff
@@ -1257,11 +1257,11 @@ exit $exit_code
 <!-- !test exit 1 -->
 <!-- !test out run uninitialized -->
 ```
+[error] Runtime error (duration).
 HOME/_1041compileConst.qnt:5:24 - error: [QNT500] Uninitialized const N. Use: import <moduleName>(N=<value>).*
 5:   action init = { x' = N }
                           ^
 
-Use --seed=0x1 --backend=typescript to reproduce.
```
bugarela, line 1264: This seed was not the actual seed of the trace, but the seed that was given in the args, so not useful at all.

This works fine in the rust backend.

evaluator/src/simulator.rs
```diff
@@ -80,50 +96,105 @@ impl ParsedQuint {
...
+                    }
+                }
+                Ok(true)
+            }));
 
-                if !invariant.execute(&mut env)?.as_bool() {
-                    trace_lengths.push(env.trace.len());
+            match result {
+                Ok(Ok(true)) => {
+                    // Successful run, continue
```
erickpintor: Would it be possible to take the rest of the loop at line 200 and move it here? Since every other branch in the match returns, it's be a bit simpler to read the data flow here.

PR #1884, opened 2026-02-02
Use custom deserializer for `LookupTable`

bugarela:
Hello :octocat: 

I spent some time trying to understand why, in the REPL PoC, we (me and Claude) had to introduce a custom deserializer for LookupTable. Some of the integration tests for `quint repl` break when I remove it, with some definitions not being found. Turns out that, depending on the complexity of the data type around LookupTable, serde is struggling to deserialize the string ids from TS into u64. So while it works fine in `quint run`, the specific datatype of the `ReplCommand` (with an enum) makes the automatic deserializer fail. A unit test represents the problematic scenario - the test would fail before this patch.


bugarela:
IDK what is happening with GitHub, but the tests are passing: https://github.com/informalsystems/quint/actions/runs/21604540356?pr=1884

evaluator/src/ir.rs
```diff
@@ -46,7 +47,81 @@ pub struct QuintOutput {
...
+    pub fn new() -> Self {
+        LookupTable(IndexMap::with_hasher(FxBuildHasher::default()))
+    }
+}
+
+impl Default for LookupTable {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
erickpintor: A few suggestions:

1. It looks like we only ever use the tables default implementation, so we don't need new + default;
2. We can keep deriving the Serialize trait since it's implementation just delegates to serde;
3. We should keep the IndexMap private so it's easier to drop this wrapper in the future once the bug is fixed (no uses of `table.0`).

```suggestion
#[derive(Default, Serialize, Debug, Clone)]
pub struct LookupTable(IndexMap<QuintId, LookupDefinition, FxBuildHasher>);
```

evaluator/src/ir.rs
```diff
@@ -46,7 +47,81 @@ pub struct QuintOutput {
...
+
+impl Serialize for LookupTable {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        self.0.serialize(serializer)
+    }
+}
+
```
erickpintor: Can be dropped in favor of `#[derive(.., Serialize, ..)]` above.

```suggestion
```

evaluator/src/ir.rs
```diff
@@ -46,7 +47,81 @@ pub struct QuintOutput {
...
+impl<'de> Deserialize<'de> for LookupTable {
```
erickpintor, line 71: Could you please link the bug and add a note so we eventually drop this once the bug is fixed upstream?

bugarela: Oh, thanks for finding that bug! Will do

PR #1885, opened 2026-02-02
Add `repl-from-stdin` command to Rust backend for REPL evaluation

bugarela:
Hello :octocat: 

This is the first PR towards supporting the Rust backend for the Quint REPL. It implements a new subcommand for the Rust evaluator which keeps a process opened, receiving commands from the STDIN and writing results to STDOUT, while keeping an evaluation state (compiled expressions, caching, `var` storage, traces, etc).

### DIFF
--- a/evaluator/tests/picker_tests.rs
+++ b/evaluator/tests/picker_tests.rs
@@ -99,7 +99,7 @@ fn powerset_large_set_pick_test() -> Result<(), Box<dyn std::error::Error>> {
     let parsed = helpers::parse(quint_content, "init", "step", None)?;
     let init_def = parsed.find_definition_by_name("init")?;
 
-    let mut interpreter = Interpreter::new(&parsed.table);
+    let mut interpreter = Interpreter::new(parsed.table.clone());
     let mut env = Env::with_rand_state(interpreter.var_storage.clone(), 0x42);
 
     let init = interpreter.eval(&mut env, init_def.expr.clone());
@@ -131,7 +131,7 @@ fn powerset_very_large_set_pick_test() -> Result<(), Box<dyn std::error::Error>>
     let parsed = helpers::parse(quint_content, "init", "step", None)?;
     let init_def = parsed.find_definition_by_name("init")?;
 
-    let mut interpreter = Interpreter::new(&parsed.table);
+    let mut interpreter = Interpreter::new(parsed.table.clone());
     let mut env = Env::with_rand_state(interpreter.var_storage.clone(), 0x42);
 
     let init = interpreter.eval(&mut env, init_def.expr.clone());


evaluator/src/repl.rs
```diff
@@ -0,0 +1,306 @@
...
+                missing_vars.push(reg.name.to_string());
+            }
+        }
+
+        // Check if anything actually changed
+        // If all nextVars are empty, nothing was set
+        let total_next_vars = storage.next_vars.len();
+        let any_changed = total_next_vars > 0 && missing_vars.len() < total_next_vars;
+
+        // Drop the borrow to avoid an old state getting referenced.
```
bugarela: @erickpintor I was confused by why this `drop` was needed here. If I remove it, the code compiles normally and I get Quint runtime errors on testing because the variables are not defined (QNT502 Variable is not set). 

I tried to understand it and wrote my findings in this comment. Can you check if it seems correct, please?

evaluator/src/evaluator.rs
```diff
@@ -159,10 +159,10 @@ pub struct Interpreter<'a> {
     // initialNondetPicks: Map<string, RuntimeValue | undefined> = new Map()
 }
 
-impl<'a> Interpreter<'a> {
-    pub fn new(table: &'a LookupTable) -> Self {
+impl Interpreter {
+    pub fn new(table: &LookupTable) -> Self {
         Self {
-            table,
+            table: Rc::new(table.clone()),
```
erickpintor: It seems to me that the lookup table is not cheap to clone line quint values are. It'd be best if we adjust the lifetimes so that the interpreter takes ownership of the lookup table instead.

```suggestion
    pub fn new(table: LookupTable) -> Self {
Self {
            table: Rc::new(table),
```

evaluator/src/repl.rs
```diff
@@ -0,0 +1,306 @@
...
+
+impl ReplEvaluator {
+    pub fn new() -> Self {
+        Self {
+            interpreter: None,
+            env: None,
+            trace_states: Vec::new(),
+            verbosity: 0,
+        }
+    }
```
erickpintor, line 89: Looks like the default constructor could be derived.

```suggestion
#[derive(Default)]
pub struct ReplEvaluator {
    interpreter: Option<Interpreter>,
    env: Option<Env>,
    trace_states: Vec<Value>,
    verbosity: u8,
}

impl ReplEvaluator {
```

evaluator/src/repl.rs
erickpintor: Are you getting those errors on the typescript side? My expectation is that if you don't drop this immutable reference to the storage you'll get a panic at line 213 once it attempts to get a mutable reference to storage. `RefCell` enforces the same data access rules as the type checker but, at runtime: you can have multiple immutable references to some value but never multiple mutable references to it.

evaluator/src/repl.rs
```diff
@@ -0,0 +1,306 @@
...
+
+    fn initialize(&mut self, table: LookupTable, seed: Option<u64>, verbosity: u8) -> ReplResponse {
+        self.verbosity = verbosity;
+
+        // Create the interpreter with the table
+        self.interpreter = Some(Interpreter::new(&table));
+
+        let storage = self.interpreter.as_ref().unwrap().var_storage.clone();
```
erickpintor: Minor: it'd be simpler to assign the newly instantiated interpreter to a local variable so you don't have to use `get_ref().unwrap()` here.

evaluator/src/repl.rs
```diff
@@ -0,0 +1,306 @@
...
+            return ReplResponse::ReplShiftResult {
+                shifted: false,
+                missing_vars: vec![],
+                old_state: None,
+                new_state: None,
+            };
+        }
+
+        // Save current state before shifting
+        let old_state = env.var_storage.borrow().as_record();
```
erickpintor: Minor: if you move the drop below this point, you could avoid borrowing storage again.

evaluator/src/repl.rs
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    fn get_trace_states(&self) -> ReplResponse {
+        let states: Vec<itf::Value> = self.trace_states.iter().map(|v| v.to_itf()).collect();
+        ReplResponse::TraceStates { states }
+    }
+
+    fn reset(&mut self) -> ReplResponse {
+        if let Some(env) = &self.env {
+            env.var_storage.borrow_mut().vars.clear();
```
erickpintor: Minor: you could assigned the mutably borrowed reference to var_storage to a local var to avoid borrowing twice here.

evaluator/src/repl.rs
bugarela: Yes, in the typescript side while using the rust backend. I read online that it can either panic or provide stale values, and I'm guessing we are getting a stale value here. The panic would have a different error message in the TS side.

erickpintor: That's interesting. Is that from an official source? [Documentation](https://doc.rust-lang.org/std/cell/struct.RefCell.html#panics-4) states that it simply panics. Could it be that it panics but the `catch_unwind` on L283 keeps the interpreter alive but without the updated table?

bugarela: You are correct! I was mishandling errors on `catch_unwind`. I have now fixed it to properly shutdown everything on panics, returning a `FatalError` from Rust which gets processes on TS properly to print the error and shutdown the REPL.

Here's one example with this particular panic (for a version without `drop`):

image

bugarela:
> I think `simulate` and `test` should consume `self` so you don't have to clone the lookup table as it moves inwards. I'm fine if you want to push and we can address that later.

I thought there might be a way of fixing that, but I didn't figure it out in 2 minutes and decided it was not worth 10 minutes of my type as it's just test env (and we need to revisit that anyway as it's super improvised at the moment).

I'll merge this to also avoid adding another commit to the crazy stack of rebasing I'm about to do.

bugarela:
Argh, updating with main broke the tests. One more commit to the stack it is

PR #1886, opened 2026-02-03
Add wrapper for Rust-backed REPL

bugarela:
Hello :octocat: 

This is the second step towards supporting the rust evaluator in the REPL.

We need a different kind of wrapper than the one for `quint run` and `quint test` since this needs to keep a process alive and send/receive multiple commands according to inputs in the REPL. We also need everything to be async in order to be able to interrupt (i.e. CTRL+C) the REPL at any point.


quint/src/rust/binaryManager.ts
bugarela, line 1: No new code here, this was extracted from the old `quintRustWrapper.ts`.

quint/src/rust/commandWrapper.ts
bugarela, line 1: Also no new code here, just moved the binary-related things away and renamed the wrapper to differentiate it from the REPL wrapper. Motivation for this name is that it takes care of one-shot commands like `quint run` and `quint test`, while the other wrapper takes care of the server required for the REPL.

quint/src/itf.ts
```diff
@@ -189,65 +189,7 @@ export function ofItf(itf: ItfTrace): QuintEx[] {
     return id
   }
 
-  const ofItfValue = (value: ItfValue): QuintEx => {
```
bugarela, line 192: This was just extracted to an exported function in order to be re-used in communication for the REPL, where we don't always get a full trace as result, but also standalone values that require this deserialization. I'm using ITF for everything.

quint/src/rust/replServerWrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  /**
+   * Update the lookup table with new definitions
+   */
+  updateTable(table: LookupTable): void {
+    // Fire and forget - async operation but sync interface
+    this.updateTableAsync(table).catch(err => {
+      console.error('Failed to update table:', err)
+    })
+  }
```
erickpintor: It feels like this fire and forget interface might cause the critical guard at line 157 but, I can't confirm until we see the code that uses this component.

erickpintor (approved):
LGTM

quint/src/rust/replServerWrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  async updateTableAsync(table: LookupTable): Promise<void> {
+    await this.initializationPromise
+    await this.sendCommand({ cmd: 'UpdateTable', table })
```
erickpintor, line 199: We should test this with large tables to understand the cost of large JSON objects flying around.

quint/src/rust/replServerWrapper.ts
```diff
@@ -0,0 +1,272 @@
...
+      console.error('Trying to send command:', command.cmd)
+    }
+
+    return new Promise<ReplResponse>((resolve, reject) => {
+      // Store the response handler
+      this.pendingResponse = resolve
+
+      // Write command to stdin
+      const commandJson = JSONbig.stringify(command, replacer)
+      this.process!.stdin!.write(commandJson + '\n', err => {
```
erickpintor, line 168: I forgot to ask: is this going to work if we have a literal string in a Quint expression with a new line in it? Are those escaped during json serialization?

quint/src/rust/replServerWrapper.ts
bugarela: I ended up removing this a few minutes ago when working on my next PR. I thought I had removed this already, my bad. We don't actually use it, we are only using the async interfaces and I unified them properly upon realizing that TS doesn't mind having the same method be async in a class (this one) and not async in other (`Evaluator`) while they are user interchangeably (`Evaluator | ReplServerWrapper`).

quint/src/rust/replServerWrapper.ts
bugarela: We don't seem to support parsing line breaks (if you press return as you are typing a string literal). Parsing "aa\n" works fine. Also, parsing multi-line things like 
```
>>> def foo(x) = {
... 1 + x
... }
...

>>>
```

works fine.

quint/src/rust/replServerWrapper.ts
bugarela: I tried using the REPL with the monad spec which is big and there's a very small difference between Rust and TS. We should improve this, but it doesn't seem like a blocker to me:
```
$ time quint -r instance.qnt::instance --verbosity 1 "init" "step" "step" ".exit"
Quint REPL 0.30.0
>>> init
true
>>> step
true
>>> step
true
>>> .exit

________________________________________________________
Executed in   10.99 secs    fish           external
   usr time   13.12 secs  669.00 micros   13.12 secs
   sys time    0.20 secs   54.00 micros    0.20 secs

$ time quint -r instance.qnt::instance --verbosity 1 --backend rust "init" "step" "step" ".exit"
Quint REPL 0.30.0 (using the Rust backend)
>>> init
true
>>> step
true
>>> step
true
>>> .exit

________________________________________________________
Executed in   11.63 secs    fish           external
   usr time   13.79 secs  770.00 micros   13.79 secs
   sys time    0.27 secs  972.00 micros    0.27 secs
```

(I tried it a few times and this small difference is consistent. I also feel like it takes a split second longer in Rust when typing the expressions manually)

PR #1887, opened 2026-02-03
Add support for integer overflow detection in Rust backend

beu5a:
Hello,

This is the first step towards enabling integer overflow detection in the rust backend. 

We refactored arithmetic operations and set cardinality operations to use the checked versions of rust arithmetic ops. We added support for propagating these errors to the user via a new Quint error code and we added panics when such reporting is not possible. 


evaluator/src/builtins.rs
```diff
@@ -159,28 +159,18 @@ pub fn compile_lazy_op(op: &str) -> CompiledExprWithLazyArgs {
         "oneOf" => |env, args| {
             // Randomly selects one element of the set.
             let set = args[0].execute(env)?;
-            // Some sets require multiple random numbers in order to pick an element efficiently.
```
bugarela, line 162: Can we bring back the comments?

evaluator/flake.nix
```diff
@@ -1,22 +1,28 @@
...
+              pkgs.clang
+              pkgs.libiconv
+            ]
+            ++ pkgs.lib.optionals pkgs.stdenv.isDarwin [
+              pkgs.darwin.apple_sdk.frameworks.CoreFoundation
+            ];
         };
       });
- }
+}
```
bugarela: Re-adding the new line at end of file which is standard
```suggestion
}

```

evaluator/src/itf.rs
```diff
@@ -88,7 +88,10 @@ impl Value {
             | ValueInner::CrossProduct(_)
             | ValueInner::PowerSet(_)
             | ValueInner::MapSet(_, _) => {
-                itf::Value::Set(self.as_set().iter().map(|v| v.to_itf()).collect())
+                let set = self
+                    .as_set()
+                    .expect("failure in ITF conversion: set could not be enumerated");
```
erickpintor: Minor: we usually put reason for the failure at it's source, in the `as_set()` method, and just add location context in the places we panic.

```suggestion
                    .expect("can convert value to set for ITF conversion");
```

evaluator/src/normalizer.rs
```diff
@@ -2,45 +2,52 @@
...
-                    .into_iter()
-                    .map(|v| v.normalize())
-                    .collect(),
-            ),
+            | ValueInner::MapSet(_, _) => {
+                let set = self.as_set()?;
+                let normalized: Result<Vec<_>, _> =
+                    set.iter().map(|v| v.clone().normalize()).collect();
+                Value::set(normalized?.into_iter().collect())
+            }
```
erickpintor: This is looping twice over the set: one for collecting values, another for building the immutable set. This should move back to a single iteration:

```suggestion
            | ValueInner::MapSet(_, _) => Value::set(
                self.as_set()?
                    .into_owned()
                    .into_iter()
                    .map(|v| v.normalize())
                    .collect::<Result<_, _>>()?,
            ),
```

erickpintor: The same strategy should apply for the other changes in this file.

evaluator/src/picker.rs
```diff
@@ -26,59 +27,72 @@ impl Value {
                 Value::int(start + idx)
             }
             ValueInner::CrossProduct(sets) => {
-                Value::tuple(sets.iter().map(|value| value.pick(indexes)).collect())
+                let elements: Result<Vec<_>, _> =
+                    sets.iter().map(|value| value.pick(indexes)).collect();
+                Value::tuple(elements?.into_iter().collect())
```
erickpintor: This is also looping twice. We should move it back to a single iteration.

evaluator/src/picker.rs
```diff
@@ -26,59 +27,72 @@ impl Value {
...
+                let key_values: Result<Vec<_>, _> = keys
+                    .iter()
+                    .map(|key| {
+                        let value = range_to_pick.pick(indexes)?;
+                        Ok((key.clone(), value))
+                    })
+                    .collect();
 
-                Value::map(ImmutableMap::from_iter(key_values))
+                Value::map(ImmutableMap::from_iter(key_values?))
```
erickpintor: Another double loop here.

evaluator/src/rand.rs
```diff
@@ -25,7 +28,7 @@ impl Default for Rand {
 impl Rand {
     pub fn new() -> Self {
         Self {
-            counter: rand::rng().random(),
+            counter: rand::thread_rng().gen(),
```
erickpintor, line 31: Don't we get a small chance that two threads will start from the same counter if their random source is not coordinated?

evaluator/src/rand.rs
```diff
@@ -48,4 +51,25 @@ impl Rand {
...
+        if let Some(bound_u64) = bound.to_u64() {
+            if bound_u64 <= usize::MAX as u64 {
+                return BigUint::from(self.next(bound_u64 as usize));
+            }
+        }
+
+        // For large bounds, create a seeded RNG from current state
+        // This maintains determinism while leveraging num-bigint's random generation
+        let mut rng = rand::rngs::StdRng::seed_from_u64(self.counter);
+        let result = rng.gen_biguint_range(&BigUint::from(0u64), bound);
```
erickpintor: minor:
```suggestion
        let result = rng.gen_biguint_range(&BigUint::ZERO, bound);
```

evaluator/src/rand.rs
```diff
@@ -48,4 +51,25 @@ impl Rand {
     pub fn get_state(&self) -> u64 {
         self.counter
     }
+
+    /// Generate a random BigUint in the range [0, bound).
+    /// Maintains determinism by using the current counter state to seed the RNG.
+    pub fn next_biguint(&mut self, bound: &BigUint) -> BigUint {
+        // For small bounds that fit in usize, use existing deterministic path
+        if let Some(bound_u64) = bound.to_u64() {
```
erickpintor, line 59: minor: Rust now supports joining if-lets with `&&` operators:

```rust
if let Some(var) = .. && var == ... {
```

evaluator/src/value.rs
```diff
@@ -173,7 +173,13 @@ impl PartialEq for ValueInner {
                 let self_value = Value(Rc::new(self.clone()));
                 let other_value = Value(Rc::new(other.clone()));
                 if self_value.is_set() && other_value.is_set() {
-                    self_value.as_set() == other_value.as_set()
+                    let self_set = self_value
+                        .as_set()
+                        .expect("failure in equality check: left set could not be enumerated");
```
erickpintor: Same thing here about location versus error cause.

evaluator/src/value.rs
```diff
@@ -252,80 +258,156 @@ impl Value {
...
+            ValueInner::List(elems) => Ok(elems.len()),
+            ValueInner::Interval(start, end) => {
+                // Check for overflow when computing interval size
+                end.checked_sub(*start)
+                    .and_then(|diff| diff.checked_add(1))
+                    .and_then(|size| size.try_into().ok())
+                    .ok_or_else(|| {
+                        QuintError::new(
+                            "QNT601",
+                            "Integer overflow in cardinality computation: interval cardinality exceeds usize::MAX",
```
erickpintor: `usize::MAX` seems too much "rust" for quint users to have to look at. Perhaps say "maximum unsigned int", or just "maximum integer" and we document what maximum mean somewhere?

erickpintor: Same thing for other errors in this file.

evaluator/src/value.rs
erickpintor: There are other occurrences on this file too.

evaluator/src/rand.rs
beu5a: We do, but this applies for the old implementation too. The current change is just reverting back to rand 0.8.5. I think that a fix for this should be done in a separate PR.

evaluator/src/rand.rs
beu5a: cargo complained about this: 
` let expressions in this position are unstable`

erickpintor: My bad. I thought the evaluator was on the 2024 edition already.

PR #1889, opened 2026-02-05
Add `--witnesses` flag support in the Rust backend

beu5a:
Hey, this is a quick PR to add support for the witnesses statistics in the Rust backend. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add error handling for rust process termination via signal (#1880)
 - The rust backend will now print the seed and trace even on runtime errors and panics (#1879)
 - Added support for `quint repl --backend=rust` to use the REPL with the Rust backend (#1891)
+- Added out of bounds integer literal detection when serializing input to the Rust backend (#1890)
 - Added support for the `--witnesses` flag in the Rust backend (#1889)
 
 ### Changed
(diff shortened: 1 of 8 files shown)


evaluator/src/main.rs
```diff
@@ -339,9 +338,24 @@ fn simulate_in_parallel(
...
     let mut samples = 0;
+    let mut aggregated_witnesses = vec![];
+
     loop {
         let mut outcome = out_rx.recv().expect("closed channel");
         samples += outcome.samples;
+
+        // Accumulate witness counts from all threads
+        if aggregated_witnesses.is_empty() {
+            aggregated_witnesses = outcome.witnessing_traces.clone();
```
erickpintor: You can take it instead of cloning it here since upon return you move the aggregated witness back in place.

```suggestion
            aggregated_witnesses = std::mem::take(&mut outcome.witnessing_traces);

```

evaluator/src/simulator.rs
```diff
@@ -135,11 +169,19 @@ impl ParsedQuint {
...
                     }
 
                     trace_lengths.push(env.trace.len());
 
+                    // Finalize witness counts for this trace
+                    for (i, &witnessed) in trace_witnessed.iter().enumerate() {
+                        if witnessed {
+                            witnessing_traces[i] += 1;
+                        }
+                    }
```
erickpintor: Why not inline the counts with the loop that checks if the witness is hit above?

evaluator/src/simulator.rs
```diff
@@ -106,6 +117,29 @@ impl ParsedQuint {
                     // Shift the state and record it in the trace
                     env.shift();
 
+                    // Evaluate witnesses after each step and skip if all satisfied
+                    if remaining > 0 {
+                        for (witnessed_flag, witness) in
+                            trace_witnessed.iter_mut().zip(compiled_witnesses.iter())
```
erickpintor: This is probably a rare occasion where a `for i in 0..len` would be better since the vector size is fixed and it won't have to allocate the iterator structures.

PR #1892, opened 2026-02-09
Introduce diagnostics to Rust evaluator

erickpintor:
Currently, `q::debug` is not very useful in the rust evaluator. It's multi-threaded design makes it so many messages appear in the console and it's really difficult to track from which traces they are coming from.

This patch is the first part of a redesign of how `q::debug` works. In it, we're tracking debug messages as per state diagnostics in the traces. The best traces are selected, preserving diagnostics when applicable, so at the CLI interface we'll report only diagnostics associated with the reported traces at the end.

### DIFF
--- a/evaluator/src/simulator.rs
+++ b/evaluator/src/simulator.rs
@@ -5,10 +5,11 @@ use crate::{
     ir::{LookupTable, QuintError, QuintEx},
     itf::Trace,
     progress::Reporter,
+    verbosity::Verbosity,
 };
 use serde::{Deserialize, Serialize};
-use std::fmt;
 use std::panic::{catch_unwind, AssertUnwindSafe};
+use std::{cmp::Ordering, fmt};
 
 /// Simulation input that depends on the typescript Quint tool.
 #[derive(Serialize, Deserialize, Clone)]
@@ -77,11 +78,12 @@ impl ParsedQuint {
         n_traces: usize,
         mut reporter: R,
         seed: Option<u64>,
+        verbosity: Verbosity,
     ) -> Result<SimulationResult, SimulationError> {
         let mut interpreter = Interpreter::new(self.table.clone());
         let mut env = match seed {
-            Some(s) => Env::with_rand_state(interpreter.var_storage.clone(), s),
-            None => Env::new(interpreter.var_storage.clone()),
+            Some(s) => Env::with_rand_state(interpreter.var_storage.clone(), s, verbosity),
+            None => Env::new(interpreter.var_storage.clone(), verbosity),
         };
 
         let init = interpreter.compile(&self.init);
@@ -182,6 +184,7 @@ impl ParsedQuint {
                             violation: !success,
                             seed,
                         },
+                        verbosity,
                     );
 
                     if !success {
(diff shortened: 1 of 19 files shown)


evaluator/src/simulator.rs
```diff
@@ -228,31 +231,63 @@ fn get_trace_statistics(trace_lengths: &[usize]) -> TraceStatistics {
...
 ///  - when a has no error: a is longer and b has no error.
-fn compare_traces_by_quality(a: &Trace, b: &Trace) -> std::cmp::Ordering {
-    match (a.violation, b.violation) {
-        (true, true) => a.states.len().cmp(&b.states.len()),
-        (true, false) => std::cmp::Ordering::Less,
-        (false, true) => std::cmp::Ordering::Greater,
-        (false, false) => b.states.len().cmp(&a.states.len()),
+///
+/// For traces considered of equal value, both errors or both non-errors, we
+/// consider the one with diagnostics to be better.
```
bugarela, line 291: I'm not convinced this makes sense. It seems to me like we shouldn't prioritize a trace for covering debug statements. For instance, when using the simulator for MBT, I'd like a more random coverage then some preference for traces that hit my debug statements (that I might have even forgotten are there).

Also, from the code, it seems like we are giving more preference to the existence of diagnostics than to the length of the trace, which is not clear from this comment (which made me assume otherwise before looking at the code)

erickpintor: Diagnostics are opt-in and no-op if disabled. In the case of MBT, they won't interfere with trace selection unless the user specifically enables higher verbosity -- in practice they can't if using Quint Connect today.

The preference here, at the moment, is (in order):

1. A trace with a violation wins;
2. A trace with diagnostics wins, if enabled;
3. The shortest trace wins, if both are violations;
4. The longest trace wins, if no violations.

I'm okay refining this selection. My only concern is that if a user has opted into diagnostics via `verbosity=3`, they'd probably want to see their diagnostics rather than getting longer or shorter traces, IMO.

bugarela: I'm not sure, maybe let's get input from the team on slack or tomorrow.
I also feel unsure about this being disabled in default verbosity. This will big a big change from current behavior, as people (few more advanced users) are used using `q::debug` with default verbosity. We had gotten requests to hide `q::debug` in lower verbosities, but still the default one could have it. I feel like we would benefit more from showing it by default than not.

And, as a user, I'd expect `q::debug` statements to have zero impact on the kind of traces I get. I think it should be a separate feature to "find" a trace that hits my debug (as also came up in conversation with users before).

beu5a: - I feel that it makes sense to prioritize the traces with diagnostics. My workflow with Quint when writing specs was always like this: 
 `write a spec -> hit a problem -> add debugs to try to understand what happens  -> fix specs -> remove debugs -> look for a better trace`
The new order that @erickpintor introduced fits really well with this workflow imo. 

- I think that the `DEBUG` statements should be part of higher than default verbosity, this is similar to most software where the info level is the default and the debug has to be explicitly requested. For the current users, they will expect to have some changes if they move to rust and it is easy to figure out this new behavior.

bugarela: Ok! We can go with this and iterate if needed.

PR #1894, opened 2026-02-09
Support for infinite sets in rust backend

bugarela:
Hello :octocat: 

This adds support for `Nat` and `Int` infinite sets to rust. For now, I'm going with a similar behavior from TS: refuse to enumerate (of course) but allow `oneOf()` by picking a number from `i64`. In TS we used to pick from `[-2^255, 2^255)`, but we already decided to not support bigints in Rust.

This also fixes an issue where `QuintEror`s from the REPL were not being deserialized properly. There was already an workaround for this in `commandWrapper`, but I'm now making it more streamlined so it's a bit harder to forget to do it if we duplicate deserialization somewhere else in the future, by using a `reviver` tranformer.

### DIFF
--- a/evaluator/tests/picker_tests.rs
+++ b/evaluator/tests/picker_tests.rs
@@ -166,7 +166,8 @@ fn int_pick_can_be_negative() -> Result<(), Box<dyn std::error::Error>> {
     let mut found_positive = false;
 
     for seed in 0..100 {
-        let mut env = Env::with_rand_state(interpreter.var_storage.clone(), seed);
+        let mut env =
+            Env::with_rand_state(interpreter.var_storage.clone(), seed, Verbosity::default());
         let init = interpreter.eval(&mut env, init_def.expr.clone());
         assert!(init.is_ok());
 
@@ -218,7 +219,8 @@ fn nat_pick_is_non_negative() -> Result<(), Box<dyn std::error::Error>> {
 
     // Try multiple seeds to ensure we never get a negative number
     for seed in 0..100 {
-        let mut env = Env::with_rand_state(interpreter.var_storage.clone(), seed);
+        let mut env =
+            Env::with_rand_state(interpreter.var_storage.clone(), seed, Verbosity::default());
         let init = interpreter.eval(&mut env, init_def.expr.clone());
         assert!(init.is_ok());
 


evaluator/src/picker.rs
```diff
@@ -83,6 +90,34 @@ impl Value {
 
                 Value::map(key_values)
             }
+            ValueInner::InfiniteInt => {
+                // Pick a random integer from the entire i64 range
+                let index = indexes
+                    .next()
+                    .expect("Internal error: too few positions. Report a bug");
+
+                // Map [0, usize::MAX) uniformly to [i64::MIN, i64::MAX]
```
bugarela: This was made by Claude, by I read about the datatypes in the docs and it seems to be correct to me.

evaluator/tests/evaluation_tests.rs
```diff
@@ -1230,6 +1230,61 @@ fn run_q_debug_single_arg() -> Result<(), Box<dyn std::error::Error>> {
     )
 }
 
+#[test]
```
bugarela, line 1234: These test cases were copied from `compile.test.ts` like everything else in this file.

evaluator/src/value.rs
```diff
@@ -169,6 +176,14 @@ impl PartialEq for ValueInner {
             (Self::CrossProduct(a), Self::CrossProduct(b)) => *a == *b,
             (Self::PowerSet(a), Self::PowerSet(b)) => *a == *b,
             (Self::MapSet(a1, b1), Self::MapSet(a2, b2)) => a1 == a2 && b1 == b2,
+            (Self::InfiniteInt, Self::InfiniteInt) => true,
+            (Self::InfiniteNat, Self::InfiniteNat) => true,
+            (Self::InfiniteInt, Self::InfiniteNat) | (Self::InfiniteNat, Self::InfiniteInt) => {
+                false
+            }
```
erickpintor: This case seems redundant with the two cases below.

evaluator/src/value.rs
```diff
@@ -413,6 +453,22 @@ impl Value {
                 ValueInner::MapSet(subset_domain, subset_range),
                 ValueInner::MapSet(superset_domain, superset_range),
             ) => subset_domain == superset_domain && subset_range.subseteq(superset_range)?,
+            // Infinite set relationships
+            (ValueInner::InfiniteNat, ValueInner::InfiniteNat) => true,
+            (ValueInner::InfiniteNat, ValueInner::InfiniteInt) => true,
+            (ValueInner::InfiniteInt, ValueInner::InfiniteInt) => true,
+            (ValueInner::InfiniteInt, ValueInner::InfiniteNat) => false,
+            // Any non-infinite set is a subset of Int
+            (_, ValueInner::InfiniteInt) if self.is_set() => true,
```
erickpintor: This seems wrong. Does it mean that a set of `str` is a subset of ints?

evaluator/src/builtins.rs
```diff
@@ -299,7 +299,28 @@ pub fn compile_eager_op(op: &str) -> CompiledExprWithArgs {
...
+            // These cannot be enumerated
+            for arg in &args {
+                match arg.0.as_ref() {
+                    ValueInner::InfiniteInt => {
+                        return Err(QuintError::new(
+                            "QNT501",
+                            "Infinite set Int is non-enumerable",
+                        ))
+                    }
+                    ValueInner::InfiniteNat => {
```
erickpintor, line 321: Does this mean that Set(Nat) is invalid? I think it makes sense. Just checking.

evaluator/src/itf.rs
```diff
@@ -87,7 +87,9 @@ impl Value {
             | ValueInner::Interval(_, _)
             | ValueInner::CrossProduct(_)
             | ValueInner::PowerSet(_)
-            | ValueInner::MapSet(_, _) => {
+            | ValueInner::MapSet(_, _)
+            | ValueInner::InfiniteInt
+            | ValueInner::InfiniteNat => {
```
erickpintor: This will fail to turn infinite sets into ITF because they are not enumerable. Is that the intended behavior?

evaluator/src/picker.rs
erickpintor: Branching on the CPU architecture seems unnecessary. We could keep just the first implementation and simplify things.

evaluator/src/value.rs
bugarela: That would be caught by the typechecker. It's a type error, not a `false` result.
I could still change it here to be more strict, but it could affect performance if we do it the safest way and look into the values of the set (thus enumerating it).

evaluator/src/builtins.rs
bugarela: It does. This is existing behavior (there's a unit test for this that expects an error), which I guess makes sense.

evaluator/src/itf.rs
bugarela: I was going to say that ITF doesn't support it, but went to check first and it [actually does](https://apalache-mc.org/docs/adr/015adr-trace.html):
```
11. An expression that cannot be serialized: { "#unserializable": "<string representation>" }. For instance, the set of all integers is represented with { "#unserializable": "Int" }. This should be a very rare expression, which should not occur in normal traces. Usually, it indicates some form of an error.
```

I'll add it!

evaluator/src/picker.rs
```diff
@@ -116,10 +144,22 @@ impl Value {
...
+            }
+            ValueInner::InfiniteInt => {
+                // For Int, we use the maximum usize value as the bound
+                // This allows picking from the full i64 range via wrapping conversion
+                vec![usize::MAX]
+            }
+            ValueInner::InfiniteNat => {
+                // For Nat, we use i64::MAX as the bound to ensure only non-negative values
+                // when converted to i64
+                vec![i64::MAX as usize]
```
beu5a: why are we casting from `i64` to `usize` here?

evaluator/src/value.rs
erickpintor: I'd rather panic if something that the type checker was supposed to catch ended up here. It'd be less confusing, IMO.

evaluator/src/picker.rs
erickpintor: perhaps we should change `bounds()` to return a vector of `u64` so we don't return different bounds depending on the architecture.

bugarela: Done https://github.com/informalsystems/quint/pull/1894/commits/f7bcd81b3c6f8a7a308a9f898be126904ae8edb8

This was a big change, but I think it turned out good.

evaluator/src/value.rs
bugarela: Sorry I missed this. It's now fixed.

PR #1895, opened 2026-02-09
Update xml files

bugarela:
Hello :octocat: 

The lessons in our website have correspondent CodeTour interactive lessons, and both of these are generated from an XML file (and an internal project tool produces both `.md` and `.tour` files from it). However, we all often forget about this (specially in PR review) and update the `.md` directly instead of updating `.xml` and re-generating it.

We have similar situations in other parts of code, but those are checked in CI. This one is not.

I'm not adding the CI check here, just fixing the existing mismatches for now


erickpintor (approved):
🐑 it.

PR #1896, opened 2026-02-10
Add `--mbt` flag support to the rust backend

beu5a:
Hello,
This adds support for the `--mbt` flag in the rust backend and duplicates the same mbt logging behavior as the TS frontend. I think that we should discuss if we want to keep the current behavior of `--mbt` or redesign it, but I think that comes after deprecating the TS backend and we should now just duplicate the same behavior on the rust side. 

### DIFF
--- a/evaluator/src/builtins.rs
+++ b/evaluator/src/builtins.rs
@@ -424,9 +424,7 @@ pub fn compile_eager_op(op: &str) -> CompiledExprWithArgs {
                 .ok_or_else(|| {
                     QuintError::new(
                         "QNT601",
-                        &format!(
-                            "Integer overflow in arithmetic operations: {base} ^ {exp}"
-                        ),
+                        &format!("Integer overflow in arithmetic operations: {base} ^ {exp}"),
                     )
                 })
                 .map(Value::int)


quint/src/cliCommands.ts
```diff
@@ -408,11 +408,6 @@ export async function runSimulator(prev: TypecheckedStage): Promise<CLIProcedure
 
   let outcome: Outcome
   if (prev.args.backend == 'rust') {
-    if (prev.args.mbt) {
```
bugarela, line 411: :tada:

bugarela (approved):
LGTM! Thanks for cleaning up and adapting the tests too

evaluator/src/storage.rs
```diff
@@ -29,27 +38,61 @@ pub struct VariableRegister {
...
+    // The action taken in the current step
+    pub action_taken: Option<QuintName>,
+    // Whether to store metadata for mbt
+    pub store_metadata: bool,
+}
+
+impl Default for Storage {
+    fn default() -> Self {
+        Self::new(false)
+    }
```
erickpintor: Minor: false is the default of a boolean so, I think you can simply derive default instead.

PR #1898, opened 2026-02-10
Support for `--n-traces` in rust backend

bugarela:
Hello :octocat: 

`--n-traces` is used for producing multiple ITF traces in a single run. We store the N best (in our quality definition) traces and return them in the end. We had a few options on how to make this work with multiple threads, and I chose to prioritize memory consumption over quality of traces which is what I'd have preferred in the cases where I used this feature before. It should be easy to change this later if needed.

So if the user asks for 10 traces and we have 2 threads, each thread will store the 5 best traces it finds, and then we'll combine those and sort them by quality. I'm a bit scared of this last sort, maybe we can remove it in favor of some statistical approach (i.e. taking all the first traces of all threads, then all seconds etc)


evaluator/tests/simulator_tests.rs
```diff
@@ -88,3 +88,99 @@ fn one_of_empty_set_ok() {
     // `step` should never succeed, so we should have a trace of length 1 (the initial state)
     assert!(result.unwrap().trace_statistics.max_trace_length == 1);
 }
+
+#[test]
```
bugarela, line 132: I think I really like these property-based style testing that Claude has started generating.

evaluator/src/main.rs
```diff
@@ -324,6 +326,7 @@ fn simulate_in_parallel(
 ) -> SimOutput {
     assert!(nthreads > 1, "nthreads must be > 1");
     nthreads = nthreads.min(nruns); //avoid spawning threads with no work
+    let per_thread_ntraces = ntraces.div_ceil(nthreads);
```
erickpintor: `div_ceil` will ended collecting more traces than requested. Perhaps it's worth doing an approach similar to `nruns` below where the first thread picks up the remainder.

PR #1902, opened 2026-02-12
Add support for multi-invariant reporting in Rust

beu5a:
Hey,
This is a PR to add support for reporting on which invariant broke to the Rust Backend. I had to change some interfaces in the TS backend (notably simulation `Outcome`) to be able to have a clean implementation in Rust. 

### DIFF
--- a/quint/src/cliReporting.ts
+++ b/quint/src/cliReporting.ts
@@ -203,7 +203,9 @@ export async function processApalacheResult(
       console.log(chalk.red(`[${status}]`) + ' Found an issue ' + chalk.gray(`(${elapsedMs}ms).`))
       const itfStates = err.traces?.[0]?.states
       const lastItfState = itfStates?.[itfStates.length - 1]
-      await printViolatedInvariantsWithRust(invariantsList, verbosityLevel, invariantExprs, table, lastItfState)
+      if (invariantExprs && table && lastItfState) {
+        await printViolatedInvariantsWithRust(invariantsList, verbosityLevel, invariantExprs, table, lastItfState)
+      }
     }
 
     if (stage.args.outItf && err.traces) {
@@ -222,17 +224,15 @@ export async function processApalacheResult(
 async function printViolatedInvariantsWithRust(
   invariantsList: string[],
   verbosityLevel: number,
-  invariantExprs?: QuintEx[],
-  table?: LookupTable,
-  itfState?: ItfState
+  invariantExprs: QuintEx[],
+  table: LookupTable,
+  itfState: ItfState
 ): Promise<void> {
-  if (invariantExprs && table && itfState) {
-    const result = await findViolatedInvariants(itfState, table, invariantExprs, verbosityLevel)
-    if (result.isRight()) {
-      printViolatedInvariantsByIndex(result.value, invariantsList)
-    } else {
-      console.warn(chalk.yellow(`Warning: could not determine violated invariants: ${result.value.message}`))
-    }
+  const result = await findViolatedInvariants(itfState, table, invariantExprs, verbosityLevel)
+  if (result.isRight()) {
+    printViolatedInvariantsByIndex(result.value, invariantsList)
+  } else {
+    console.warn(chalk.yellow(`Warning: could not determine violated invariants: ${result.value.message}`))
   }
 }
 
--- a/quint/src/rust/commandWrapper.ts
+++ b/quint/src/rust/commandWrapper.ts
@@ -19,7 +19,7 @@ import { debugLog } from '../verbosity'
 import JSONbig from 'json-bigint'
 import { LookupDefinition, LookupTable } from '../names/base'
 import { reviver } from '../jsonHelper'
-import { ItfState, diagnosticsOfItf, ofItf } from '../itf'
+import { ItfState, ItfValue, diagnosticsOfItf, ofItf } from '../itf'
 import { Presets, SingleBar } from 'cli-progress'
 import readline from 'readline'
 import { spawn } from 'child_process'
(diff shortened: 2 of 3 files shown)


beu5a:
> I see you changed the approach significantly to what I suggested in the card and that I failed to provide some important context there.
> 
> Although this is simpler that what we currently have, this one will not work for Apalache and TLC traces. The one we have takes the trace, regardless of source (TS backend, Rust backend, Apalache, TLC) and evaluates the invariants on the last state of that trace. Apalache and TLC are not capable of this.

I see, I missed this point about tlc and apalache and this approach seemed more suitable for the simulator. I will restart this.

beu5a:
> I see you changed the approach significantly to what I suggested in the card and that I failed to provide some important context there.
> 
> Although this is simpler that what we currently have, this one will not work for Apalache and TLC traces. The one we have takes the trace, regardless of source (TS backend, Rust backend, Apalache, TLC) and evaluates the invariants on the last state of that trace. Apalache and TLC are not capable of this.

Actually, I think I can build on this change and expose one more function in Rust that has the same interface as the current rust simulation function (ie what you called `evaluate_at_state` in the card) that is used only when we have broken invariants with TLC/Apalache that way we don't have to spin a new rust process for the last state evaluation when we use the simulator. wdyt ?

bugarela:
> Actually, I think I can build on this change and expose one more function in Rust that has the same interface as the current rust simulation function (ie what you called `evaluate_at_state` in the card) that is used only when we have broken invariants with TLC/Apalache that way we don't have to spin a new rust process for the last state evaluation when we use the simulator. wdyt ?

Sounds good! I liked how simple this turned out here, so using it for our main usage (simulator) is probably a good idea.

beu5a:
I decided to not return an error in case the invariants evaluation fails on the rust side but rather print a warning that reports the error. I felt that a full error is an overkill if Apalache runs successfully but the rust backend fails to print the violated invariant. The warning will contain the info about the source of failure: 
image

evaluator/src/itf.rs
```diff
@@ -118,6 +118,61 @@ impl Trace {
 }
 
 impl Value {
+    /// Convert an ITF value directly to a Quint [`Value`].
+    pub fn from_itf(itf_value: itf::Value) -> Self {
```
erickpintor, line 122: There's an implementation block for `Value` inside `itf.rs`. It looks like this function should be there.

erickpintor (commented):
Q: would it make sense to return violations ready to be printed instead of calling the rust backend again to get the violation results for reporting?

beu5a:
> Q: would it make sense to return violations ready to be printed instead of calling the rust backend again to get the violation results for reporting?

@erickpintor I didn't get your question. Can you maybe clarify/reformulate ?

erickpintor:
Nevermind, I think I misunderstood the purpose of `evaluate_at_state`. Is that only for reporting on invariants coming from the MC (not the evaluator)?

beu5a:
Yes currently it is only for tlc/apalache violations.

evaluator/src/itf.rs
beu5a: The `from_itf` function is already inside the `Value` implementation block of `itf.rs`. Am I missing something here?

quint/src/cliReporting.ts
```diff
@@ -144,52 +164,76 @@ export function processTlcResult(
...
+
+/**
+ * Print violated invariants using the Rust evaluator.
+ */
+async function printViolatedInvariantsWithRust(
+  invariantsList: string[],
+  verbosityLevel: number,
+  invariantExprs?: QuintEx[],
+  table?: LookupTable,
+  itfState?: ItfState
```
bugarela: Seems like `table` and `itfState` should not be optional here. We should return some sort of error if they are not present.

quint/src/rust/commandWrapper.ts
```diff
@@ -235,15 +232,42 @@ export class CommandWrapper {
...
+   * @param {LookupTable} table - The lookup table for name resolution.
+   * @param {QuintEx[]} exprs - The expressions to evaluate at the given state.
+   *
+   * @returns The evaluation results for each expression.
+   */
+  async evaluateAtState(
+    state: ItfState,
+    table: LookupTable,
+    exprs: QuintEx[]
+  ): Promise<Either<QuintError, { results: { value?: any; error?: QuintError }[] }>> {
```
bugarela: I don't like returning this weird type here and then having to handle it at `findViolatedInvariants` later on. If there is any error here, we should return `left(error)`, and the right side of `Either` should be a list of Values or ItfValues.

quint/verification-tests.md
```diff
@@ -482,9 +482,10 @@ error: tpos is used before it is assigned. You need to have either `tpos == <exp
 
 ### Shows which property broke when checking inductive invariants
 
-Here we pass only `TypeOK` as the inductive invariant, which is not strong enough to imply `TerminationDetection`. 
+Here we pass only `TypeOK` as the inductive invariant, which is not strong enough to imply `TerminationDetection`.
 
-<!-- !test in weak inductive invariant -->
+<!-- TODO: Re-enable once the Rust evaluator release includes the `evaluate-at-state-from-stdin` command -->
```
bugarela, line 487: Were you able to test this with your local rust evaluator?

beu5a: Yes this test was working locally as far as I remember.

evaluator/src/itf.rs
erickpintor: I could swear gh showed it closer to the value.rs file. :shrug:

erickpintor (approved):
All good on my end.

beu5a:
> Great work, I'm convinced that this solves the problem 😄. I'm not a big fan of the interfaces you chose, with the indexes of the invariants playing a central role. Maybe we could have some structure like:
> 
> ```
> interface Invariant {
>   identifier: str,
>   expression: QuintEx,
> }
> ```
> 
> so then we can easily filter (by expression) during evaluation and print (identifier) during printing.
> 
> This is just a suggestion, but I left some comments on some more concrete improvements.

I addressed the concrete comments. WRT to the interface, I'd skip it for now. We might need to polish the TS/Rust interface in the future (the current interfaces are heavily influenced by the old TS code) and we will think about a better solution than.

PR #1904, opened 2026-02-18
Add `find all references`  feature to vscode extension

oakenknight:
Hello 🐙 

I added new feature to vscode extension allowing to find all references. I found this very helpful when reviewing longer specs.

### DIFF
--- a/vscode/quint-vscode/server/src/definitions.ts
+++ b/vscode/quint-vscode/server/src/definitions.ts
@@ -1,7 +1,8 @@
 import { Loc, ParserPhase3 } from '@informalsystems/quint'
-import { findName, isPositionInLoc } from './reporting'
+import { findName } from './reporting'
 import { Position } from 'vscode-languageserver/node'
 import { URI } from 'vscode-uri'
+import { selectBestBindingMatch } from './symbolLookup'
 
 export interface QuintDefinitionLink {
   nameId: bigint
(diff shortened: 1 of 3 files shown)


oakenknight:
I did not have any performance issues with monad-spec (it listed references in a second or less), and it did not stop me from interacting with the spec in the meantime. But I might be able to optimize it.

vscode/quint-vscode/server/src/definitions.ts
```diff
@@ -27,7 +27,33 @@ export function findDefinition(
...
+      }
+    }
+
+    const declaration = candidates.find(([refId, binding]) => refId === binding.id)
+    const byName = declaration?.[1] ?? candidates[0][1]
+    return {
+      nameId: id,
+      name,
+      definitionId: byName.id,
+    }
```
erickpintor: Q: Does it make sense to merge the `filter` and both `find` predicates into a single call to `find` to potentially short-circuit early when a suitable candidate is found? Unless, of course, the order of the `find` calls are relevant, meaning that you first want to examine for local ref, then by name.

oakenknight:
> LGTM. The multiple passes over containers like the table are concerning but, if the editor isn't freezing I think it's okay. I have a question about one pattern that happens in a few places: build a list of candidates, try to find some thing, then try to find something else. Other than that, I think we're good.

I refactored the filter + find + find pattern into a shared single-pass helper (selectBestBindingMatch) used by both definition and references paths. It short-circuits on local matches while preserving precedence (local > declaration > first name match).

PR #1905, opened 2026-02-18
[QNT-164] Fixing KSet examples spec

miloradtrninic:
- Rename symbols for readability: paper-style names (N, F, V, r, y, S, x, Proc)
  → descriptive names (NUM_PPROCESSES, MAX_FAULTS, receivedValues,
  currentRound, decision, valuesCurrentRound, proposalValue, Process)
  in ConsensusAlg and KSetAgreementConsensus.

- Fix run parameters: use ((MAX_FAULTS + 1) * 4 + 1) step reps (and
  (MAX_FAULTS/K + 1) * 4 + 1 for K-set) so runs execute enough steps
  per round. The previous (F + 1) / (F/K + 1) reps were incorrect.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -63,17 +63,33 @@ jobs:
       - name: Install node
         uses: actions/setup-node@v4
 
-      - name: Install Quint
+      # --- Run benchmarks on main (baseline) ---
+      - name: Checkout main branch
+        run: git checkout main
+
+      - name: Install Quint (main)
         run: cd ../quint && npm ci && npm run compile && npm link
 
-      - name: Build benchmarks
+      - name: Build benchmarks (main)
         run: cargo build --release --benches
 
-      - name: Run benchmark
-        run: |
-          # Run benchmarks and save results
-          cargo criterion --message-format=json > benchmark_results.json
+      - name: Run benchmarks (main)
+        run: cargo criterion --message-format=json > baseline_results.json
 
+      # --- Run benchmarks on PR branch ---
+      - name: Checkout PR branch
+        run: git checkout ${{ github.head_ref }}
+
+      - name: Install Quint (PR)
+        run: cd ../quint && npm ci && npm run compile && npm link
+
+      - name: Build benchmarks (PR)
+        run: cargo build --release --benches
+
+      - name: Run benchmarks (PR)
+        run: cargo criterion --message-format=json > benchmark_results.json
+
+      # --- Compare results ---
       - name: Process benchmark results
         id: benchmark-processing
         uses: actions/github-script@v7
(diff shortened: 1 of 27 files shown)


examples/classic/distributed/ConsensusAlgorithm/ConsensusAlg.qnt
```diff
@@ -253,37 +252,37 @@ module ConsensusAlg {
...
-  /// We crash process p, and the set s does not receive p's messages
-  run stepHidePsMessagesFromS(p, s) = {
-    crashProcess(p).then(sendMessages).then(receiveMessage(s)).then(computeAction)
+  /// We crash one process, and the set of excluded processes does not receive that process's messages
+  run stepHideProcessMessagesFromSet(crashedProcess, excludedProcesses) = {
+    crashProcess(crashedProcess).then(sendMessages).then(receiveMessage(excludedProcesses)).then(computeAction)
   }
 
-  /// If there is at most one fault (F = 1) two rounds should be enough to reach agreement.
+  /// If there is at most one fault (MaxFaults = 1) two rounds should be enough to reach agreement.
```
bugarela:
```suggestion
  /// If there is at most one fault (MAX_FAULTS = 1) two rounds should be enough to reach agreement.
```

examples/classic/distributed/ConsensusAlgorithm/ConsensusAlg.qnt
```diff
@@ -45,26 +45,24 @@ module ConsensusAlg {
...
-    val newV = s.V.union(flatten(s.S))
-    val newR = s.r + 1
-    val newY = if (s.r == MAX_ROUNDS) Some(minValue(newV)) else s.y
+  pure def compute(state: LocalState): LocalState = {
+    val newReceivedValues = state.receivedValues.union(flatten(state.valuesCurrentRound))
+    val newRound = state.currentRound + 1
+    val newDecision = if (state.currentRound == MAX_ROUNDS) Some(minValue(newReceivedValues)) else state.decision
 
-    { V: newV, r: newR, y: newY, S: Set(), x: s.x }
+    { receivedValues: newReceivedValues, currentRound: newRound, decision: newDecision, valuesCurrentRound: Set(), proposalValue: state.proposalValue }
```
bugarela: We probably should break this (and similar instances) into multiple lines. Ideally we should keep it at 80 columns across all specs since that works best on most people's setup (and also on git side-by-side diffs)

miloradtrninic: image
I did leave it at 100, 80 felt too narrow.

This is a screenshot of my laptop screen.

bugarela: Looks good! Yeah, 100 is great, thanks!

examples/classic/distributed/ConsensusAlgorithm/ConsensusAlg.qnt
```diff
@@ -253,37 +276,41 @@ module ConsensusAlg {
 
 module properValues {
   // quint run --main properValues --invariant agreement ConsensusAlg.qnt
-  import ConsensusAlg(N = 6, F = 1, ACTUAL_FAULTS = 1, MAX_ROUNDS = 2).*
+  import ConsensusAlg(NUM_PPROCESSES = 6, MAX_FAULTS = 1, ACTUAL_FAULTS = 1, MAX_ROUNDS = 2).*
 
   run consensusRunTest =
     init
-      .then((F + 1).reps(_ => step))
+      .then(((MAX_FAULTS + 1) * 4 + 1).reps(_ => step))
```
josef-widder: Great catch! I am just wondering about the '+1' here. Wouldn't that be always the stuttering step. Would something like this
```suggestion
      .then(((MAX_FAULTS + 1) * 4).reps(_ => step))
```
also work

josef-widder: Also, I guess we didn't catch it earlier because we don't check whether someone decided. Perhaps we should add something like `.expect(procState.keys().exists(p => procstate.get(p).decided != None))` before we check agreement.

miloradtrninic: I wanted to have one stutter step included. I can remove it. 

Will add the expect !

josef-widder: makes sense. Perhaps add a comment that says that the +1 is for the stutter step.

miloradtrninic: I removed stutter after all. Added expect. Can you double check this so we can merge?

miloradtrninic: @josef-widder ping for visibility

PR #1910, opened 2026-02-20
Add stack trace to Quint errors (Rust only)

beu5a:
Hey,
This a PR to add a stack trace in case of runtine error in the rust backend and changed the error rendering a little bit. I aimed to have a compact trace with pointers to the problematic expressions (screenshot below). I had to add some glue code on the TS side to keep backward compatibility with TS backend but it should be very easy to remove when we decide to fully deprecate it. I added unit tests and updated the integration tests required to pass the CI checks. The rest of integration test updates will be updated and organized in a separate PR. 

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for the `--n-traces` flag in the Rust backend (#1898)
 - Added support for `--invariants` flag in the Rust backend (#1902)
 - The rust backend will report diagnostics for the REPL (#1909)
+- The rust backend now displays a stack trace on runtime errors (#1910)
 
 ### Changed
 


evaluator/tests/evaluation_tests.rs
```diff
@@ -1309,3 +1309,31 @@ fn cardinality_overflow() -> Result<(), Box<dyn std::error::Error>> {
...
+}
+
+#[test]
+fn error_trace_through_user_def() {
+    // When a user-defined operator triggers an error, the trace should grow
+    // with each call site in the chain
+    let result = eval_expr(
+        "def f(x) = 1 / x
+         def g(y) = f(y)
+         g(0)",
```
bugarela, line 1331: Can we add this test case to either rust-repl.test.ts or the integration tests so we have the full output as expectation as we can see in the screenshot in your PR?

evaluator/src/ir.rs
```diff
@@ -19,25 +19,30 @@ pub type QuintName = HipStr<'static>;
...
+            trace: Vec::new(),
         }
     }
 
     pub fn with_reference(self, reference: QuintId) -> Self {
         QuintError {
             code: self.code,
             message: self.message,
-            reference: Some(reference),
+            trace: vec![reference],
```
erickpintor, line 42: Q: should we add a `debug_assert!` to ensure that we don't accidentally override a full stack-trace with a single reference id?

erickpintor (approved):
LGTM.

evaluator/tests/evaluation_tests.rs
beu5a: I will add this test as part of part of the integration tests refactoring PR

PR #1882, opened 2026-01-31
Stream JSON values in/out to avoid extra mallocs

erickpintor:
Serde supports reader/writer interface which avoid the allocation of an intermediate string in memory for json serialization. This can reduce peak memory allocation when serializing large traces back to quint CLI.

- [x] I have read and I understand the [Note on AI-assisted contributions](https://github.com/informalsystems/quint/blob/main/CONTRIBUTING.md#note-on-ai-assisted-contributions)
- [ ] Changes manually tested locally and confirmed to work as described
      (including screenshots is helpful)
- [ ] Tests added for any new code
- [ ] Documentation added for any new functionality
- [x] Entries added to the respective `CHANGELOG.md` for any new functionality

PR #1911, opened 2026-02-20
Fix progress bar in Rust backend

beu5a:
Hey,
This is a one liner to fix a recently discovered bug with the progress bar in Rust where single threaded simulations and tests don't advance the progress bar. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added support for `--invariants` flag in the Rust backend (#1902)
 - The rust backend will report diagnostics for the REPL (#1909)
 - The rust backend now displays a stack trace on runtime errors (#1910)
+- Added support for `--seed` int the REPL CLI (#1912)
 
 ### Changed
 
--- a/docs/content/docs/quint.md
+++ b/docs/content/docs/quint.md
@@ -54,6 +54,9 @@ Options:
                  the module                                             [string]
       --verbosity  control how much output is produced (0 to 5)
                                                            [number] [default: 2]
+      --seed       random seed to use for non-deterministic choice      [string]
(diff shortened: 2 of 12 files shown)


beu5a:
It might. @erickpintor introduced this change in a performance optimization PR (#1882). He might have more context on the impact of this on multithreaded code.

evaluator/src/progress.rs
```diff
@@ -54,7 +54,8 @@ impl Reporter for JsonStdErr {
     fn next_sample(&mut self) {
         self.current_samples += 1;
         let progress = Self::fmt(self.current_samples, self.total_samples);
-        serde_json::to_writer(io::stderr(), &progress).expect("failed to write progress to stdout");
+        let mut stderr = io::stderr().lock();
+        writeln!(stderr, "{progress}").expect("failed to write progress to stderr");
```
erickpintor: No need for locking. What's fixing the bug is the fact that we're writing a new line here (which I missed in my earlier patch).
```suggestion
        let mut stderr = io::stderr();
        serde_json::to_writer(stderr, &progress).expect("failed to write progress to stdout");
        write!(stderr, "\n").expect("failed to write \n to stderr");
```

erickpintor: BTW: I should have applied the same pattern to the parallel reporter at `reporter_thread_loop` (this time with a lock), in case you want to fix that as well.

beu5a: lol I thought that I added the lock to the parallel reporter not the single threaded one, but the code seems to be working for the multithreading without the lock so I don't know if that's a good idea to add the overhead of a lock for this

erickpintor: The parallel one has a lock already ;) I was referring to changing the parallel one to use `serde_json::to_writer(..) + write!(stderr, "\n")` for efficiency.

PR #1912, opened 2026-02-20
Fix REPL dot-commands

bugarela:
Hello :octocat: 

I forgot to cover this on the REPL implementation, and our tests were not enough to test the behavior (only the interface, which kept working as that is in TS). So I fixed things and added new tests.

I also added the `--seed` CLI option to REPL as I found odd it was not supported.

### DIFF
--- a/evaluator/src/evaluator.rs
+++ b/evaluator/src/evaluator.rs
@@ -519,9 +519,8 @@ impl Interpreter {
         let compiled_expr = self.compile_expr_core(expr);
         let wrapped_expr = CompiledExpr::new(move |env| {
             compiled_expr.execute(env).map_err(|err| {
-                // This is where we add the reference to the error, if it is not already there.
-                // This way, we don't need to worry about references anywhere else :)
-                if err.reference.is_none() {
+                // Add the reference to the error trace if it's empty (first error location)
+                if err.trace.is_empty() {
                     return err.with_reference(id);
                 }
                 err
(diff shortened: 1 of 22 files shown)


evaluator/src/evaluator.rs
```diff
@@ -761,13 +760,22 @@ impl Interpreter {
                         .map(|t| t.states)
                         .unwrap_or_default();
 
+                    // Extract diagnostics from the simulation trace states
+                    for state in &env.trace {
+                        env.diagnostics.extend(state.diagnostics.iter().cloned());
+                    }
+
```
erickpintor: I'm a bit concerned this will duplicate debug messages in both the trace and the diagnostics vector (see `shift()`). Is there a way to avoid this duplication?

bugarela: What I understand here is that `simulate` will create a different `env` for its internal usage, and the `env` out here needs to be updated with the info returned by the `simulate` for it to expose some functionality to the user. 

Maybe we could re-use the `env` inside `simulate` which also makes sense for performance. I'll see how weird the interfaces might get with this option.

erickpintor: I see. It wasn't clear to me that we had 2 environments. I'd be cleaner if we could move the one we have at hand inwards, like you suggested.

bugarela: I gave it a shot of moving the env inwards, and I'm not sure whether I'm happy with the interfaces - can you take a look? We can revert this and maybe use a `take` here with a comment explaining things.

I also looked into the scenario I mentioned about `q::debug` not being displayed on runtime errors and the reason seems closely related to this, so you might want to take a look at QNT-234 for a fuller context.

evaluator/src/itf.rs
```diff
@@ -24,7 +24,7 @@ pub struct State {
     pub diagnostics: Vec<DebugMessage>,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
erickpintor: Looks like we don't need this clone anymore?

evaluator/src/evaluator.rs
erickpintor: I think it's fine. We can merge it as is, fix 234 and continue iterating on it.

PR #1917, opened 2026-02-24
Refactoring unit and integration tests

beu5a:
Hello, 
this is a PR to update and organize the unit tests, the integration tests and the CI workflows. A big chunk of the PR is composed of updates to the expectations of the unit and integration tests to match to the latest updates (replacing the TS backend with Rust, adding stack traces to runtime errors, introducing diagnostics...). 

I also changes the behavior of the integration tests, so now most of the tests (beside the distribution tests) build a rust evaluator from `./evaluator` instead of downloading the released evaluator and grouped them in a dedicated folder with a better separation of concerns.
I also renamed some of the CI workflows for better consistency with the rest of the CI checks.

I would really appreciate if you can double check with me if I am using the right `--backend ` flag in the updated tests, I might have missed some tests. 


quint/integration-tests/lang/io.md
```diff
@@ -0,0 +1,347 @@
...
+ Error [QNT405]: Module 'importee2' not found
+
+  at       HOME/_1060case.qnt:3:3
+  3:   import importee2 as I from "_1022IMPORTEE2"
+       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error: parsing failed
+```
+
+### Error when --n-traces is greater than --max-samples
```
bugarela, line 312: Maybe this file should be named CLI or compiler? I was surprised to see these `quint run` tests here, but then realize they are only checking CLI validation of `quint run`. I find the `lang` name kinda confusing.

quint/integration-tests/runtime/rust/test.md
```diff
@@ -0,0 +1,670 @@
+# Integration tests with locally-built Rust backend
+
+Tests in this script verify features of the Rust backend that require
+running against a locally-built evaluator (not the released version).
+
+These tests run in PR CI after building the Rust evaluator from source.
```
bugarela: Does this comment still makes sense? IIUC, all integration tests besides the distribution ones have this now.

quint/test/runtime/rust/repl.test.ts
```diff
@@ -379,46 +389,6 @@ describe('repl ok', () => {
     await assertRepl(input, output)
   })
 
-  it('update the seed between evaluations', async () => {
```
bugarela, line 382: why was this removed?

quint/test/runtime/rust/repl.test.ts
```diff
@@ -652,48 +634,56 @@ describe('repl ok', () => {
...
       |true
+      |{ x: 2 }
       |>>> 2 <= x and x <= 5
       |true
       |>>> nondet t = oneOf(tuples(2.to(5), 3.to(4))); x' = t._1 + t._2
       |true
+      |{ x: 2 => 6 }
       |>>> 5 <= x and x <= 9
       |true
-      |>>> nondet i = oneOf(Nat); x' = i
```
bugarela, line 687: why is this changed? We should be supporting `oneOf(Nat)` etc

quint/test/runtime/deprecated/repl.test.ts
```diff
@@ -57,7 +57,7 @@ const withIO = async (inputText: string): Promise<string> => {
   // Use { end: false } to prevent ending output when input ends
   input.pipe(output, { end: false })
 
-  const rl = quintRepl(input, output, { verbosity: 2, backend: 'rust' }, () => {})
+  const rl = quintRepl(input, output, { verbosity: 2 }, () => {})
```
bugarela: We should make `rust` the default in this function.
```suggestion
  const rl = quintRepl(input, output, { verbosity: 2, backend: 'typescript' }, () => {})
```

quint/test/runtime/deprecated/repl.test.ts
```diff
@@ -77,11 +77,12 @@ const withIO = async (inputText: string): Promise<string> => {
   await once(rl, 'close')
 
   chalk.level = savedChalkLevel
-  return output.buffer
+  // Remove trailing newline that gets added when the REPL closes
+  return output.buffer.replace(/\n$/, '')
 }
 
 // the standard banner, which gets repeated
-const banner = `Quint REPL ${version} (using the Rust backend)
```
bugarela, line 84: I'm confused by this changes. It might be the github renaming thing that is fooling me. Did we need to change all these seed and state display things from the existing typescript tests? I didn't expect to see changes on the diff here, just the renaming

quint/test/runtime/rust/repl.test.ts
```diff
@@ -57,7 +57,7 @@ const withIO = async (inputText: string): Promise<string> => {
   // Use { end: false } to prevent ending output when input ends
   input.pipe(output, { end: false })
 
-  const rl = quintRepl(input, output, { verbosity: 2 }, () => {})
+  const rl = quintRepl(input, output, { verbosity: 2, backend: 'rust' }, () => {})
```
bugarela, line 60: if we make rust the default, we can remove this
```suggestion
  const rl = quintRepl(input, output, { verbosity: 2 }, () => {})
```

quint/test/runtime/deprecated/repl.test.ts
beu5a: the github renaming causes a lot of confusion here. It is detecting a wrong original file.

beu5a: During the reorg, I might have confused the repl-rust and repl unit tests, which resulted in this heavily updated version of the test file. I will revert the file back to the version on main.

quint/test/runtime/rust/repl.test.ts
beu5a: this was removed on the rust side because picking very large integers and displaying them (sometimes over multiple lines) was causing problems with the expected output.

quint/test/runtime/rust/repl.test.ts
beu5a: same as above, but this was only supposed to be removed from the rust side.

beu5a:
> Looking good. I have some minor things, and a lack of understanding on the unit test changes

@bugarela , I have to revert back the repl tests for the deprecated backend to main and apply some small changes to the test expectations (to include state change displays on verbosity=2) which were not covered by the expected output because that part of the code was writing directly to the console log instead of the out const and were thus not displayed in the test output (which uses it's own output).  I guess that file was the source of the confusion.

quint/integration-tests/lang/io.md
beu5a: This was named `io` based on the original test file called `cli-io`. There is already another file named `cli` which kept the same name as the original file. I find compiler more confusing than the current name.

PR #1918, opened 2026-02-26
Flush pending diagnostics on runtime errors

beu5a:
Hello
This is a PR to address the case where the diagnostics of the latest state from the rust evaluator are not being printed as part of the trace if a runtime error occurs. This functionality was added by extending the `SimulationError `  structure to include a field for the failed diagnostics.

image

### DIFF
--- a/quint/integration-tests/runtime/rust/run.md
+++ b/quint/integration-tests/runtime/rust/run.md
@@ -659,3 +659,19 @@ rm /tmp/debug_test.qnt
 ```
 [DEBUG] this tests debug 42
 ```
+
+## Pending diagnostics are flushed on runtime error
+
+Regression test for the diagnostics flushing feature: q::debug messages accumulated
+during a step that fails with a runtime error should be printed as pending diagnostics.
+
+<!-- !test in pending diagnostics flushed on runtime error -->
+```
+quint run --backend=rust --verbosity=3 --seed=0xf3358ea77be2efe2 \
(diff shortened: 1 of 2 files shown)


evaluator/src/itf.rs
```diff
@@ -92,6 +100,13 @@ impl Trace {
                 "ok".to_string()
             },
         );
+        if !pending_diagnostics.is_empty() {
+            other.insert(
+                "pendingDiagnostics".to_string(),
```
bugarela: I believe we are using snake_case for all fields in ITF.
```suggestion
                "pending_diagnostics".to_string(),
```

quint/src/itf.ts
```diff
@@ -270,6 +270,17 @@ export interface DebugMessage {
   value: QuintEx
 }
 
+/** Extracts pending diagnostics from the ITF trace-level metadata.
+ * These are diagnostics from a failing step that never produced a new state. */
+export function pendingDiagnosticsOfItf(itf: ItfTrace): DebugMessage[] {
+  if (itf['#meta']?.pendingDiagnostics) {
+    const msgs: { label: string; value: any }[] = JSON.parse(itf['#meta'].pendingDiagnostics)
```
bugarela:
```suggestion
  if (itf['#meta']?.pending_diagnostics) {
    const msgs: { label: string; value: any }[] = JSON.parse(itf['#meta'].pending_diagnostics)
```

bugarela (commented):
We could use a regression test here

beu5a:
> We could use a regression test here

I will add a regression test after we merge the test updates PR #1917

PR #1932, opened 2026-03-10
Add `leadsTo` and fix fairness effects

bugarela:
Hello :octocat: 

We were missing one temporal operator for no reason, `leadsTo` (or `~>` in TLA+). This adds that, as this felt too big of a shame now that we support TLC. I'm bumping Apalache as I added support for `leadsTo` transpilation to it.

I also fixed one bug where Quint would complain about the effects of `weakFair` and `strongFair` in some scenarios. The effect was wrong, but the infrastructure for effects made it not so trivial to fix it. I ended up updating the effect parser to allow for a smooth change that keeps the nice format of the effect signatures (without needing to define the signature manually for this special case).

I also added some cool toy examples for fairness so we can properly test it, on top of the existing EWD 840 spec.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `leadsTo` temporal operator (#1932)
 
 ### Changed
+
+- TLC backend now ensures Apalache distribution is available locally before running TLC
+- Improved error propagation when the Rust evaluator fails to start, now shows the actual error instead of `Error: write EPIPE` (#1958)
+
 ### Deprecated
 ### Removed
 ### Fixed
--- a/quint/src/rust/commandWrapper.ts
+++ b/quint/src/rust/commandWrapper.ts
@@ -312,19 +312,27 @@ export class CommandWrapper {
     const startTime = Date.now()
 
     // Spawn the Rust evaluator in subprocess
+    let spawnError: Error | undefined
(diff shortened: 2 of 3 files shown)


examples/classic/distributed/ewd840/ewd840.qnt
```diff
@@ -122,10 +122,6 @@ module ewd840 {
   // FIXME(#630): OrKeep takes an arbitrary type for now
   val vars = (active, color, tpos, tcolor)
 
-  // FIXME(#1236): we cannot add `init` to this property since, in quint, it is an
-  // action - not a predicate
-  // temporal Spec = init and always(step.orKeep(vars) and System.weakFair(vars))
```
bugarela, line 127: I don't think we ever want to support this. I'd much rather have people always add the fairness assumptions to the properties. It's a nightmare to explain that we have an alternative to `init` and `step` with `spec` that takes a temporal formula, blabla.

bugarela:
PR on Apalache side: https://github.com/apalache-mc/apalache/pull/3294

beu5a:
@bugarela I updated the test expectations and made them check the temporal properties with TLC. The CI was failing with the examples dashboard because of the Apalache JAR not being found, so I decided to fix this once for all and made the `verify --backend=tlc` command always try to download apalache if it can't find it. This was added in this PR (#1959). The last commit I added here was cherry-picked to PR #1959. The CI is passing now, I just need to merge the other PR before merging this.

PR #1969, opened 2026-04-08
fix: detect state variable shadowing --init/--step action names

pilt:

**AI disclosure:** This PR was developed with AI assistance (Claude). The fix, tests, and manual verification were reviewed by a human.

---

## Problem

When a module defines a state variable named `step` or `init`, `quint run` resolves the `--step`/`--init` defaults to the variable instead of the transition action. The failure mode depends on the variable type and backend:

| | `var step: int` | `var step: bool` (init false) | `var step: bool` (init true) | `var init` |
|---|---|---|---|---|
| **rust** | `[QNT500] Expected boolean` | Silent: only init state, reports ok | `[QNT502] Variable x not set` | `[QNT502]` + error reporter crash |
| **typescript** | Uncaught JS exception | Silent: only init state, reports ok | `[QNT502] Variable x not set` | `[QNT502]` + error reporter crash |

The `var step: bool` case initialized to `false` is the most misleading. The simulator evaluates the variable (which is `false`) as the step result, treats the step as disabled, and reports success with only the init state. No error, no warning:

```
$ quint run example.qnt --invariant=inv --max-steps=10
[State 0] { step: false, x: 0 }
[ok] No violation found
Trace length statistics: max=1, min=1, average=1.00
```

When initialized to `true`, the simulator treats the step as succeeding but no state variables are assigned, so the next invariant check crashes with `Variable x not set`.

The same issue affects `quint compile` and `quint verify`. The generated `q::step`/`q::init` definitions silently bind to the variable.

**Repro (`var step: int`):**
```quint
module example {
  var x: int
  var step: int
  action init = all { x' = 0, step' = 0 }
  action doStep = all { x' = x + 1, step' = step + 1 }
  val inv = x >= 0
}
```

**Repro (`var step: bool`, silent wrong result):**
```quint
module example {
  var x: int
  var step: bool
  action init = all { x' = 0, step' = false }
  action doStep = all { x' = x + 1, step' = not(step) }
  val inv = x >= 0
}
```

Workaround: `--step=doStep` or `--init=myInit`.

## Fix

Add an optional `expectedRole` parameter to `toExpr()` in `cliHelpers.ts`. When resolving `--init` or `--step` arguments, check if the resolved name is a state variable (`kind === 'var'`) and return a descriptive error instead of letting downstream code crash or silently misbehave.

Validation runs before backend selection, so all backends (rust, typescript) and all commands (run, compile, verify) are covered.

**After:**
```
error: [QNT501] 'step' resolves to a state variable, not an action.
Rename the variable or use --step to specify a different action name.
```

## Changes

- `quint/src/cliHelpers.ts`: Add `expectedRole?: { kind: 'action', flag: string }` to `toExpr()`, check resolved definition kind, include the correct flag name in the error message
- `quint/src/cliCommands.ts`: Pass the role constraint for `--init` and `--step` in both `runSimulator()` and `compile()`
- `quint/test/cli.test.ts`: 4 new tests (rejects `var step`, rejects `var init` with `--init` in message, accepts action, accepts var without role constraint)
- `CHANGELOG.md`: Entry under UNRELEASED / Fixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 ### Removed
 ### Fixed
+
+- Fixed `--step`/`--init` resolving to a state variable instead of an action when the variable is named `step` or `init` (#1969)
+
 ### Security
 
 ## v0.32.0 -- 2026-03-31
--- a/quint/integration-tests/verification.md
+++ b/quint/integration-tests/verification.md
@@ -30,7 +30,7 @@ quint verify --init=invalidInit ../examples/language-features/booleans.qnt
 <!-- !test err invalid init -->
 ```
 error: [QNT404] Name 'invalidInit' not found
-error: name resolution failed
+error: Argument error
 ```
 
 
--- a/quint/src/cliCommands.ts
+++ b/quint/src/cliCommands.ts
@@ -398,9 +398,11 @@ export async function runSimulator(prev: TypecheckedStage): Promise<CLIProcedure
 
   const recorder = newTraceRecorder(options.verbosity, options.rng, options.numberOfTraces)
 
-  const argsParsingResult = mergeInMany(
-    [prev.args.init, prev.args.step, invariantString, ...prev.args.witnesses].map(e => toExpr(prev, e))
-  )
+  const argsParsingResult = mergeInMany([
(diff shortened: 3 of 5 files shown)


beu5a:
Thank you for this nice contribution! I extended the fix to handle the case where `step` and `init` are used by a val/const too. I will merge this asap 😄

pilt:
Thanks, glad to contribute! And thanks for Quint!

beu5a (approved):

PR #1971, opened 2026-04-14
Remove `init/step` for JSON compilation

### LINKED ISSUE #1584: Compiling into JSON should not require `init` and `step`

beu5a:
Hey,  
This PR makes `--init` and `--step` optional for `quint compile --target json` (Closes #1584 )

Previously `quint compile` always required `init` and `step` to resolve, even when compiling to JSON where they're not needed:

  ```
  $ quint compile test.qnt
  error: [QNT404] Name 'init' not found
  ```
  #### New Behavior

  - `quint compile foo.qnt` (default target JSON) → works even if `foo.qnt` has no `init`/`step` definitions
  - `quint compile foo.qnt --init=myInit --step=myStep` → resolves and validates normally; errors on failure
  - `quint compile foo.qnt --target=tlaplus` → defaults to `init`/`step`; errors if they don't resolve
 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed `--step`/`--init` resolving to a state variable instead of an action when the variable is named `step` or `init` (#1969)
+- `quint compile --target=json` no longer requires `init` and `step` to exist in the module (#1971)
 
 ### Security
 
--- a/evaluator/benches/tuples.rs
+++ b/evaluator/benches/tuples.rs
@@ -10,8 +10,6 @@ fn run_in_rust(input: &str) -> Result<(), Box<dyn std::error::Error>> {
     let quint_content = format!(
         "module main {{
           val input = {input}
-          val init = true
-          val step = true
         }}"
     );
 
@@ -46,8 +44,6 @@ fn run_in_quint_repl(input: &str) -> Result<(), Box<dyn std::error::Error>> {
     let quint_content = format!(
         "module main {{
           val input = {input}
-          val init = true
-          val step = true
         }}"
     );
 
--- a/evaluator/src/helpers.rs
+++ b/evaluator/src/helpers.rs
@@ -18,20 +18,13 @@ use std::process::Command;
 use std::{error::Error, io::Write};
 use tempfile::NamedTempFile;
 
-pub fn parse(
-    quint_content: &str,
-    init: &str,
-    step: &str,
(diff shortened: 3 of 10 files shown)


beu5a:
CI Benchmark is failing because of the failure on main.

bugarela:
We might still need to require them when `--flatten=true`. When flattening, we drop any unused definitions, so if we don't have the entry point defs (init and step), I think it might just drop everything. Can you please check?

erickpintor (approved):
:sheep: it.

beu5a:
> We might still need to require them when `--flatten=true`. When flattening, we drop any unused definitions, so if we don't have the entry point defs (init and step), I think it might just drop everything. Can you please check?

I didn't manage to find the code path where the dead code elimination is happening when flattening. Do you have any pointers to where the unused defs are dropped ?

bugarela:
> I didn't manage to find the code path where the dead code elimination is happening when flattening. Do you have any pointers to where the unused defs are dropped ?

There is not a specific place where this happens, this is just a consequence of the nature of the flattening algorithm. The flattening algorithm will start from definitions already in the main module and recurse from there into the things those definitions use, so if you have:

```
module A {
  action init = ...
  action step = ...
  action otherStep = ...
}

moudle B {
  import A.*

  val my_inv = ...
}
```

we will automatically add:
```
moudle B {
  import A.*

  val my_inv = ...

  action q::init = init
  action q::step = step
}
```

which will make it so `init` and `step` are not dropped, since they are used. `otherStep` would be dropped in this situation (wouldn't be copied to the resulting module of flattening)

beu5a:
> > I didn't manage to find the code path where the dead code elimination is happening when flattening. Do you have any pointers to where the unused defs are dropped ?
> 
> There is not a specific place where this happens, this is just a consequence of the nature of the flattening algorithm. The flattening algorithm will start from definitions already in the main module and recurse from there into the things those definitions use, so if you have:
> 
> ```
> module A {
>   action init = ...
>   action step = ...
>   action otherStep = ...
> }
> 
> moudle B {
>   import A.*
> 
>   val my_inv = ...
> }
> ```
> 
> we will automatically add:
> 
> ```
> moudle B {
>   import A.*
> 
>   val my_inv = ...
> 
>   action q::init = init
>   action q::step = step
> }
> ```
> 
> which will make it so `init` and `step` are not dropped, since they are used. `otherStep` would be dropped in this situation (wouldn't be copied to the resulting module of flattening)

It makes sense. The current approach breaks flattening in the case where we don't explicitly pass `--init/--step` to `compile`.  I will add the defaults for the case where `flatten=true`.

PR #1992, opened 2026-06-29
Use `reduce` to not overflow the stack

bkchr:
Huge number of traces was leading to stack overflows in simulation.ts. This is being solved by using `reduce` instead of generating one big `max`/`min` call that is taking all traces as an argument.

      - Claude was used to find and fix this.


beu5a (approved):
LGTM 😄

